PR #2, opened 2019-07-03
Fix crash with Aqara S2 lock Koenkk/zigbee2mqtt#349

dzungpv:
Fix issue: Koenkk/zigbee2mqtt#349 (comment)
This code need for the pull Koenkk/zigbee-shepherd-converters#497
@Koenkk I make less change to zcl-packet, i also test it not affect exist device support in zigbee-shepherd-converters, it only crash with Aqara S2 doorlock and struct-string data of Xiaomi, atribute id: 65281 so i whitelist only Xiaomi lock from this code.

### DIFF
--- a/src/zcl-packet/lib/foundation.ts
+++ b/src/zcl-packet/lib/foundation.ts
@@ -625,8 +625,9 @@ ru.clause('strPreLenUint8', function (name) {
     parsedBufLen += 1;
     this.uint8('len').tap(function () {
         var attrId = this.vars['attrId'];
-        // special xiaomi struct-string
-        if (attrId === 65281) {
+        var deviceId = zclId.device(260, 10).key;
+        // special xiaomi struct-string, except xiaomi doorlock because it will crash
+        if (attrId === 65281&&deviceId!=='doorLock') {
             ru['xiaoMiStruct'](name)(this);
         } else {
             parsedBufLen += this.vars.len;


Koenkk:
Thanks!

Koenkk:
Unfortunately I have to revert this as it's causing issues (https://github.com/Koenkk/zigbee2mqtt/issues/1710). 

I've checked the code again but it doesn't make sense to me, 

`var deviceId = zclId.device(260, 10).key;` always evaluates to `doorLock`, therefore the `if (attrId === 65281&&deviceId!=='doorLock') {` will always evaluate to false. Can you check how to fix this?

dzungpv:
@Koenkk i will check this, but with xiaomi struct decode it will crash with some devices, i remember the author of those code said that we must let the decode without xiaomi struct and then apply it with specific device later, because it again the standard. Example we can apply those code for devices in zigbee seppherd conveter only, after convert ascii string back to hex, so it will never crash

dzungpv:
@Koenkk We can replace function **ru.clause('xiaoMiStruct', function (name)** with those code in the zigbee-shepherd-converters: https://github.com/TedTolboom/com.xiaomi-mi-zigbee/blob/e7a7ac0e9ec4129cefabc5d2fba82e939f09529e/drivers/sensor_magnet.aq2/device.js#L65

Koenkk:
Can you make a pr and test?

dzungpv:
@Koenkk Yes, i will test it with some Xiaomi devices.
https://github.com/Koenkk/zigbee-shepherd-converters/pull/514
https://github.com/Koenkk/zigbee-herdsman/pull/8

dzungpv:
@Koenkk i tested the code but for some data it not decode like original "xiaoMiStruct" (example voltage, power usage of xiaomi plug not decode) function here: https://github.com/Koenkk/zigbee-herdsman/blob/772a3a5bc0b9929fb309d0c86efa408e68a1dc3b/src/zcl-packet/lib/foundation.ts#L648

I am not very good at nodejs or Javascript, you can convert it to normal function and add it to zigbee-shepherd-converters **parseData** function? I wrote here https://github.com/dzungpv/zigbee-shepherd-converters/blob/6c3ab10f2e1fa022f6f94452b222a23f0dffda5c/converters/fromZigbee.js#L7

If we can move it here, it will fix the bug with Aqara S2 Lock and other undecode data, because those data is original string in Zigbee standard Xiaomi use to encode their data

PR #8, opened 2019-07-11
Remove "xiaoMiStruct", move decode to zigbee-shepherd-converters

dzungpv:


src/zcl-packet/lib/foundation.ts
```diff
@@ -624,14 +624,8 @@ ru.clause('uint64', function (name) {
 ru.clause('strPreLenUint8', function (name) {
     parsedBufLen += 1;
     this.uint8('len').tap(function () {
-        var attrId = this.vars['attrId'];
-        // special xiaomi struct-string
-        if (attrId === 65281) {
-            ru['xiaoMiStruct'](name)(this);
```
Koenkk: Instead of moving it to zigbee-shepherd-converters, couldn't we just simply add a try catch here? If decoding fails (for eg the doorlock) we use the default code.

dzungpv: @Koenkk nice ideas, i will try to catch it.

Koenkk:
The Zcl refactoring is now almost finished, I think we can easily try catch it here now: https://github.com/Koenkk/zigbee-herdsman/blob/zcl/src/zcl/buffaloZcl.ts#L99.

If you provide the buffer from the Aqara S2 lock we can also add a test here: https://github.com/Koenkk/zigbee-herdsman/blob/zcl/test/zcl.test.ts#L212

PR #10, opened 2019-07-24
Send default response to report

clockbrain:
Refer https://github.com/Koenkk/zigbee2mqtt/issues/1722

This patch works for me but it would be great if others could review and also confirm.

It sends a default response back whenever the coordinator receives an attribute report thereby heading off repeated sending of the report. I figured generating the response in zigbee-herdsman was easier than passing it up to zigbee2mqtt and back for what is effectively just an ack.

Here's the sort of traffic I see without the default response patch.
image

and here is the sort of traffic I see with this patch. The sequence number of the response have to match the report to prevent the repeated report messages.
image


Koenkk:
Awesome, I also prefer to solve it here! Can https://github.com/Koenkk/zigbee2mqtt/issues/1722 be closed?

PR #24, opened 2019-09-13
Do not close the port if it is not open

kirovilya:
To prevent "uncaught exception: Port is not open" when port is busy or wrong.


Koenkk:
Thanks!

PR #34, opened 2019-09-21
extend events.ts with new commands

tb-killa:
add checkin and enrollReq as requested by Konke Devices via log output.

### DIFF
--- a/src/controller/events.ts
+++ b/src/controller/events.ts
@@ -55,6 +55,8 @@ const CommandsLookup: {[s: string]: MessagePayloadType} = {
     'emergency': 'commandEmergency',
     'operationEventNotification': 'commandOperationEventNotification',
     'statusChangeNotification': 'commandStatusChangeNotification',
+    'checkin': 'commandCheckin',
+    'enrollReq': 'commandEnrollReq',
 };
 
 type MessagePayloadType =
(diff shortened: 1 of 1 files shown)


src/controller/events.ts
```diff
@@ -55,6 +55,8 @@ const CommandsLookup: {[s: string]: MessagePayloadType} = {
     'emergency': 'commandEmergency',
     'operationEventNotification': 'commandOperationEventNotification',
     'statusChangeNotification': 'commandStatusChangeNotification',
+    'checkin': 'commandCheckin',
+    'enrollReq': 'commandEnrollReq',
```
Koenkk: Where are we going to use this event?

tb-killa: The commands are from log (herdsman) if i want to join Konke Devices.
Will provide the needing parts from log-file if you want.
We could ignore them if we doesnt need them but maybe Konke Devices need them to do some special ?

tb-killa: Got one for the Contact Sensor of Konke:
``
zigbee-herdsman:controller:log Skipping command 'enrollReq' because it is missing from the lookup +1ms
``

Koenkk: Yes  I understand, but what would be the use to provide this command to zigbee2mqtt? enrolling should already be handled by zigbee-herdsman itself.

Koenkk: Does that mean that for example it will send this when the device joins?

tb-killa: Right if EndDevice support Cluster it could be CE to EndDevice automatically but could also be that the EndDevice themselve ask for the Configuration who it seems to be actually by Konke if you join them.

tb-killa:
Close because the commands are now in own issues.

PR #39, opened 2019-09-27
Unregister adapter events on Controller.stop()

hobbyquaker:
Have the problem that the event listeners get doubled after each Controller.stop() and Controller.start().
Need to stop/start in node-red-contrib-zigbee because the Nodes don't have knowledge if they are started/stopped because Node-RED starts/closes or just because the flows get re-deployed.


Koenkk:
Thanks!

PR #42, opened 2019-09-29
add Controller.setLED method, remove Controller.disableLED

hobbyquaker:


Koenkk:
I think we should merge the `disableLED` and `enableLED` into one function. 

Something like `setLED(enabled: boolean)`?

hobbyquaker:
makes sense, will do :)

Koenkk:
Great, thanks again!

hobbyquaker:
You're welcome :-)

PR #45, opened 2019-09-29
introduce Endpoint.binds property

hobbyquaker:


hobbyquaker:
I think we need a save() after adding members to the bind array, but I have no clue how to trigger that in endpoint.ts, do I have access to Controller.getDeviceByIeeeAddr() there?

Koenkk:
Thanks! (will fix tests later)

hobbyquaker:
Cool, thanks! Looks much cleaner then my approach, looking at the diffs is a good training for me :)

hobbyquaker:
Question for my understanding: isn't it necessary to explicitly call Device.save() after a bind is done?

Koenkk:
@hobbyquaker you are too fast; will push that soon.

Note to zigbee-herdsman 0.8.0 will have breaking changes, I will add them to README.md

hobbyquaker:
I‘m right now preparing another PR: get rid of jsdoc, use typedoc. much better. direct code coupling, no efforts anymore for creating api docs, no doc comments needed...

Koenkk:
Ok great! Then the only thing we would need to do is add proper comments to each function?

hobbyquaker:
Exactly. All Params/Returns/Fulfils/Interfaces/Types will be autogenerated.

Koenkk:
Awesome!

hobbyquaker:
Yeah :) Leads to the question: why didn't I google "typescript jsdoc" before starting the docs ;-)

PR #49, opened 2019-10-02
add JSON serialization for devices, endpoints and groups

hobbyquaker:
Hi Koenkk! I'd like to have the possibility to JSON.stringify these Objects (would be more convenient for me, usecase: user interface) - so my first attempt was to add toJSON() methods (those get automatically applied when using JSON.stringify on the objects). This breaks a bunch of tests... If you would accept this PR I would adapt the tests, just wanted to ask you if that PR makes sense and is acceptable for you before putting effort in adapting the tests.

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -71,6 +71,29 @@ class Device extends Entity {
 
     private meta: KeyValue;
 
+    private toJSON(): Object {
+        return {
+            ID: this.ID,
+            ieeeAddr: this.ieeeAddr,
+            applicationVersion: this.applicationVersion,
+            endpoints: this.endpoints,
+            interviewCompleted: this.interviewCompleted,
+            interviewing: this.interviewing,
+            lastSeen: this.lastSeen,
+            type: this.type,
+            dateCode: this.dateCode,
+            hardwareVersion: this.hardwareVersion,
+            manufacturerName: this.manufacturerName,
+            modelID: this.modelID,
+            networkAddress: this.networkAddress,
+            powerSource: this.powerSource,
+            softwareBuildID: this.softwareBuildID,
+            stackVersion: this.stackVersion,
+            zclVersion: this.zclVersion,
+            meta: this.meta
+        };
+    }
+
     // This lookup contains all devices that are queried from the database, this is to ensure that always
     // the same instance is returned.
     private static devices: {[ieeeAddr: string]: Device} = null;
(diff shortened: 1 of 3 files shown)


hobbyquaker:
Another thing: I'm not sure if I'm messing up the database with this change? Can you point me in the right direction? ;)

hobbyquaker:
More thoughts: How about just renaming the existing toDatabaseRecord methods to toJSON and changing `Entity.database.update(this.toDatabaseRecord());` to `Entity.database.update(this);`

Koenkk:
`toDatabaseRecord` also takes care of the legacy format, this is probably not what you want to see as a library user. 

What about `JSON.stringify(device)`, doesn't that already work?

hobbyquaker:
when I just stringify the device object I get the private properties in the resulting JSON, it's ignoring the getters/accessors. So the change from 0.7 to 0.8 broke my UI and I thought it would be a cleaner solution to use toJSON to utilize the getters instead of adapting the UI to use the underscore-prefixed private properties.

Koenkk:
Somehow I don't feel that this logic should be here; what if you e.g. want to show/hide some properties in your UI in the future?

Logic that converts an object to something useful to be shown on the UI should be part of the UI code and not the library code.

hobbyquaker:
I see your point - but I have also another possible view on that question: Isn't it "unbeautiful" that a JSON stringified device object shows up with properties that are private? Another point is that the getters do not always just hand-through the private properties, at some places they do a transformation that then would be lost in the JSON stringified object.

hobbyquaker:
And another thought: if we don't introduce the toJSON that uses the getters we encourage zigbee-herdsman api consumers to work with private properties and should remove the `excludePrivate` from the typedoc config.

Koenkk:
Instead of a `JSON.stringify(device)` in your code; you should explicitly map the properties you want to show on the UI e.g. `toShow = {networkAddress: device.networkAddress, ieeeAddr: device.ieeeAddr}`

How do we encourage them to use private properties? Private should indeed private; if you make a typescript project based in zigbee-herdsman you should even get a compiler error when accessing private properties.

hobbyquaker:
Ok, then the recommendation for herdsman api consumers is so to say "just don't JSON.stringify a device object directly"?

hobbyquaker:
Implemented it in node-red-contrib-zigbee now: https://github.com/hobbyquaker/node-red-contrib-zigbee/blob/master/lib/json.js
But to be honest: that's fine for me, but I'm still not convinced that this belongs there and not in the herdsman. ;-)

PR #80, opened 2019-10-27
add initial support for tcp socket (instead of using socat)

hdo:
Hi Koenkk,

since socat does not work in a docker environment i added initial support for tcp sockets to zigbee-herdsman. The 'serial' handshake/setup portion has been kept as it is no changes there. To use TCP instead of serial one must set the path to 'tcp://<host>:<port>' in the configuration.yaml file. I tested this successfully with my tcp2serial adapter in my enviroment. Please let me know if this pull request is ok for you. Thanks. Regards, hdo


src/adapter/adapter.ts
```diff
@@ -47,15 +48,28 @@ abstract class Adapter extends events.EventEmitter {
...
+                }
+                else {
+                    // check serial port
+
+                    // Path can be a symlink, resolve it.
+                    serialPortOptions.path = RealpathSync(serialPortOptions.path);
+
+                    // Determine adapter to use
+                    for (const candidate of adapters) {
+                        if (await candidate.isValidPath(serialPortOptions.path)) {
```
Koenkk: I don't feel there should be separate branches here, `isValidPath` should also provide the `isValidTcpPath` functionallity. Why has this been separated?

hdo: I first used that but 'RealpathSync' does not work with 'tcp://' entries. You're right this could be also checked at 'isValidPath' but then RealpathSync should be skipped for 'tcp://' entries.

Koenkk: Yes please, realPathSync should be moved to the adapter itself I think

hdo: I reverted the changes on adapter.ts. However to prevent potential side effects with symbolic links on serial paths i modified 'realpathsync' to ignore 'tcp:' path syncing. I also improved tcp path validation in znp.ts with proper exception throwing. Moving 'realpathsync' from adapter.ts to znp.ts would be much more complicated and prone to side effects. What do you think?

Koenkk: Why would this be much more complicated? Now `realpathsync` has unexpected behaviour when passing `tcp:` paths.

hdo: You're right. I reverted realpathsync and moved the calling method from adapter.ts to znp. It's in the constructor of znp now and also in the static isValidPath method of znp.

Koenkk: Good, how can I make a test setup?

hdo: You can test this locally with the ser2net package:

(see https://linux.die.net/man/8/ser2net)

1) Install ser2net package

`sudo apt-get install ser2net`

On my Ubuntu 18.04 LTS only version 2.10 is available but works fine.

2) Edit configuration to bind serial port to network port

`sudo vi /etc/ser2net.conf`

Add following configuration (could also be /dev/ttyACM0 depending on your machine)

`20108:raw:0:/dev/ttyACM1:115200 8DATABITS NONE 1STOPBIT`


3) Edit zigbe2mqtt configuration to use network port

```
vi ./data/configuration.yaml
port: 'tcp://127.0.0.1:20108'
```


Note: The string 'localhost' is currently invalid so i'm using '127.0.0.1'. I'll fix this.


4) Start zigbee2mqtt

`npm start`

Koenkk: Ok thanks, so last thing to do here is add the tests (so it passes CICD checks).

hdo: Alright, i'll try to add the missing tests later today :-)

hdo:
@Koenkk  I finally managed to implement the test cases for the coverage tests. It wasn't trivial as i thought. I had to write 'manual' mocks for the core 'net' package of node.js. However it should work fine now ;-) Can you review the changes and merge those changes? Thanks.

Koenkk:
I will look at it asap, thanks!

src/adapter/z-stack/znp/znp.ts
```diff
@@ -159,7 +228,11 @@ class Znp extends events.EventEmitter {
     }
 
     public static async isValidPath(path: string): Promise<boolean> {
-        return SerialPortUtils.is(path, autoDetectDefinitions);
+ 
+        if (SocketPortUtils.isTcp(path)) {
+            return SocketPortUtils.isValidTcpPath(path);
```
Koenkk: I have one concern here, for TCP it checks wether this is a vaild TCP path instead of being a Zstack adapter. This will give problems when e.g. the Deconz adapter is integrated (when also used via TCP). I see 2 possible solutions here:

- Not sure if this is possible, but retrieve the vendor ID, similar to what is done with the autodetectdefinitions.
- Return false here. it will default to the zstack adapter anyway (https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/adapter.ts#L66). When the deconz adapter is introduced we should then allow to specify the user what adapter is used. But that is for later.

hdo: Retrieving Vendor ID is not possible since sockets don't have such. I'll try the 'false' approach. Do you mean to return false whether 'tcp' is the path prefix?

Koenkk: Yes, as we cannot determine if it is a valid path is should return false.

hdo: @Koenkk, i have changed that and added the neccessary test cases. It works as expected. Can you check again? Thank.

Koenkk:
I did some test and refactors, everything works well :smile:

Can you do a final check? Then this can be merged.

hdo:
> I did some test and refactors, everything works well 
> 
> Can you do a final check? Then this can be merged.

Quite some changes you did there ;-)

I did some finally tests and can happily say that everything works as expected :-)

I think you can merge now.

Can you tell when the next latest-dev docker image will be build/available?

I can't wait to check out the docker version.

Thanks for your help.

Koenkk:
Ok, thanks!

Koenkk:
Docker image should be available now, also updated docs https://github.com/Koenkk/zigbee2mqtt.io/blob/develop/docs/how_tos/how_to_connect_to_a_remote_adapter.md

hdo:
Awesome! Thanks a lot ;-)

peterforeman:
Works very well, many thanks!

Issue #82, opened 2019-11-02
Bad interview of TERNCY-PP01


kirovilya:
Need help analyzing a failed interview with the new TERNCY PP01 device.
[bad_interview.log](https://github.com/Koenkk/zigbee-herdsman/files/3800872/bad_interview.log)

First of all, I added a new endpoint 110 for the coordinator (in startZnp.js), as this is where some commands from the device come.
Second, I tried to add a 'checkin' command in the events.js file, but not handle. Yes, I seen not implemented issue https://github.com/Koenkk/zigbee-herdsman/issues/73

But I don’t understand why the interview is interrupted ... maybe the device did not wait for an answer to its checkin request ...
Even after such an unsuccessful interview, messages from the device come correctly. But the fact that the interview was not completed. Any ideas?

tb-killa:
Hi maybe your don't do a default Response and need this tweak too?

https://github.com/Koenkk/zigbee-herdsman/commit/2de4d56521d4f5d88026c999886b367ef895d8bc

kirovilya:
@tb-killa execution does not reach 'ssIasZone enrolling'. falls on receiving active endpoints https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts#L311
```Interview failed for '0x000d6f00106a64e8 with error 'Error: AREQ - ZDO - activeEpRsp after 10000ms'```

kirovilya:
very strange ... yesterday I mated several times and there has always been a failure ...
Today I performed pairing through the debugger and it turned out to be successful ... maybe some pauses (stops in the debugger) did the trick, because in the normal pairing mode I again failed.
[good_interview.log](https://github.com/Koenkk/zigbee-herdsman/files/3801262/good_interview.log)

kirovilya:
Sometimes it fall with ```Interview failed for '0x000d6f00106a64e8 with error 'Error: AREQ - ZDO - simpleDescRsp after 10000ms'``` on line https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts#L319
:((

Koenkk:
I guess that while pairing the device falls asleep. Can you sniff the traffic while pairing it? https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html

kirovilya:
@Koenkk yes, i can. but pairing will be only from 2531, there is no native gateway

Koenkk:
[...]

PR #85, opened 2019-11-03
Support for  TERNCY devices.

kirovilya:
part of #82 
also proof https://github.com/fairecasoimeme/ZiGate/blob/master/Module%20Radio/Firmware/src/sdk/JN-SW-4170/Components/ZCIF/Source/zcl.c#L782


Koenkk:
Thanks, I will merge this after we completed supported for this device (#82)

kirovilya:
thank you!

PR #90, opened 2019-11-14
Touchlink

Koenkk:


sjorge:
What spec are you basing this off? I googled a bit and it seems to be part of ZLL but did not find much details of how it is implemented.

Koenkk:
It is implemented through a combination of cluster commands and interpan communication, documentation is indeed not very good. I reverse engineered most of the things atm.

sjorge:
OK so it wasn't just my google foo failing me then, good luck

Sineos:
Interesting reading material:
[wisec2017_touchlink.pdf](https://github.com/Koenkk/zigbee-herdsman/files/3863141/wisec2017_touchlink.pdf)

PR #103, opened 2019-12-03
2 attempts to get a list of active endpoints

kirovilya:
It is very difficult to pair new Aqara Opple devices lumi.remote.b286opcn01, lumi.remote.b486opcn01, lumi.remote.b686opcn01...

Interview crashes when receiving a list of the endpoints. Sometimes it goes fine, sometimes not.
https://github.com/Koenkk/zigbee2mqtt/issues/2433

If make two attempts to get this list, then the chance of success is almost 100% (for me) :)
[zigbee-herdsman.log](https://github.com/Koenkk/zigbee-herdsman/files/3917320/zigbee-herdsman.log)

Perhaps for many of zdo-requests we need to make several attempts (not only getting active endpoints).

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -348,6 +348,7 @@ class Device extends Entity {
         if (!activeEndpoints) {
             throw new Error(`Interview failed because can not get active endpoints ('${this.ieeeAddr}')`);
         }
+ 
         // Some devices, e.g. TERNCY return endpoint 0 in the active endpoints request.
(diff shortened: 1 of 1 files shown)


Koenkk:
Why not use the approach of https://github.com/zigpy/zigpy/pull/266, wouldn't this be more stable?

kirovilya:
Because I got another error - request timeout... no result on request

kirovilya:
But the second request worked and already checked several times

Koenkk:
okay, thanks again :smile:

PR #105, opened 2019-12-08
Add "Read Attribute" support from zigbee specifications - Require for Legrand Netatmo support

cemonneau:
Only an example of what I had to modify to add the support for Legrand devices.
See https://github.com/Koenkk/zigbee2mqtt/issues/2399 for details

A (non clean yet/debug) code that needs to be added to devices.js:
```js
    {
        zigbeeModel: [' Shutter switch with neutral\u0000\u0000\u0000'],
        model: 'Shutter',
        vendor: 'Legrand',
        description: 'Netatmo single wired shutter switch',
        // the physical LED will be green when permit join is true, off otherwise and red when not linked
        supports: 'open, close, stop, position, tilt',
        fromZigbee: [
            // Devices can send an identify message when the configuration button is pressed
            // (behind the physical buttons)
            // Used on the official gateway to send to every devices an identify command (green)
            {
                cluster: 'genIdentify',
                type: ['attributeReport', 'readResponse'],
                convert: (model, msg, publish, options) => {
                    return {identify: 'configuration'};
                },
            },
            // fz.ignore_genIdentify,
            // support legrand security
            {
                cluster: 'genBasic',
                type: ['attributeReport', 'readResponse', 'readGlobal'],
                convert: (model, msg, publish, options) => {
                    if (msg.type === 'readGlobal' && msg.cluster === 'genBasic' && msg.data && msg.data['61440']){
                        console.log('===========> SEND');
                        const endpoint = msg.device.getEndpoint(1);
                        endpoint.readResponse('genBasic', msg.frame.Header.transactionSequenceNumber, {legrandUnknownResponse:23}, {manufacturerCode:0x1021});
                        /*endpoint.readResponse('genBasic', {0xf00: {value: 23, type: 35}}, {manufacturerCode:0x1021});*/
                    }
                },
            },
            // support binary report on moving state (supposed/to be more tested)
            {
                cluster: 'genBinaryInput',
                type: ['attributeReport', 'readResponse'],
                convert: (model, msg, publish, options) => {
                    console.log('msg', JSON.stringify(msg));
                    return {
                        moving: false
                    };
                },
            },
        ],
        toZigbee: [
            {
                key: ['identify'],
                convertSet: async (entity, key, value, meta) => {
                    if (!value.timeout) {
                        const effects = {
                            'blink3': 0x00,
                            'fixed': 0x01,
                            'blinkgreen': 0x02,
                            'blinkblue': 0x03,
                        };
                        // only works for blink3 & fixed
                        const colors = {
                            'default': 0x00,
                            'red': 0x01,
                            'green': 0x02,
                            'blue': 0x03,
                            'lightblue': 0x04,
                            'yellow': 0x05,
                            'pink': 0x06,
                            'white': 0x07,
                        };

                        const selectedEffect = effects[value.effect] | effects['blink3'];
                        const selectedColor = colors[value.color] | colors['default'];

                        const payload = {effectid: selectedEffect, effectvariant: selectedColor};
                        await entity.command('genIdentify', 'triggerEffect', payload, {});
                    } else {
                        // seems to have no impact, to remove
                        await entity.command('genIdentify', 'identify', {identifytime: 10}, {});
                    }
                    // await entity.command('genIdentify', 'triggerEffect', payload, getOptions(meta));
                },
            },
            {
                key: ['enableLedIfOn'],
                convertSet: async (entity, key, value, meta) => {
                    const endpoint = entity.getEndpoint(1);
                    const enableLedIfOn = !!value;
                    // TODO working on it
                    // endpoint.write('XXX legrand cluster', );

                },
            },
        ],
        meta: {configureKey: 1},
        configure: async (device, coordinatorEndpoint) => {
            // genBinaryInput : 0x000f
            // closuresWindowCovering : 0x0102 XXXX
            // genIdentify : 0x0003
            const endpoint = device.getEndpoint(1);
            await bind(endpoint, coordinatorEndpoint, ['genBinaryInput', 'closuresWindowCovering', 'genIdentify']);
            // await endpoint.read('genBasic', ['legrandUnknown']);
        },
    },
```

### DIFF
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -10,7 +10,7 @@ function attributeList(frame: ZclFrame): KeyValue {
             const attribute = frame.Cluster.getAttribute(item.attrId);
             payload[attribute.name] = item.attrData;
         } catch (error) {
-            payload[item.attrId] = item.attrData;
+            payload[item.attrId] = typeof item.attrData !== 'undefined' ? item.attrData : item.attrId;
         }
     }
 
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -2,7 +2,7 @@ import {ZclFrame} from '../../zcl';
 
 interface KeyValue {[s: string]: number | string};
 
-function attributeList(frame: ZclFrame): KeyValue {
+function attributeKeyValue(frame: ZclFrame): KeyValue {
     const payload: KeyValue = {};
 
     for (const item of frame.Payload) {
@@ -10,7 +10,22 @@ function attributeList(frame: ZclFrame): KeyValue {
             const attribute = frame.Cluster.getAttribute(item.attrId);
             payload[attribute.name] = item.attrData;
         } catch (error) {
-            payload[item.attrId] = typeof item.attrData !== 'undefined' ? item.attrData : item.attrId;
+            payload[item.attrId] = item.attrData;
+        }
+    }
+
+    return payload;
+}
+
+function attributeList(frame: ZclFrame): Array<string | number> {
+    const payload: Array<string | number> = [];
+
+    for (const item of frame.Payload) {
+        try {
+            const attribute = frame.Cluster.getAttribute(item.attrId);
+            payload.push(attribute.name);
+        } catch (error) {
+            payload.push(item.attrId);
         }
     }
 
@@ -18,5 +33,6 @@ function attributeList(frame: ZclFrame): KeyValue {
 }
 
 export {
+    attributeKeyValue,
     attributeList,
 };
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -46,7 +46,6 @@ const Cluster: {
             appProfileVersion: {ID: 8, type: DataType.enum8},
 
             legrandUnknown: {ID: 0x00a, type: DataType.octetStr},
-            legrandUnknownResponse: {ID: 0xf00, type: DataType.uint32},
 
             swBuildId: {ID: 16384, type: DataType.charStr},
             locationDesc: {ID: 16, type: DataType.charStr},
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -3530,6 +3530,17 @@ const Cluster: {
         commandsResponse: {
         }
     },
+    manuSpecificLegrandDevices: {
+        ID: 0xfc01,
+        manufacturerCode: ManufacturerCode.LegrandNetatmo,
+        attributes: {
+            // attributes seems to vary depending on the device. Can't be static
+        },
+        commands: {
+        },
+        commandsResponse: {
+        }
+    },
 };
 
 export default Cluster;
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -44,9 +44,6 @@ const Cluster: {
             dateCode: {ID: 6, type: DataType.charStr},
             powerSource: {ID: 7, type: DataType.enum8},
             appProfileVersion: {ID: 8, type: DataType.enum8},
-
-            legrandUnknown: {ID: 0x00a, type: DataType.octetStr},
-
             swBuildId: {ID: 16384, type: DataType.charStr},
             locationDesc: {ID: 16, type: DataType.charStr},
             physicalEnv: {ID: 17, type: DataType.enum8},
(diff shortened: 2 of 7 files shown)


cemonneau:
The second commit allows to replace the line 
```js
endpoint.readResponse('genBasic', msg.frame.Header.transactionSequenceNumber, {legrandUnknownResponse:23}, {manufacturerCode:0x1021});
```
by
```js
endpoint.readResponse('genBasic', msg.frame.Header.transactionSequenceNumber, {0xf00: {value: 23, type: 35}}, {manufacturerCode:0x1021});
```

A bit more permissive and removes the dependency on cluster.ts for specific trames

src/zcl/definition/cluster.ts
```diff
@@ -44,6 +44,9 @@ const Cluster: {
             dateCode: {ID: 6, type: DataType.charStr},
             powerSource: {ID: 7, type: DataType.enum8},
             appProfileVersion: {ID: 8, type: DataType.enum8},
+
+            legrandUnknown: {ID: 0x00a, type: DataType.octetStr},
```
Koenkk: Is this still needed after your second commit?

Koenkk:
To respond to the read requests please use the onEvent instead of a converter. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L715

src/controller/helpers/zclFrameConverter.ts
```diff
@@ -10,7 +10,7 @@ function attributeList(frame: ZclFrame): KeyValue {
             const attribute = frame.Cluster.getAttribute(item.attrId);
             payload[attribute.name] = item.attrData;
         } catch (error) {
-            payload[item.attrId] = item.attrData;
+            payload[item.attrId] = typeof item.attrData !== 'undefined' ? item.attrData : item.attrId;
```
Koenkk: Why is this needed?

src/zcl/definition/cluster.ts
Koenkk: I see why it's needed now, I would like to provide the ID instead (so that there is no need to map it here). Can you provide the herdsman debug log when receiving this message?

src/controller/helpers/zclFrameConverter.ts
Koenkk: This shouldn't be needed anymore after my changes, the `read` data payload will now be an array instead of an object with an `undefined` value.

Koenkk:
I've made some breaking changes:

- removed the frame from payload (I do not want to expose this), use `meta.zclTransactionSequenceNumber` now
- I've removed the cluster legrand attribute (should not be needed now anymore)
- The payload of a read is now an array instead of an object with an `undefined` value.

Let me know if this is OK for you then I will merge this.

src/zcl/definition/cluster.ts
cemonneau: After more tests it's removable. I was seeing those frames on the official gateway but it's only if i poll them. still haven't found what the purpose is ...

cemonneau:
Thanks for your fast work, amazing ! 
I will convert to the event binding, when I wrote the prototype I wasn't honestly interested in "beautifull" code, only working one. I found about that binding today, thanks for the reminder !
I will report back on the progress with those modifications

cemonneau:
Ok so, I made the modifications and tested again. 
It works, but I got once the same event twice, which led to the same readResponse been sent twice too. The device didn't paired (left). Not sure why it triggered twice, only happened once.

I added a the same time a new commit on top of yours that adds Legrand manufacturer & cluster (settings) required for the full support. If you want me to make a separate PR I will undo it.

In the current state I think it's enough for that PR BUT regarding the zigbee specifications I think the code should send (I don't know how to implement it in the current state) the default response "not supported" (that's what other devices answer to the device when I made my tests).

Thanks again!

Koenkk:
What do you mean with the default response stuff? Zigbee2mqtt should already do this

cemonneau:
When i was testing, the switch was doing a read attribute call (as "expected") but the coordinator (latest version of everything) wasn't answering anything

Koenkk:
That should be done here: https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/controller.ts#L490, please try to debug and find out why it's not going there.

I will merge this PR as it already is a step in the good direction.

cemonneau:
Hi,

I tested again today with/without the custom device.
When the device sends a read attribute, the coordinator doesn't answser when the custom code is not there (device.js). But the frame contains "disable default reponse=true" which might explain why https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/controller.ts#L498 is not called, and would probably not send the anwser that should be sent according to the Zigbee specs : attribute not supported (on read attribute).

Not sure if the default answer ask to disable that part through. There is still a warning in the log now saying "unsupported XXX". I think that's enough for now.

Koenkk:
According to the zigbee spec there should not be a default response when its disabled. As a workaround you can add it to `devices.js`.

Koenkk:
According to the zigbee spec there should not be a default response when its disabled. As a workaround you can add it to `devices.js`.

Eniot666:
hi,
this post has something to do with my problem : https://github.com/Koenkk/zigbee2mqtt/discussions/21502 ?

PR #111, opened 2019-12-19
Add command "moveColorTemp" for ZBT-CCTSwitch-D0001

qosmio:
This command seems to be missing, even though
a few converters (aqara remote) seem to reference
it.

This is used for the "hold"/"release" for controlling color temp for this remote.


Koenkk:
Thanks!

PR #114, opened 2019-12-23
Add optional zcl status result to endpoint write and command actions

allofmex:
Fixed  (`functional-command`) and added (`foundation-write`) zcl status codes (`success`, `unsupportedCluster`,..) in device results ( needed by ioBroker's Developer-Tool).

Currently it's impossible to see if a command was successfull or had error, because response is always empty in case of `foundation-write` or `functional-command`:
grafik

This PR adds/fixes `statusCode` property in response.
grafik

Note for zclFrameConverter.ts:
<del>I just added a "global" status field, but this may fail if there are multiple attrId's present. 
```
REMOVED
```
 A better implementation would be to have individual status per item.attrId, but **this would require to move item.attrData to child property (breaks compatibility)**
```
{   attrId: {
        data: attrData,
        statusCode: 0 
    }, 
    attrId: {
        data: attrData,
        statusCode: 195 }
```

### DIFF
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -217,7 +217,7 @@ class ZStackAdapter extends Adapter {
     ): Promise<Events.ZclDataPayload> {
         return this.queue.execute<Events.ZclDataPayload>(async () => {
             const command = zclFrame.getCommand();
-            if (!command.hasOwnProperty('response')) {
+            if (!command.hasOwnProperty('response') && zclFrame.Header.frameControl.disableDefaultResponse) {
                 throw new Error(`Command '${command.name}' has no response, cannot wait for response`);
             }
 
@@ -747,9 +747,12 @@ class ZStackAdapter extends Adapter {
             payload.endpoint === matcher.endpoint &&
             (!matcher.transactionSequenceNumber || transactionSequenceNumber === matcher.transactionSequenceNumber) &&
             payload.frame.Cluster.ID === matcher.clusterID &&
-            matcher.frameType === payload.frame.Header.frameControl.frameType &&
-            matcher.commandIdentifier === payload.frame.Header.commandIdentifier &&
-            matcher.direction === payload.frame.Header.frameControl.direction;
+            (
+                (matcher.commandIdentifier === payload.frame.Header.commandIdentifier &&
+                matcher.direction === payload.frame.Header.frameControl.direction) ||
+                (payload.frame.Header.commandIdentifier === 0x0b && // defaultResponse device->coordinator
+                payload.frame.Header.frameControl.direction === Direction.SERVER_TO_CLIENT)
+            );
     }
 }
 
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -12,6 +12,7 @@ function attributeKeyValue(frame: ZclFrame): KeyValue {
         } catch (error) {
             payload[item.attrId] = item.attrData;
         }
+        payload["statusCode"] = item.status;
     }
 
     return payload;
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -1,18 +1,25 @@
 import {ZclFrame} from '../../zcl';
 
-interface KeyValue {[s: string]: number | string};
+interface KeyValue {[s: string]: {data: number | string, statusCode: number}};
 
-function attributeKeyValue(frame: ZclFrame): KeyValue {
+function attributeKeyValue(frame: ZclFrame, flat=true): KeyValue {
     const payload: KeyValue = {};
 
     for (const item of frame.Payload) {
         try {
             const attribute = frame.Cluster.getAttribute(item.attrId);
-            payload[attribute.name] = item.attrData;
+            if (flat) {
+                payload[attribute.name] = item.attrData;
+            } else {
+                payload[attribute.name] = {data:item.attrData, statusCode: item.status};
+            }
         } catch (error) {
-            payload[item.attrId] = item.attrData;
+            if (flat) {
+                payload[item.attrId] = item.attrData;
+            } else {
+                payload[item.attrId] = {data: item.attrData, statusCode: 1}; // 1=failure
+            }
         }
-        payload["statusCode"] = item.status;
     }
 
     return payload;
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -383,7 +383,7 @@ class Device extends Entity {
             for (const chunk of ArraySplitChunks(Object.keys(Device.ReportablePropertiesMapping), 3)) {
                 const result = await endpoint.read('genBasic', chunk);
                 for (const [key, value] of Object.entries(result)) {
-                    Device.ReportablePropertiesMapping[key].set(value, this);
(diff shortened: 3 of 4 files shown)


src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -217,7 +217,7 @@ class ZStackAdapter extends Adapter {
     ): Promise<Events.ZclDataPayload> {
         return this.queue.execute<Events.ZclDataPayload>(async () => {
             const command = zclFrame.getCommand();
-            if (!command.hasOwnProperty('response')) {
+            if (!command.hasOwnProperty('response') && zclFrame.Header.frameControl.disableDefaultResponse) {
```
Koenkk: I don't understand this, a default response is something different than a response to a command, I don't think this is right.

allofmex: In case of error (example read/write to not supported cluster/attribute,...), the device responds with a defaultResponse.
See third example here: https://www.digi.com/support/forum/2086/default-response-for-zcl (second post)

The current matcher does not hit, because the commandId and direction of B->A defaultResponse is different than the read/write frame A->B was. So in current implementation, the NON-SUCCESS-defaultResonse can not be matched to a pending transaction.

Koenkk: When a device doesn't support a certain attribute, it should respond with a write attributes response. Below is a screenshot when I'm trying to write `genBinaryInput` `activeText` attribute to a TRADFRI 5 button dimmer.

image

allofmex: > When a device doesn't support a certain attribute, it should respond with a write attributes response

Correct. But now read or write an unsupported **cluster**:

grafik

allofmex:
I updated the PR since the old caused some problems. This one should be cleaner.

Some examples:
### Old implementation
(results of `endpoint.read`)
**Foundation**:
successfull: `{"zclVersion":1}`
unsupportedCluster: `"Failed to publish: Timeout - 32811 - 3 - 2 - 9 - 1 after 10000ms"` 
unsupportedAttribute: `{}`
1 supported Attribute, 1 unsupported: `{"zclVersion":2}}`
**Functional**
(results of `endpoint.command`)
successfull: `{}`
failure: `{}`

### With this PR
**Foundation**:
successfull:` {"zclVersion":{"data":1,"statusCode":0}}`
unsupportedAttribute: `{"alarmMask":{"statusCode":134}}`
1 supported Attribute, 1 unsupported: `{"zclVersion":{"data":2,"statusCode":0},"alarmMask":{"statusCode":134}}`
unsupportedCluster: `{"error":"Read failed","statusCode":195}`

**Functional**:
successfull: `{"cmdId":3,"statusCode":0}`
unsupportedCluster:` {"cmdId":0,"statusCode":195}`
unsupClusterCmd:` {"cmdId":1,"statusCode":129}`

allofmex:
**The new implementation requires small change in dependent projects**, since read result is now an Object with property "data" `{"zclVersion":{"data":1,"statusCode":0}}` instead data direct `{"zclVersion":1}`
Sorry for this, but herdsman should really have error codes submitted. This will allow some nice features like auto-generated-converters-for-unsupported-devices..

src/adapter/z-stack/adapter/zStackAdapter.ts
Koenkk: I still cannot reproduce this, with the following device:
```
{"id":8,"type":"Router","ieeeAddr":"0x0017880104c9cd33","nwkAddr":49209,"manufId":4107,"manufName":"Philips","powerSource":"Mains (single phase)","modelId":"LWB010","epList":[11,242],"endpoints":{"11":{"profId":49246,"epId":11,"devId":256,"inClusterList":[0,3,4,5,6,8,4096],"outClusterList":[25],"clusters":{"genBasic":{"attributes":{"modelId":"LWB010","manufacturerName":"Philips","powerSource":1,"zclVersion":1,"appVersion":2,"stackVersion":1,"hwVersion":1,"dateCode":"20181115","swBuildId":"1.46.13_r26312"}},"seMetering":{"attributes":{}}},"binds":[]},"242":{"profId":41440,"epId":242,"devId":97,"inClusterList":[33],"outClusterList":[33],"clusters":{},"binds":[]}},"appVersion":2,"stackVersion":1,"hwVersion":1,"dateCode":"20181115","swBuildId":"1.46.13_r26312","zclVersion":1,"interviewCompleted":true,"meta":{},"lastSeen":1578592996991}
```

I read the following value from an unsupported cluster:

```js
const device = this.resolveEntity("0x0017880104c9cd33");
device.endpoint.read('msTemperatureMeasurement', ['measuredValue']);
```

As you can see, it doesn't support this cluster, `msTemperatureMeasurement` = `1026`.

And I get the following response:

image

allofmex: Did some tests (disableDefaultResponse=true):
## readAttribute to not existing cluster


- PhilipsHue, Ikea, ZBT(MuellerLicht) respond with `ReadAttributeResponse (status: unsupAttribute)`

- Osram  responds with `DefaultResponse (status: failure)`

- Eurotronics responds with `DefaultResponse (status: unsupCluster)`

## readAttribute to existing cluster but not existing Attribute

- all respond with `readAttrResp` with status unsupAttr)

## writeAttribute to existing cluster but read-only Attribute

- all respond with `readAttrResp` with status readOnly)

### Summary
A response in case of error can be both, normal `(read/..)AttributeResponse` or `DefaultResponse`, even if disableDefaultResponse=true. Depends again on manufacturer..

Koenkk: Thanks, I will run some local tests but it may take some time.

allofmex: Thanks. No problem. I'm also ok if you prefer to implement it in another way. PR is just a suggest. It's only important, that the error/status codes reach the application.

Koenkk: What Osram device did you try with? I've tried the code of  https://github.com/Koenkk/zigbee-herdsman/pull/114#discussion_r364885878 with https://www.zigbee2mqtt.io/devices/AC0251100NJ_AC0251700NJ.html  but it responds the same as https://github.com/Koenkk/zigbee-herdsman/pull/114#discussion_r364885878

allofmex: This one
https://www.zigbee2mqtt.io/devices/AB3257001NJ.html

Koenkk: So it seems that the amount of devices affected by this is limited, how big of a problem is this?

allofmex: In my opinion it is important to handles this kind of response as well. The mentioned Osram Plug for instance is a very popular device (cheap and widely available). And as of the digi.com support link above, it even seems to be the correct behavior, to reply with defaultResponse in case of error.

Another suggest:
The condition  `if (!command.hasOwnProperty('response')... throw new Error(Command .. has no response, cannot wait for response)` may be removed completely, since there can always be a response (namely in case of error), even for commands that normally do not have a response.

Then we only need to make sure that `waitressValidator` matches **any** response, that results from our sent request (same sequence number), no matter if readAttrResp or defaultResp (implemented in PR already)

Koenkk: Yes I understand, I will buy this device so I can investigate further. thanks!

PR #115, opened 2019-12-26
Fix parsing of payload for configReportRsp

carldebilly:
The parsing of the payload for `configReportRsp` was incorrect and leading to `raw` packet type in `zigbee2mqtt` as defined in this comment: https://github.com/Koenkk/zigbee2mqtt/issues/2058#issuecomment-568941285

✔️ Also include mfgr clusters for Stelpro

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -1717,7 +1717,7 @@ const Cluster: {
             SinopeOccupancy: {ID: 1024, type: DataType.enum8, manufacturerCode: ManufacturerCode.Sinope},
             SinopeBacklight: {ID: 1026, type: DataType.enum8, manufacturerCode: ManufacturerCode.Sinope},
             StelproSystemMode: {ID: 0x401c, type: DataType.enum8},
-            StelproOutdoorTemp: { ID: 0x4001, type: DataType.int16, manufacturerCode: ManufacturerCode.Stelpro },
+            StelproOutdoorTemp: {ID: 0x4001, type: DataType.int16, manufacturerCode: ManufacturerCode.Stelpro},
         },
         commands: {
             setpointRaiseLower: {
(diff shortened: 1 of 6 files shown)


carldebilly:
Reference payload for the new test:
image

carldebilly:
@Koenkk I suspect the failing test is incorrect.  Also I'm wordering if the parsing should be `"flat"` instead of `"repetitive"`...

Koenkk:
@carldebilly thanks for the PR, unfortunately I'm on holiday now, will dive into it after.

Koenkk:
The reason why this was here is: 

image

However I'm not sure what it exactly means, need to check with some devices.

carldebilly:
You're right: it's breaking somewhere else.
I think the right fix is to switch the command parsing strategy to `flat`.

carldebilly:
@Koenkk Ok I fixed the problem.

Now the payload is "optional" but when present, it won't crash the parsing as before.

src/zcl/definition/cluster.ts
```diff
@@ -1716,6 +1716,8 @@ const Cluster: {
             acCapacityFormat: {ID: 71, type: DataType.enum8},
             SinopeOccupancy: {ID: 1024, type: DataType.enum8, manufacturerCode: ManufacturerCode.Sinope},
             SinopeBacklight: {ID: 1026, type: DataType.enum8, manufacturerCode: ManufacturerCode.Sinope},
+            StelproSystemMode: {ID: 0x401c, type: DataType.enum8},
```
Koenkk: Can we also add `manufacturerCode: ManufacturerCode.Stelpro` here?

carldebilly: From what I saw it's not using the manufacturer code there...  but maybe I should check again to be sure.

carldebilly: Anyway this mode is not well supported by the converter right now. The thermostat is supporting an _eco_ mode which is not implemented.

Koenkk:
The parsing should definitely repetitive, otherwise when multiple status records are returned only this first one will be parsed. I've updated the PR to do this, test are OK now.

@carldebilly let me know if this is ok for you then this can be merged.

carldebilly:
I'm confortable with those changes.

Koenkk:
Awesome, thanks!

PR #117, opened 2019-12-27
fix api docs src links

hobbyquaker:
Somehow the links to the source files in the API docs got messed up (doubled `/src/src` in the URLs), possibly because tsdoc uses the current working directory and `npm run docs` was executed in the src dir? Anyway, this PR includes recreated docs with correct src links.
Best Regards, Sebastian


Koenkk:
Thanks!

PR #119, opened 2020-01-02
Add new cluster for Drayton Wiser

arcanefoam:


src/zcl/definition/cluster.ts
```diff
@@ -3541,6 +3541,14 @@ const Cluster: {
         commandsResponse: {
         }
     },
+    draytonDeviceInfo: {
```
sjorge: Maybe rename this to `wiserDeviceInfo`, my test unit branded as 'Eberle Wiser' also reports with the exact same database entry.

I looks like both Drayton and Eberle are local rebrands from the Wiser device (which seems to be made by Schnieder)

Koenkk:
Thanks!

Issue #121, opened 2020-01-07
[Request] Include type definitions


anthonynichols:
First of all, thank you for the work on this library. I was considering picking and continuing work on the previously stale/abandoned libraries that influenced this one myself, so it is great to see active development from others in the community. Secondly, thank you for using Typescript! As a fellow Typescript user, I was glad to see `.ts` files in the `src` directory.

Could you generate type definitions and publish them along with your transpiled files? I would love to be able to use this in my Typescript project and have all of the type information.

Thanks!

Koenkk:
@anthonynichols I haven't looked into this but I don't know how to do it, probably you have more experience and can do this faster, could you make a pr?

anthonynichols:
Sure, I'd be happy to. PR incoming...

anthonynichols:
https://github.com/Koenkk/zigbee-herdsman/pull/123

PR #123, opened 2020-01-09
Include type definitions in build

anthonynichols:
Fixes https://github.com/Koenkk/zigbee-herdsman/issues/121

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,8 @@
     "cc253x"
   ],
   "license": "MIT",
-  "main": "index.js",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "name": "zigbee-herdsman",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -73,7 +74,7 @@
     "build": "./node_modules/.bin/tsc",
     "start": "./node_modules/.bin/tsc -w",
     "test": "jest test",
-    "test-with-coverage": "jest test --coverage --collectCoverageFrom '**/*.{ts,tsx}'",
+    "test-with-coverage": "jest test --coverage --collectCoverageFrom 'src/**/*.ts'",
```
anthonynichols: `test-with-coverage` was failing for me, and I don't think there are any React `.tsx` files, so I changed this to only look at `.ts` files in the `src` directory.

tsconfig.json
```diff
@@ -6,18 +6,14 @@
         "noImplicitAny": true,
         "moduleResolution": "node",
         "sourceMap": true,
+        "declaration": true,
+        "declarationMap": true,
         "outDir": "dist",
         "baseUrl": ".",
-        "allowJs": true,
```
anthonynichols: There aren't any `.js` files being transpiled by TypeScript, so this can be removed.

tsconfig.json
```diff
@@ -6,18 +6,14 @@
...
         "resolveJsonModule": true,
-        "paths": {
-            "*": [
-                "node_modules/*"
-            ]
-        }
```
anthonynichols: This is unnecessary as the default module resolution strategy for `node` is to look first in `node_modules/*`.

Koenkk:
Thanks for all the improvements!

PR #124, opened 2020-01-09
Add manufacturerCode in meta

cemonneau:
Required for https://github.com/Koenkk/zigbee-herdsman-converters/pull/891
It allows to filter in zigbee-herdsman-converter based on the manufacturer for specific implementations when the device is not fully paired


Koenkk:
Thanks!

PR #126, opened 2020-01-12
Add mandatory filed timeStatus for getTime cluster

dmpolukhin:
Also add localTime for devices that require it.


src/controller/controller.ts
```diff
@@ -509,8 +509,13 @@ class Controller extends events.EventEmitter {
             // Reponse to time reads
             if (frame.isGlobal() && frame.isCluster('genTime') && frame.isCommand('read')) {
                 const time = Math.round(((new Date()).getTime() - OneJanuary2000) / 1000);
+                const localTime = time + (new Date()).getTimezoneOffset() * 60;
                 try {
-                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, {time});
+                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, {
```
Koenkk: Although the attributes are mandatory, it doesn't mean they always have to be returned.

In order to determine what attributes need to be responded to, please check `dataPayload.frame.Payload` (it contains the attributes the device wants to read.

dmpolukhin: Done, please take another look.

Koenkk: Looks good, I will add test cases and merge it.

Koenkk:
Thanks!

DerDreschner:
@dmpolukhin : You've set the `timeStatus` to 3, which indicates master time and synchronized. This is a combination that isn't possible if the device is following the ZCL spec as "master time" indicates it has, e.g., an internal real-time clock which means that the time can't be written via Zigbee, and "synchronized" means that the time on the device was written via Zigbee.

Did you experience any issues with other values (`1` for example) or did you just assume that "synchronized" means "synchronized via external time servers over the internet"?

dmpolukhin:
> @dmpolukhin : Did you experience any issues with other values (`1` for example) or did you just assume that "synchronized" means "synchronized via external time servers over the internet"?

Unfortunately I don't remember and I don't have this device anymore to test.

PR #128, opened 2020-01-18
Support for getWeeklyScheduleRsp

dmpolukhin:


Koenkk:
I've implemented the complete transition parsing in herdsman now, please review https://github.com/Koenkk/zigbee-herdsman/pull/128/files#diff-8158e442a03e3ab9574aeb1a2259d330R282 and then this can be merged.

src/zcl/definition/cluster.ts
```diff
@@ -1761,7 +1761,7 @@ const Cluster: {
                     {name: 'numoftrans', type: DataType.uint8},
                     {name: 'dayofweek', type: DataType.uint8},
                     {name: 'mode', type: DataType.uint8},
-                    {name: 'thermoseqmode', type: BuffaloZclDataType.LIST_UINT16},
+                    {name: 'transitions', type: BuffaloZclDataType.LIST_THERMO_TRANSITIONS},
```
dmpolukhin: I might be wrong but I think now we also need write transition. For LIST_UINT16 it was already implemented  but now it is new type...

dmpolukhin: To be more precise, I think in setWeeklySchedule makes sense to use the same type and in this case write transformation will be required.

Koenkk:
Write is implemented, please review and this can be merged.

test/zcl.test.ts
```diff
@@ -279,6 +279,34 @@ describe('Zcl', () => {
...
+        expect(frameCool.Payload).toStrictEqual({numoftrans:6, dayofweek:64, mode:2, transitions: [{transitionTime:360,coolSetpoint:2300},{transitionTime:570,coolSetpoint:2200},{transitionTime:720,coolSetpoint:2150},{transitionTime:840,coolSetpoint:2150},{transitionTime:990,coolSetpoint:2300},{transitionTime:1380,coolSetpoint:2100}]});
+
+        const bufferHeatAndCool = [9, 7, 0, 1, 64, 3, 104, 1, 252, 8, 58, 2];
+        const frameHeatAndCool = Zcl.ZclFrame.fromBuffer(Zcl.Utils.getCluster("hvacThermostat").ID, Buffer.from(bufferHeatAndCool));
+        expect(frameHeatAndCool.Payload).toStrictEqual({numoftrans:1, dayofweek:64, mode:3, transitions: [{transitionTime:360,coolSetpoint:570, heatSetpoint: 2300}]});
+    });
+
+    it('ZclFrame to buffer setWeeklyScheduleRsp (hvacThermostat)', () => {
+        const payloadHeat = {numoftrans:6, dayofweek:64, mode:1, transitions: [{transitionTime:360,heatSetpoint:23},{transitionTime:570,heatSetpoint:22},{transitionTime:720,heatSetpoint:21.5},{transitionTime:840,heatSetpoint:21.5},{transitionTime:990,heatSetpoint:23},{transitionTime:1380,heatSetpoint:21}]};
```
dmpolukhin: I think we should be consistent about units (i.e. in degrees or in degree * 100). In read tests I see that you removed / 100 but in write test you expect temperature divided by 100.

Koenkk: Forgot to update, fixed now.

Koenkk:
great, thanks!

PR #140, opened 2020-02-22
Increase concurrency to 16 in zStackAdapter

ex3ndr:
I was in deep investigation why my network is sooo unresponsive some time, but it turns out that the bottleneck is concurrency in Queue in zStackAdapter.

Most of the operations are done almost immediately, but some are timeouts after 10(?) second that blocks this queue without any actual work done by controller.

Increasing this number made a huge impact on my setup and everything is super responsive.

I am not sure why this limitation was here in the first place, but with new CC26x2R1 chip it is possible to increase this concurrency dramatically.

### DIFF
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -62,7 +62,6 @@ class ZStackAdapter extends Adapter {
 
         this.transactionID = 0;
         this.closing = false;
-        this.queue = new Queue(16);
         this.waitress = new Waitress<Events.ZclDataPayload, WaitressMatcher>(
             this.waitressValidator, this.waitressTimeoutFormatter
         );
@@ -90,6 +89,8 @@ class ZStackAdapter extends Adapter {
             debug(`Failed to get zStack version, assuming 1.2`);
             this.version = {"transportrev":2, "product":0, "majorrel":2, "minorrel":0, "maintrel":0, "revision":""};
         }
+
+        this.queue = new Queue(this.version.product === ZnpVersion.zStack3x0 ? 16 : 2);
 
         debug(`Detected znp version '${ZnpVersion[this.version.product]}' (${JSON.stringify(this.version)})`);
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Please add a check if this.version = zStack3x0 here, we dont want to increase this for the cc2531 e.g.

ex3ndr:
Done!

Koenkk:
Thanks!

james-fry:
> I was in deep investigation why my network is sooo unresponsive some time, but it turns out that the bottleneck is concurrency in Queue in zStackAdapter.
> 
> Most of the operations are done almost immediately, but some are timeouts after 10(?) second that blocks this queue without any actual work done by controller.
> 
> Increasing this number made a huge impact on my setup and everything is super responsive.
> 
> I am not sure why this limitation was here in the first place, but with new CC26x2R1 chip it is possible to increase this concurrency dramatically.

I hope this solves the "delay" issue I see in my network. Thanks!

gev:
What about cc2538? Which value will be good for?

Koenkk:
I guess it would also work for that, but currently we cannot distinguish between zStack 3 CC2531 and a zStack 3 CC2538, this has to be fixed in the firmware. @reverieline Could you set the product to `4` in the firmware?

reverieline:
> 
> 
> I guess it would also work for that, but currently we cannot distinguish between zStack 3 CC2531 and a zStack 3 CC2538, this has to be fixed in the firmware. @reverieline Could you set the product to `4` in the firmware?

Yes, shure.

gev:
I've check queue length 16 for cc2538. All works as well!

gev:
I suggest add queue length into controller constructor  options

gev:
Also, think, better to use `productId ` instead znp version

ex3ndr:
Can we make this configurable? I really want to play around with different values, it seems that such limitation is not needed at all.

gev:
Make concurrent configurable: https://github.com/Koenkk/zigbee-herdsman/pull/159

gev:
need a help, can't understand why coverage tests have been failed

PR #144, opened 2020-02-26
Add device.getEndpointByDeviceId() method

nixent:
Zigbee devices might have multiple endpoints in the same device. In some cases, like Gledopto GL-C-007 and GL-C-008, rgb and white endpointIds are not consistent from device to device and `gledopto_light_onoff_brightness/gledopto_light_color_colortemp_white` methods are trying to identify which endpoint is rgb and which one is white/dimmable based on known combinations of endpointIds. With `device.getEndpointByDeviceId()` method endpoint identification can be simplified and will eliminate need to know all possible endpointId combinations.

`Endpoint.deviceId` property reflects type/functionality as per Zigbee specification. ZLL specification defines following values for deviceId:
```
const zllDeviceId = {
    'onOffLight': 0x0000,
    'onOffPluginUnit': 0x0010,
    'dimmableLight': 0x0100,
    'dimmablePluginUnit': 0x0110,
    'colorLight': 0x0200,
    'extColorLight': 0x0210,
    'colorTempLight': 0x0220,
};
```

Usage example:

```
rgbEndpoint = meta.device.getEndpointByDeviceId(zllDeviceId.extColorLight);
if (rgbEndpoint === undefined){
    rgbEndpoint = meta.device.getEndpointByDeviceId(zllDeviceId.colorLight);
}

```

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -138,9 +138,9 @@ class Device extends Entity {
     }
 
     // There might be multiple endpoints with same DeviceId but it is not supported and first endpoint is returned
-    public getEndpointByDeviceType(devType: string): Endpoint {
-        const devId = Zcl.EndpointDeviceType[devType];
-        return this.endpoints.find((d): boolean => d.deviceID === devId);
+    public getEndpointByDeviceType(deviceType: string): Endpoint {
+        const deviceID = Zcl.EndpointDeviceType[deviceType];
+        return this.endpoints.find((d): boolean => d.deviceID === deviceID);
     }
 
     public updateLastSeen(): void {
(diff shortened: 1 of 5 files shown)


src/controller/model/device.ts
```diff
@@ -137,6 +137,11 @@ class Device extends Entity {
         return this.endpoints.find((e): boolean => e.ID === ID);
     }
 
+    // There might be multiple endpoints with same DeviceId but it is not supported and first endpoint is returned
+    public getEndpointByDeviceId(devId: number): Endpoint {
```
Koenkk: Instead of a number, would it be nicer to pass a string down here? e.g. `onOffLight` instead of `0x0000`. The lookup can be added here: https://github.com/Koenkk/zigbee-herdsman/tree/master/src/zcl/definition

nixent: Yes, string lookup would be more consistent with the rest of the code

Koenkk:
Thanks!

PR #149, opened 2020-03-02
Alias data8 with uint8 (instead of int8)

felixstorm:
The ZCL specification does IMHO not indicate whether `data8` should be signed or unsigned, but as it is supposed to be used for discrete values and `data16`, `data24` etc. are all treated as unsigned, this PR also changes `data8` to unsigned.


Koenkk:
Makes sense, thanks!

PR #153, opened 2020-03-12
Revert "deconz adapter for zigbee-herdsman (beta)"

Koenkk:
Reverts Koenkk/zigbee-herdsman#152


Koenkk:
@ChrisHae the PR makes the build fail, I will check what's still missing but for now I revert it.

PR #154, opened 2020-03-17
Deconz adapter - permit join fix

ChrisHae:
Test fix for permit join
Added writeparameter request to permit join request


Koenkk:
Thanks!

PR #155, opened 2020-03-18
Fix genTime reads

presslab-us:
My sprinkler controller requests time from the coordinator.  I noticed the clock was off by +3 hrs.  I tracked it down to this fix.

The `Date` constructor was using an implied offset, so explicitly stating it fixed `OneJanuary2000`.

And `getTimezoneOffset()` has the opposite sign of what you'd expect, the documentation states: "_The time-zone offset is the difference, in minutes, from local time to UTC. Note that this means that the offset is positive if the local timezone is behind UTC and negative if it is ahead._"

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -55,7 +55,7 @@ const debug = {
 /**
  * @ignore
  */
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks for fixing 👍

PR #156, opened 2020-03-22
Support OTA ImagePageRequest, add EP 47 for Insta/Gira/Jung OTA

felixstorm:
OTA update for the Insta/Gira/Jung wall switch needs ImagePageRequest (in addition to ImageBlockRequest). There will also shortly be a matching PR for the converters containing the OTA implementation supporting ImagePageRequest.

Also, the wall switch is buggy when it tries to find a matching EP for OTA: It queries for ZLL profile, but then contacts with HA profile. Therefore I added EP 47 for the coordinator since the wall switch seems to contact this EP as a fallback.


Koenkk:
Thanks!

PR #162, opened 2020-03-31
Handle the command 0x02 of the cluster manuSpecificTuyaDimmer

dzegarra:
Tuya uses the following commands:

* 0x00 Commands send to the ZED from the Zigbee coordinator
* 0x01 Reporting state of the ZED to the Zigbee coordinator
* 0x02 Reporting state of the ZED to the Zigbee coordinator **after the coordinator sent a 0x00 command**

I found this after sniffing the comm between my Siterwell TRVs and my Tuya gateway. The implementation of the Siterwell GS361 radiator valves has its [own Pull Request]( https://github.com/Koenkk/zigbee-herdsman-converters/pull/1120).


Koenkk:
Thanks!

PR #167, opened 2020-04-16
Patch for cases when hasConfigured does not exists

kirovilya:


Koenkk:
Fixed with https://github.com/Koenkk/zigbee-herdsman/commit/c68d43aca38da06d4438c1b3a3f551e9dbbb8ed9

PR #177, opened 2020-05-12
Updated zclFrameToBuffer

ChrisHae:
Use given .toBuffer() function to transform zclPayload to buffer instead of own implementation.


Koenkk:
Thanks!

PR #178, opened 2020-05-18
Improve ZNP constants and typings

r0b1n:


src/adapter/z-stack/constants/common.ts
```diff
@@ -315,4 +196,89 @@ const COMMON: {
     }
 };
 
-export default COMMON;
\ No newline at end of file
+export enum ZnpCommandStatus {
+    SUCCESS = 0x00,
+    FAILURE = 0x01,
+    INVALID_PARAM = 0x02,
+    NV_ITEM_UNINIT = 0x09,
```
Koenkk, line 203: Wouldn't `NV_ITEM_INITIALIZED` be a better name?

From the docs:
> 0x09 = Success, item created and initialized

src/adapter/z-stack/constants/common.ts
```diff
@@ -315,4 +196,89 @@ const COMMON: {
     }
 };
 
-export default COMMON;
\ No newline at end of file
+export enum ZnpCommandStatus {
```
Koenkk, line 199: Why was this moved out of `COMMON` (in my opinion it makes it inconsistent with the rest).

Koenkk:
Thanks for the PR, this is a good improvement!

Added some questions/comments.

src/adapter/z-stack/constants/common.ts
r0b1n: Good point, indeed. I copy-pasted the name from `SWRA442` doc. Apparently it is a typo in there. Renamed.

src/adapter/z-stack/constants/common.ts
r0b1n: Agree. Using some regexp magic I made them enums and moved everything under the COMMON namespace in `index.ts`. Also did a bit of cleanup in files which use those constants to make code a bit less verbose.

r0b1n:
@Koenkk I did one more iteration, should be better now.

Koenkk:
Looks great! Big thanks 🎖️

PR #182, opened 2020-05-27
Add more configurable options for sending message

AlexLucky1983:
Add ability to specify direction and reserved bits for zcl frame in endpoint.write method
It will help to create converter for new livolo devices: https://github.com/Koenkk/zigbee2mqtt/issues/3560

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -164,9 +164,7 @@ abstract class Adapter extends events.EventEmitter {
 
     public abstract setChannelInterPAN(channel: number): Promise<void>;
 
-    public abstract sendZclFrameInterPANToIeeeAddr(
-        zclFrame: ZclFrame, ieeeAddress: string
-    ): Promise<void>;
+    public abstract sendZclFrameInterPANToIeeeAddr(zclFrame: ZclFrame, ieeeAddress: string): Promise<void>;
 
     public abstract sendZclFrameInterPANBroadcast(
         zclFrame: ZclFrame, timeout: number
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -151,10 +151,10 @@ abstract class Adapter extends events.EventEmitter {
      */
 
     public abstract sendZclFrameToEndpoint(
-        networkAddress: number, endpoint: number, zclFrame: ZclFrame, timeout: number,
+        networkAddress: number, endpoint: number, zclFrame: ZclFrame, timeout: number, sourceEndpoint?: number
     ): Promise<ZclDataPayload>;
 
-    public abstract sendZclFrameToGroup(groupID: number, zclFrame: ZclFrame): Promise<void>;
+    public abstract sendZclFrameToGroup(groupID: number, zclFrame: ZclFrame, sourceEndpoint?: number): Promise<void>;
 
     public abstract sendZclFrameToAll(endpoint: number, zclFrame: ZclFrame, sourceEndpoint: number): Promise<void>;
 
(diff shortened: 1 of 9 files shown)


src/adapter/adapter.ts
```diff
@@ -151,7 +151,7 @@ abstract class Adapter extends events.EventEmitter {
      */
 
     public abstract sendZclFrameToEndpoint(
-        networkAddress: number, endpoint: number, zclFrame: ZclFrame, timeout: number,
+        networkAddress: number, endpoint: number, zclFrame: ZclFrame, timeout: number, srcEndpoint?: number
```
Koenkk, line 154: For consistency purposes, please add the `srcEndpoint` to every `sendXXX` method.

src/controller/model/endpoint.ts
```diff
@@ -181,7 +184,7 @@ class Endpoint extends Entity {
     }
 
     public async write(
-        clusterKey: number | string, attributes: KeyValue, options?: Options
+        clusterKey: number | string, attributes: KeyValue, options?: Options, transactionSequenceNumber?: number
```
Koenkk, line 187: Please move `transactionSequenceNumber` into the `options`.

src/controller/model/endpoint.ts
```diff
@@ -202,12 +205,13 @@ class Endpoint extends Entity {
         debug.info(log);
 
         try {
+            transactionSequenceNumber = transactionSequenceNumber || ZclTransactionSequenceNumber.next();
             const frame = Zcl.ZclFrame.create(
```
Koenkk, line 209: The options should also be honoured in all other method, like `read`, `command`, etc.

src/zcl/zclFrame.ts
```diff
@@ -82,6 +82,7 @@ class ZclFrame {
 
     private writeHeader(buffalo: BuffaloZcl): void {
         const frameControl = (
+            this.Header.frameControl.reservedBits |
```
Koenkk, line 85: What is the purpose of reserved bits.

Koenkk:
Thanks for your pr, I've added some review comments/questions. 

It seems that after this change some tests fail, so please investigate this (https://dev.azure.com/koenkk/zigbee/_build/results?buildId=4655&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=dee4e6ec-be1b-5671-4678-ab293b4004c5&l=1926)

src/adapter/adapter.ts
AlexLucky1983: Added

src/controller/model/endpoint.ts
AlexLucky1983: Moved. But some methods already has it as parameter. I didn't change their signature.

src/controller/model/endpoint.ts
AlexLucky1983: Updated

src/zcl/zclFrame.ts
AlexLucky1983: It's part of the control frame according to the ZigBee standart: 5-7 bits. Masked them by 0xE0. I know that it's suggested to be 0, but Livolo uses them. 
After these changes I'm going to update converters for their new sockets and switches. 
https://github.com/Koenkk/zigbee2mqtt/issues/3560

Need only understand how to setup modelID manually in configuration file, like we do for 'friendly_name', so if you can help here it would be great.

Koenkk:
Thanks, updated the code a bit.

Breaking change for you is that the reservedBits are now shifted, so in case were you now use 224 to set all 3 reserved bits to 1 you now have to use 7.

Regarding the modelID, I assume the device does not provide a modelID at all? If this is the case you can create a fingerprint to identify the device, e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L3529

AlexLucky1983:
> Regarding the modelID, I assume the device does not provide a modelID at all? If this is the case you can create a fingerprint to identify the device, e.g. https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/devices.js#L3529

Both devices has the same returned info. modelID and any other stuff are fully identicall. I really don't know how to separate them except by manually intervention in database.db. 
Maybe Livolo use some their own databse with network address to match device and its type...

Koenkk:
@AlexLucky1983 can you post the full database.db entry?

kirovilya:
@Koenkk I can confirm - there are no differences in the structure of endpoints and clusters.
we (with @AlexLucky1983 ) have found so far the only way to distinguish devices - they send different data by an unknown command.
socket:
image
new single switch:
image
old single switch:
image

kirovilya:
and another one: transactionSequenceNumber can bee 0 for toggle command.
now if 0, then a new one is generated

PR #191, opened 2020-06-16
Zigbee Environment Sensor

iharyadi:
Hello Koen,

This is another change that I need to make for my sensor to work with ZIgbee2mqtt.  My sensor has one unintended type for Pressure Measurement.  I need this change to make the pressure measurement work.   

I have sent you separate pull request for the conventer.

Thanks
Iman

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -2268,6 +2268,18 @@ const Cluster: {
         commandsResponse: {
         },
     },
+
+    kmpcilPressureMeasurement: {
+        ID: 1027,
+        attributes: {
+            measuredValue: {ID: 0, type: DataType.uint16},
+        },
+        commands: {
+        },
+        commandsResponse: {
+        },
+    },
+
     msFlowMeasurement: {
         ID: 1028,
         attributes: {
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -2311,6 +2311,17 @@ const Cluster: {
         commandsResponse: {
         },
     },
+    pHMeasurement: {
+        ID: 1033,
+        attributes: {
+            measuredValue: {ID: 0, type: DataType.uint16},
+            minMeasuredValue: {ID: 1, type: DataType.uint16},
+            maxMeasuredValue: {ID: 2, type: DataType.uint16},
+            tolerance: {ID: 3, type: DataType.uint16},
+        },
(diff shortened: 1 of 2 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -2268,6 +2268,18 @@ const Cluster: {
         commandsResponse: {
         },
     },
+
+    kmpcilPressureMeasurement: {
+        ID: 1027,
```
Koenkk, line 2273: I've received your sensors, thanks!

I guess this is still needed, can you add your manufacturerCode here? (e.g. https://github.com/Koenkk/zigbee-herdsman/blob/d564422bab9ceffab1f019f8cd8dbbdc7063710b/src/zcl/definition/cluster.ts#L3694)

iharyadi: Hi Koen ,  

Thank you for letting me know that you have received the sample.

Yes,  this change is needed to make the opration clean.  Without this change,  the pressure parameter still get reported.  However,  the reporting configuration cannot be controled.

This change is needed because I made a mistake on the firmware.  The pressure measurement should be int16.  I declare it as uint16.  I have shipped the sensor with the mistake and used in other hubs.  I am concern if I make correction to int16 it will cause issue.  I will keep consider making corrrection in the firmware in the future.  For now,  I would like to keep it with uint16.   As a context,  the pressure data from Bosch 280 is unit32.   I may have considered this to cover more range when I did choose uint16 instead of the ZCL specification.

I do not have a mfg code.  The volume that I make is so small to get mfg code.  The design that you made is very good.  I can isolate this exception using specific attribute to work around the issue.   This attribute will probably never be used by other devices.

BTW,  I did not request a push for the code to convert to Homeassitant topic.   I am assuming that a lot of people may want to use it in HA.  I will request a pull for the code.  There is nothing controversial on that part of the code change.  It is just adding my sensor and the list of attribute descriptions for HA.

Thanks
Iman

Koenkk: I figured out it might be much easier as we can easily override the `dataType`. I've created a PR if you can test that we can merge that one and close this. (https://github.com/Koenkk/zigbee-herdsman-converters/pull/1413)

Koenkk: Closing this because it's obsolete due https://github.com/Koenkk/zigbee-herdsman-converters/pull/1413

PR #193, opened 2020-06-26
add feature to do not invoke devices bootloader.

slaesh:
Option for devices with automatic "reset to bootloader".
Otherwise it will be held in RESET mode if the serial port is opened.

The "skipBootloader" is not working in this case.
during writing this comment, i guess it would be better and more explicit to move my code into this function ;)

TODO: 

- [ ] - clarify naming convention for settings
- [ ] - clarify if it's possible to use serialport-object instead of seperate arguments
- [ ] - implement settings pass through-part in zigbee2mqtt
- [ ] - move into "skipBootloader" method if it will be approved!
- [ ] - add tests

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -11,6 +11,12 @@ interface SerialPortOptions {
     rtscts: boolean;
     path: string;
     adapter: 'zstack' | 'deconz';
+
+    /**
+     * Set to true will fore DTR and RTS to FALSE.
+     * This will ensure that devices Bootloader is NOT invoked.
+     */
+    do_not_invoke_bootloader?: boolean;
 };
 
 interface AdapterOptions {
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -61,7 +61,7 @@ class ZStackAdapter extends Adapter {
         serialPortOptions: SerialPortOptions, backupPath: string, adapterOptions: AdapterOptions) {
 
         super(networkOptions, serialPortOptions, backupPath, adapterOptions);
-        this.znp = new Znp(this.serialPortOptions.path, this.serialPortOptions.baudRate, this.serialPortOptions.rtscts);
+        this.znp = new Znp(this.serialPortOptions.path, this.serialPortOptions.baudRate, this.serialPortOptions.rtscts, this.serialPortOptions.do_not_invoke_bootloader);
 
         this.transactionID = 0;
         this.closing = false;
--- a/src/adapter/z-stack/znp/znp.ts
+++ b/src/adapter/z-stack/znp/znp.ts
@@ -55,6 +55,11 @@ class Znp extends events.EventEmitter {
     private path: string;
     private baudRate: number;
     private rtscts: boolean;
+    /**
+     * local copy of incoming serialPortOptions
+     * -> maybe use this object instead of single properties?
+     */
+    private do_not_invoke_bootloader: boolean;
 
     private portType: 'serial' | 'socket';
     private serialPort: SerialPort;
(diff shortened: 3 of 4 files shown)


Koenkk:
Would it be possible to automatically detect this? Which adapters require this?

slaesh:
Would be possible, could try to check the USB devices name..
My own CC2652 stick. Used for dev stuff, zigbee coordinator or thread
border router..

Problem is that the CC2652's bootloader seems to don't have this magic
start byte..

Will try to programm some custom name or something like this into the USB
chip.. and trying to detect it then.

Thanks so far :)

Will come back to u ✌🏻

Koen Kanters <notifications@github.com> schrieb am Sa., 27. Juni 2020,
13:21:

slaesh:
hey @Koenkk ,

detection will work!

my idea is now, to build generic "device-callbacks" .. so if a given "filter" matches (just a function), a set of given callbacks (maybe before-open, after-open, before-close, after-close, ...) will be called.

what do you think? :)

Koenkk:
@slaesh what I'm also wondering, could there be an alternative for the magic CC2531 byte for the CC2652? Could you perhaps ask this to TI? (https://e2e.ti.com/)

slaesh:
asked! in the official docs is nothing.. 

RUN is not available:
image

maybe RESET will work, but if its a hard-reset we will go again into the BL ;D

slaesh:
that "magic byte" seems to be the EOF cmd.. so BL thinks that the upload process is done and starts the "newly" flashed APP.. could work on the CC2652 too, will test that..

anyway! i need another workaround.. cause the "normal" state of the RTS causes a permanent RESET on my board.. 

any idea how i could do that? :)
first way using the config.. the second using a generic device-callback-thing.. or dont u want to support any of those?

i guess i would keep my branch up to date... =/

Koenkk:
> that "magic byte" seems to be the EOF cmd.. so BL thinks that the upload process is done and starts the "newly" flashed APP.. could work on the CC2652 too, will test that..

Please let me know.

Regarding setting DTR to false, involving @omerk here as I think this is also needed for the zzh-p (as I needed to unplug the DTR to make it survive the reset). Is that right?

omerk:
So I feel there is a bit of a confusion here (or I might well be missing something): There is no "magic byte" to enter the BSL on CC1352/CC2652, you trigger it by pulling the BSL trigger pin (default: `DIO_13`) low before resetting the device. @Koenkk on the "zzh-p" prototype you have DTR is connected to DIO_13 and RST is connected to RESET directly, DTR is probably set low as the device is powered up hence BSL being triggered. I would suggest adding a "set DTR high and toggle RTS (with final state high)" as a precautionary step as part of comms init to make sure device doesn't stay in BSL mode even if it was accidentally triggered.

This requires a bit more thought and experimentation before I lock down the zzh-p design...

Sidenote: Why DTR/RTS and not any of the other RS232 control signals? Well, `cc2538-bsl.py` already uses these for autoreset so just sticking to that.

slaesh:
thats the point @omerk ! thanks for your resume! =)
and exactly that is what my PR is actually is adressing ;)

putting DTR to false and RTS to true, and then putting them both to false..
my board uses the same schematic: RTS --> RESET and DTR --> BL_EN

try it out, should do the trick ;)

this will reset the device on startup, but wont hold it in bootloader-mode.

omerk:
My apologies, should have read the actual PR and not just the last few comments. At least we are all talking about the same thing though :-)

Koenkk:
Would it be problematic if `dtr` and `rst` is always set to low on start? In that case we can get rid of the option.

slaesh:
depends on the timings i guess.. could lead to problems.
will make some tests without doing it in two steps..

slaesh:
seems like it's working.. got a problem like once every 50 starts..
but do make it bullet-proof i would do it in two steps, just to be sure!

the 100% solution would be to make it configurable (BL_EN: high or low, RTS: reset or BL_EN, ...)

what do you think @Koenkk @omerk ?
how will we proceed and who wants to do it? =)

slaesh:
i see those options:

1. just put them to false (two-stepped) in case that "rtscts" is false
2. add a new option in our configuration for this behavior
3. build a generic device-callback thing based on the serial-device-infos

option 1 will be done in several hours including tests and so on.. its easy!
option 2 will take some more time, but not that much.. i guess option 1 would be better..
option 3 sound cool, would like to implement it.. but maybe there are some devices that cannot or wont change the serial-device's info.. could be an overkill for too less devices may support it..

Koenkk:
I will test if it's possible to set `dtr`, `rst` and `rtscts` to false by default and see if this works for various adapter (zzh, zzh-p, cc2531, cc2652r launchpad, conbee II)

slaesh:
I guess the launchpad isn't wired like this. but sure, test it! :)
any time plans? can I support you?

Koenkk:
@slaesh I've did some checks and the following works well for the zzh-p and doesn't break anything for existing supported adapters (launchpads, zzh, cc2531 1.2/3.0): https://github.com/Koenkk/zigbee-herdsman/pull/196

slaesh:
#196 works just fine!

closing this.. thanks!

PR #194, opened 2020-06-27
Allow programming locks

jhenkens:
This adds the command definitions needed to support getting and programming pin codes in zigbee keypad locks.


Koenkk:
Thanks!

PR #195, opened 2020-07-01
Fix for unhandled data indications and confirms

ChrisHae:
Evaluate the device state every time when the device state is polled.


Koenkk:
Thanks!

PR #197, opened 2020-07-05
Let adapter decide defaults options and make rtscts false by default

Koenkk:
Setting `rtscts` to `false` by default doesn't seem to have any regression and will make the use of e.g. the CC2530 and zzh easier (no need to set this manually anymore). Tested with a CC2531 on macOS and Linux.

@sjorge if I'm correct, you are using a CC2531 on FreeBSD, right? To cross-check, can you check if this doesn't break (if Zigbee2mqtt start OK this works)? No need to pull anything, just set the following in `configuration.yaml`:

```yaml
advanced:
  rtscts: false
```

FYI: @omerk @slaesh

Zigbee2mqtt changes: https://github.com/Koenkk/zigbee2mqtt/pull/3854

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -7,10 +7,10 @@ interface NetworkOptions {
 }
 
 interface SerialPortOptions {
-    baudRate: number;
-    rtscts: boolean;
-    path: string;
-    adapter: 'zstack' | 'deconz';
+    baudRate?: number;
+    rtscts?: boolean;
+    path?: string;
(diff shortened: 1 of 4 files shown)


sjorge:
I'm using a CC2531 on illumos, let me try and pair a bulb again and after let me apply this patch to see if it works.

sjorge:
I added  ```rtscts: false``` to my advanced section and everything still works after restarting z2m.

Koenkk:
Thanks for confirming @sjorge !

PR #199, opened 2020-07-16
specific modification for the Zemismart component (ZM-CSW032-D)

ullrik:
This update is intended for an interactive movement by definition of the position. Zemismart use the commands "open, close" and not "upOpen, downClose" (define in closuresWindowCovering).
This will then be used by the "zigbee-herdsman-converters" component (toZigbee.js)


ullrik:
Need this update in "zigbee-herdsman-converters" : https://github.com/Koenkk/zigbee-herdsman-converters/pull/1404

PR #200, opened 2020-07-16
Fix OTA Update with deCONZ adapter

ChrisHae:
Added slip encode to send message


Koenkk:
Thanks!

src/adapter/deconz/driver/driver.ts
```diff
@@ -396,6 +404,12 @@ class Driver extends events.EventEmitter {
             return;
         }
 
+        if (this.apsRequestFreeSlots !== 1) {
+            debug("no free slots. Delay sending of APS Request");
+            this.sleep(1000);
```
ammmze, line 409: I know this was already merged ... i've just been following #72 and saw mention of the OTA fix, so came to check out what changed and noticed this sleep here that I don't think is doing what was intended here, so I thought I'd bring it up. The sleep is asynchronous, so the method is returning immediately rather than waiting 1 second.

PR #217, opened 2020-09-05
Scene support

K-r-i-s-t-i-a-n:
Bare support for scenes: added responses to event.js

Corresponding pull request for zigbee-herdsman-converters: [#1531](https://github.com/Koenkk/zigbee-herdsman-converters/pull/1531)

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -2,6 +2,7 @@ name: "Close stale issues/pull requests"
 on:
   schedule:
   - cron: "0 0 * * *"
+  workflow_dispatch:
 
 jobs:
   stale:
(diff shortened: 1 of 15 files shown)


Koenkk:
Hello, thanks for the pr.

I've updated `clusters.ts` in your PR, this should fix the timeout issues you were having. This should also eliminate the need for adding any `fromZigbee.js` converters  and changing `events.ts` since you will be able to directly access the response with: `const response = await entity.command('genScenes', 'view', {'groupid': value.groupid, 'sceneid': value.sceneid}, getOptions(meta.mapped));`

K-r-i-s-t-i-a-n:
How should we add those toZigbee converters to the devices in devices.js? I think it would be pretty inefficient to add them to each device separately, they should be according to the specification and thus be supported by every device. 

My thought: add as default to every device, exclude if device is not compatible.

Koenkk:
Thanks!

PR #219, opened 2020-09-10
Add TH1300ZB specific attributes to Sinope manufacturer cluster

embak:
Prerequisite to the addition of converters to Sinope TH1300ZB floor thermostat.


Koenkk:
Thanks!

PR #221, opened 2020-09-16
Added ssIasAce cluster to 11 endpoint

Hacker-CB:
Coordinator should have ssIasAce cluster to receive button event from devices like HEIMAN HS1RC-N (remote control).

**NOTE:**
Endpoint with new cluster will be not registered inside `registerEndpoints()` untill full CC2538 reset. (I made such reset by reflashing firmware).
@Koenkk - I don't know how to re-register endpoint. I hope you can do that fix. May be you need to implement something like `meta.configureKey` in `devices.js`

Thank you.

P.S.

This can be possible fix of the next issues:
https://github.com/Koenkk/zigbee2mqtt/issues/2750
https://github.com/Koenkk/zigbee2mqtt/issues/1957

### DIFF
--- a/src/adapter/z-stack/adapter/startZnp.ts
+++ b/src/adapter/z-stack/adapter/startZnp.ts
@@ -46,7 +46,10 @@ const Endpoints = [
         appprofid: 0x0104,
         appdeviceid: 0x0400,
         appnumoutclusters: 2,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks, I will check if we can re-register, currently we did not do that and just require reflashing of the adapter.

Issue #233, opened 2020-10-08
NV items naming error.


Hacker-CB:
Currently we are working for backup migration 2538 <-> 2652 and seems that we found some naming error.

https://github.com/Koenkk/zigbee-herdsman/blob/7e4b3f052842f2c9e4a21e0005dba260f98ee92e/src/adapter/z-stack/constants/common.ts#L143

`257 = 0x101`

`ZComDef.h` **from the TI Z-Stack 1.2**
```
// NV Items Reserved for Trust Center Link Key Table entries
// 0x0101 - 0x01FF
#define ZCD_NV_TCLK_TABLE_START           0x0101
#define ZCD_NV_TCLK_TABLE_END             0x01FF
```

 **TI Z-Stack 3.0:**

```
// NV Items Reserved for Trust Center Link Key Table entries
// 0x0101 - 0x01FF
#define ZCD_NV_TCLK_SEED                  0x0101  //Seed
#define ZCD_NV_TCLK_JOIN_DEV              0x0102  //Nv Id where Joining device store their APS key. Key is in plain text.
#define ZCD_NV_TCLK_DEFAULT               0x0103  //Not accually a Nv Item but Id used by SecMgr

#define ZCD_NV_TCLK_IC_TABLE_START        0x0104  //IC keys, refered with shift byte
#define ZCD_NV_TCLK_IC_TABLE_END          0x0110

#define ZCD_NV_TCLK_TABLE_START           0x0111  //Entries to store users of the keys
#define ZCD_NV_TCLK_TABLE_END             0x01FF
```

**I Z-Stack 3.x**

```
// NV Items Reserved for Trust Center Link Key Table entries
// 0x0101 - 0x01FF
#define ZCD_NV_TCLK_SEED                  0x0101  //Seed
#define ZCD_NV_TCLK_JOIN_DEV              0x0102  //Nv Id where Joining device store their APS key. Key is in plain text.
#define ZCD_NV_TCLK_DEFAULT               0x0103  //Not accually a Nv Item but Id used by SecMgr

#define ZCD_NV_LEGACY_TCLK_IC_TABLE_START 0x0104  // Deprecated. Refer to ZCD_NV_EX_TCLK_IC_TABLE
#define ZCD_NV_LEGACY_TCLK_IC_TABLE_END   0x0110  // IC keys, refered with shift byte

[...]

PR #234, opened 2020-10-08
Fix invalid ID for LEGACY_TCLK_TABLE_START

Hacker-CB:
Fix for https://github.com/Koenkk/zigbee-herdsman/issues/233


Hacker-CB:
Currently backup fails with
```
Error: SREQ '--> SYS - osalNvRead - {"id":511,"product":2,"offset":0,"osal":true}' failed with status '(0x02: INVALID_PARAM)' (expected '(0x00: SUCCESS)')
```

So, I continue to work

Hacker-CB:
Now all done.
Was tested on 2538 and 2652.
@Koenkk You can merge this.

P.S. Bi-directional backup migration beween 2538 and 2562 can work with this backup. I will make another PR for such migration tomorrow.

Hacker-CB:
Please keep in mind, that it is breaking changes in `coordinator_backup.json`:

`ZCD_NV_TCLK_SEED` was added with ID 257
`LEGACY_TCLK_TABLE_START` now has id 273 and another length.

Hacker-CB:
Please wait before merge. I will submit one small patch to tests

Hacker-CB:
Now it's ok :)

Koenkk:
This will break https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/z-stack/adapter/startZnp.ts#L191 (zstack 1.2), it will write to 273 instead of 257 after this. We should add:

```
    LEGACY_TCLK_TABLE_START_12 = 257, // Valid for <= Z-Stack 1.2
    LEGACY_TCLK_TABLE_START_30 = 273, // Valid for <= Z-Stack 3.0.x
```

Hacker-CB:
I added LEGACY_TCLK_TABLE_START_12 but instead of using LEGACY_TCLK_TABLE_START_30 I suggest keep LEGACY_TCLK_TABLE_START

src/adapter/z-stack/adapter/nvItems.ts
```diff
@@ -91,7 +91,7 @@ export default {
     },
     tcLinkKey: (): NvItem => {
```
Koenkk, line 92: This should be named `tcLinkKey12`, to make clear it's for Zstack 12 only.

src/adapter/z-stack/adapter/backup.ts
```diff
@@ -72,6 +72,12 @@ const items = {
         osal: true,
         product: PRODUCT_ALL,
     },
+    ZCD_NV_TCLK_SEED: {
```
Koenkk, line 75: Last question; is it actually needed to backup this? For Zstack 3.x.0 we never made a backup of `0x0101` and we did not encounter any issues because of this.

Hacker-CB: I'm not sure, but in all previous releases you backed that up because of the wrong `LEGACY_TCLK_TABLE_START` ID.

@0x3ec, how do you think?

Koenkk: We never backed it up for ZStack 3.x.0, I expect the same for ZStack 3.0.x.

Hacker-CB: 2592 <-> 2538 migration works fine together with this PR. My collegue @0x3EC helped me with this commit. He is more familar in Z-Stack. So let's wait his answer.

In any case, I can try to migrate 2538 <-> 2562 without backing up `LEGACY_TCLK_TABLE_START`, but I'm not sure that it will be stable.

Koenkk: `LEGACY_TCLK_TABLE_START` should be backed up and restored for sure (I'm not sure how this would work out yet because Zstack 3.x.0 doesnt have this). But I think `ZCD_NV_TCLK_SEED` is not needed.

Hacker-CB: @Koenkk , actually you never made backup of the `LEGACY_TCLK_TABLE_START=273` and it was also worked. But I don'n know how :)

Hacker-CB: `ZCD_NV_TCLK_SEED` was generated by the random. It used during genrating link key.
There is no official inforamtion from the TI about migration between stack revisions and this parameter.
We found some info in the TI forums about that, but can't confirm: https://e2e.ti.com/support/wireless-connectivity/zigbee-and-thread/f/158/t/814208

Koenkk: The TCLK table contains the Zigbee 3.0 device security key of each device. I however don't know what happens when this is not restored :smile: (but doesn't look like a big deal?).

Hacker-CB: I tried to migrate 2538 -> 2562 without TCLK table and failed.
May be I had not enough patience to wait that ZigBee networked will be restored, but after 5-10mins that did not happned.
So,I think we have enough disk space on the user devices to backup extra 100-300 bytes :)
If you prefer to delete this table from the backup - I can do that. Let me know your desicion.

Hacker-CB: About TCLK table: https://www.ti.com/lit/an/swra671/swra671.pdf?ts=1602185076595&ref_url=https%253A%252F%252Fwww.google.com%252F page 4.

0x3EC: The **ZCD_NV_TCLK_SEED** value is used in the **_ZDSecMgrGenerateKeyFromSeed_** function. The function description is as follows: _"Generate the TC link key for an specific device usign seed and ExtAddr"_.

The **_ZDSecMgrGenerateKeyFromSeed_** function is called when a **ZDO_REQUEST_KEY_IND** request is received. 
I have not encountered such requests (RequestKey). 
Maybe a device manufactured in accordance with all the zigbee 3 safety rules can send this request.

Koenkk: @Hacker-CB backing up TLCK_SEED can be removed. My point is that NV item `257` was never backed up for ZStack 3.x.0, so I don't see a reason to start backing it up now. This would also prevent a breaking change on the ZStack 3.x.0 backups.

Hacker-CB: I can remove, but in any case it will be breaking change in the backup.
`LEGACY_TCLK_TABLE_START` now has id 273 and before it had 257

Koenkk: I've checked, but now I remember that the IDs of the NV items are stored in backup file. So for an old backup LEGACY_TCLK_TABLE_START will have an ID of 257, which is good. So I think there is no breaking change.

Still I think we should remove this, as we shouldn't backup/restore items which we are not sure we should back up at all.

Hacker-CB: I removed TCLK_SEED from the backup.

Hacker-CB:
@Koenkk what about to merge this?

Koenkk:
@Hacker-CB yes, thanks!

PR #235, opened 2020-10-09
zStack 3.x.0 <-> zStack 3.0.x bidirectional migration

Hacker-CB:
Was tested with 2538 and 2652 adapters on Zigbee network with 45 devices.
Should be meged after https://github.com/Koenkk/zigbee-herdsman/pull/234


Hacker-CB:
I don't know why tests was failed on ci. My local tests are fine.

Hacker-CB:
In any case we need to rebase it after merging https://github.com/Koenkk/zigbee-herdsman/pull/234

Hacker-CB:
@Koenkk I rebased this PR to master, but tests failed againg, but there is no problem when I run `npm run test` in my PC.
Could you point me, which is wrong?

Koenkk:
It fails because tests are missing for the new functionality. You can reproduce the result by running `npm run test-with-coverage` and view the coverage report in the created `coverage` directory (search for `index.html`), everything should be 100%.

src/adapter/z-stack/adapter/backup.ts
```diff
@@ -181,7 +214,8 @@ async function Restore(znp: Znp, backupPath: string, options: NetworkOptions): P
 
     await znp.request(Subsystem.SYS, 'osalNvWrite', backup.data.ZCD_NV_EXTADDR);
     await znp.request(Subsystem.SYS, 'osalNvItemInit', ZCD_NV_NIB, null,
-        [ZnpCommandStatus.NV_ITEM_INITIALIZED]);
+        [ZnpCommandStatus.SUCCESS, ZnpCommandStatus.NV_ITEM_INITIALIZED]);
+    await znp.request(Subsystem.SYS, 'osalNvWrite', backup.data.ZCD_NV_NIB);
```
Koenkk, line 218: Could you explain why this was added, besides that I have no comments. looks good :smile:

Hacker-CB: This handle case, if NV item was already initialized, but we need to ovreride it.
zStack return `ZnpCommandStatus.SUCCESS` in this case.
We are overriding it with new value in line 218.

Koenkk: Thanks for claryfying.

Hacker-CB:
@Koenkk now all tests are ok

Koenkk:
Thanks!

PR #241, opened 2020-10-21
WIP: Adapter ZiGate

G1K:
My comrades (@devbis, @kirovilya) and I prepared an initial implementation of an adapter for interacting with NXP JN5168 / JN5169 chips
We are using zigbee2mqtt on a xiaomi gateway with openwrt operating system.
https://github.com/openlumi

Some places are not yet sufficiently perfected and I have not managed to implement all of them, but I want to share with the community so that my efforts are not wasted.

For successful operation, you need to use the firmware for the chip with a memory leak fix, it will be released in the ZiGate 3.1d firmware, but it is not stable yet, I added this fix to 3.1c
You can download the firmware or source codes here https://github.com/openlumi/ZiGate/releases/tag/snapshot-20201008

After flashing, you need to do a hard reset or `./zigate.sh erase_pdm` (script next to the firmware)

You can flash using official utilities from NXP for windows or
https://github.com/Jerome-PS/JennicModuleProgrammer

What you can try as a coordinator

- Xiaomi gateway DGNWG05LM, ZHWG11LM
- End devices from xiaomi https://github.com/Jerome-PS/JN516xSniffer

   - Xiaomi Door sensor
   - Xiaomi smart button
   - Xiaomi smart button UART1 ZiGate coordinators

- Ready modules https://www.seeedstudio.com/Mesh-Bee-Open-Source-Zigbee-Pro-Module-with-MCU-JN516-p-1751.html
- DIY module + TTL adapter
E75-2G4M10S Zigbee JN5169 aliexpress.com/item/32973411720.html

Thanks to the creators of these libraries (@Neonox31 , @nouknouk), they helped in understanding the protocol and implementation of this adapter.
https://github.com/Neonox31/zigate
https://github.com/nouknouk/node-zigate


src/controller/model/device.ts
```diff
@@ -447,6 +447,7 @@ class Device extends Entity {
         this.save();
         debug.log(`Interview - got active endpoints for device '${this.ieeeAddr}'`);
 
+        debug.log(this.endpoints);
```
Koenkk, line 450: Please remove this.

Koenkk:
Cool!

Your pr looks ok (added one small comment).

You marked your pr as wip, but can this already be integrated?

Note to self: after this merge: https://github.com/Koenkk/zigbee2mqtt/pull/4737

G1K:
Thank you, I accidentally committed that line.
 
I think for a general test it is better to merge this and give dev a chance to test a wide range of people. Functionally, this adapter does not overlap with the others, so it should not cause problems for the majority of users.

WIP: I marked it because the functionality does not fully cover the capabilities and not everything is implemented, but it should be enough for basic use.
I have very few devices to perform tests, almost all of these are end devices without power.

But I will definitely not have the opportunity to implement it in the near future.

An approximate list of problems:
- Groups were not implemented
- lqi for the map is not fully tested and finalized (perhaps you need to make a routingTable)
- Broadcast messages have not been tested either
- Green Power does not work - there is no default in the firmware, but even when I put it together it was not possible to send,  perhaps I did not fully understand the parameters.
- InterPAN - I don't know what it is and I don't even want to guess)

and others I don't know ... I'm not an expert in ZigBee)

.npmignore
```diff
@@ -14,7 +14,7 @@ coverage
 
 #editor settings
 .vscode
+.idea
 
 #src
-src
```
Koenkk, line 19: Please readd src here, (somehow I cannot change this myself in this PR)

G1K: Please readd src here, (somehow I cannot change this myself in this PR)

Koenkk:
Thanks!

Koenkk:
yes please open a new pr

PR #253, opened 2020-11-09
Fixed definition of manuSpecificTuya cluster

### LINKED ISSUE #249: Incorrect Tuya protocol fields
Hi,

After getting some specs from manufacturer and looking at Tuya pages I've found out that current Tuya protocol fields are incorrect, the `dp` field should be `uint8` not `uint16`, the first byte is type of data as specified [here](https://developer.tuya.com/en/docs/iot/device-development/access-mode-mcu/zigbee-general-solution/tuya-zigbee-module-uart-communication-protocol/tuya-zigbee-module-uart-communication-protocol?id=K9ear5khsqoty#title-8-Command%20issued)

Over the weekend (and maybe some of the next week) I'm willing to do a sizeable Tuya rework:
1. Rename the endpoint from `manuSpecificTuyaDimmer` to just `manuSpecificTuya`
2. Fix the endpoint fields
3. Create a list of data points in converters common.ts from all Tuya devices
4. Modify Tuya commands to use proper `data type` and `data point` combo
5. Add full support for Saswell TRVs

Are you willing to accept such a big change? I'm willing to provide additional support in case of problems with my change

drzony:
Fixes #249

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -3796,7 +3796,7 @@ const Cluster: {
         commands: {},
         commandsResponse: {}
     },
-    manuSpecificTuyaDimmer: {
+    manuSpecificTuya: {
         ID: 0xEF00,  // 61184
         attributes: {
         },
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #257, opened 2020-11-21
zigate: return firmware version in unified format

devbis:
I modified the fetching of the firmware version to match the ones from other adapters.
It removes the error while accessing the revision property in bridgeLegacy.js


Koenkk:
I think this issue should be fixed in bridgeLegacy.js instead of here.

devbis:
Anyway, it makes the CoordinatorVersion content consistent along all adapters.

Koenkk:
Agree, thanks

PR #262, opened 2020-11-23
WORK IN PROGRESS Cleanup

nurikk:
Use ts-test instead of babel
before
```
added 999 packages, and audited 999 packages in 18s
```
after
```
added 707 packages, and audited 707 packages in 23s
```


package.json
```diff
@@ -76,20 +70,10 @@
...
     "test": "jest test",
-    "test-with-coverage": "jest test --coverage --collectCoverageFrom 'src/**/*.ts'",
+    "test-with-coverage": "jest test --coverage",
     "test-watch": "jest test --watch",
     "eslint": "node_modules/.bin/eslint . --ext .ts",
     "docs": "typedoc --tsconfig typedoc-tsconfig.json"
   },
-  "version": "0.13.33",
-  "jest": {
-    "coverageThreshold": {
```
Koenkk, line 86: Why was this removed?

nurikk: It's moved to corresponding.json file

nurikk: jest.config.js

src/controller/controller.ts
```diff
@@ -73,7 +73,7 @@ class Controller extends events.EventEmitter {
      *
      * To auto detect the port provide `null` for `options.serialPort.path`
      */
-    public constructor(options: Options) {
+    public constructor(options: Partial<Options>) {
```
Koenkk, line 76: Why put partial here?

nurikk: My bad, fixed

PR #264, opened 2020-11-28
Update events.ts

dettrick:
Updating to allow zigbee alarm panels to request for panel status from the server. at the moment when a issues the command the herdsman converter skips it because it is not contained in the lookup (see below image). I will also update the fromzigbee.js in that repo to allow it decode the message

image


Koenkk:
Thanks!

PR #266, opened 2020-11-29
Adds all known manufacturer codes

pedrolamas:
This adds the Atmel manufacturer code to the list of known manufacturer codes.

Doing this as an initial step to [fix this issue](https://github.com/Koenkk/zigbee2mqtt/issues/4969), as I came to realize that my Philips Hue lamps are built by Atmel (or at least have their components in them) so we need to check for both Philips and Atmel manufacturer codes when deciding to poll these lamps for their state.

I'm adding this code specifically, but I wonder if it would be worth adding all the ones in [this list](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/de_web_plugin_private.h)?

### DIFF
--- a/src/zcl/definition/manufacturerCode.ts
+++ b/src/zcl/definition/manufacturerCode.ts
@@ -1,11 +1,1072 @@
+// https://github.com/wireshark/wireshark/blob/master/epan/dissectors/packet-zbee.h
+const knownManufacturerCodes = {
+    PANASONIC_RF4CE: 0x0001,
+    SONY_RF4CE: 0x0002,
+    SAMSUNG_RF4CE: 0x0003,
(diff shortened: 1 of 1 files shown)


Koenkk:
I'm fine adding the others, can you update this PR?

pedrolamas:
Doing it now then, will be ready in 5 mins! 🙂

pedrolamas:
@Koenkk updated to add all knows codes from that list, but tried to ensure compatibility by maintaining the existing names!

Koenkk:
Thanks!

pedrolamas:
Thanks, I will wait for the next herdsman version so I can then fix the main issue!

Issue #273, opened 2020-12-16
deCONZ driver has a 60ms delay when reading a button press


ndfred:
I am trying to make my lights as responsive as possible. However, when I press the button on any of my remotes there is a noticeable delay for the press to appear in the logs. The button I am using here is bound to a light, and the light turns on noticeably quicker than the Z2M logs show.

I pulled some debug logs with `DEBUG=zigbee-herdsman* /usr/bin/npm start` and you can clearly see the sequence here:

* we get a `DEVICE_STATE` frame from the button indicating something happened, that gets logged exactly when I press the button
* we send the read data request to the button to (I guess) know exactly what happened 79ms after that
* 6ms later the button answers with what changed, and Z2M logs that as a button press

Basically we introduce an artificial 80ms delay when reading button clicks. Looking at [the source code](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/deconz/driver/driver.ts), even if `adapter_delay` set to 0 we are still setting `PROCESS_APS_QUEUES_DELAY` to 50ms and `READY_TO_SEND_TIMEOUT` to 10ms for a total of 60ms waiting to send any message.

Why do we need to wait that long, when nothing is happening in the network and presumably in the deCONZ stick? Can't we just cut all of these delays to 0 and look at the deCONZ stick's message queue (or whatever indicates it is under load) and only delay sending commands if it is busy / will drop messages if we don't? Or is this for devices in the Zigbee network which might not respond well to a high frequency of network frames?

I would be happy to run a custom version of Z2M with delays set to 0 to see if my network (which is composed of IKEA and Philips bulbs and switches) works well that way, but am not super familiar with node.js and how to set a specific package to dev mode. The way zigbee-herdsman uses TypeScript which then gets transpiled also makes it difficult to do a few changes in the dependent packages after you run `npm ci`.

I tried checking what the behaviour is for the z-stack adapters but the code is laid out differently which makes this a bit more challenging.

This is with the latest dev checkout from today, `adapter_delay` set to 0, an [IKEA E1743](https://www.zigbee2mqtt.io/devices/E1743.html) switch bound to an [IKEA LED driver](https://www.zigbee2mqtt.io/devices/ICPSHC24-30EU-IL-1.html) and the ConBee 2 adapter with firmware 0x26680700.

Details of the debug logs here:

    zigbee-herdsman:deconz:frameParser DEVICE_STATE changed: 10101010 +7s
[...]

PR #278, opened 2020-12-28
Fix delay setup on deCONZ driver #273

kind3r:
Move variable intervals to a separate function so they can be updated via `setDelay`


Koenkk:
@ChrisHae can you review this?

PR #281, opened 2021-01-04
Add startUpColorTemperature to lightingColorCtrl cluster

sjorge:
I want to have a go at implementing a convertor for StartUpColorTemperatureMireds, so this PR adds it to herdsman as it is missing. 

I noticed we are also missing CoupleColorTempToLevelMinMireds, so added that one too, although I have no plans to use that one.


Koenkk:
Thanks!

PR #283, opened 2021-01-08
Add StartUpCurrentLevel to genLevelCtrl

sjorge:
Noticed we were missing it when coming across https://github.com/Koenkk/zigbee2mqtt/discussions/5555

Shouldn't be too hard to add both OnOffTransitionTime and StartUpCurrentLevel, I sadly have nothing to test this with so hopefully the user in the discussion can.

image


Koenkk:
Thanks!

Issue #242, opened 2020-10-21
ZiGate implementation


G1K:
Discussion of the implementation of a new adapter based on the NXP chip JN516x ( JN5168 / JN5169 )
https://github.com/Koenkk/zigbee-herdsman/pull/241

What you can try as a coordinator

- Xiaomi gateway DGNWG05LM, ZHWG11LM
- End devices from xiaomi JN5169 https://github.com/Jerome-PS/JN516xSniffer

   - Xiaomi Door sensor
   - Xiaomi smart button
   - Xiaomi smart button UART1 ZiGate coordinators

- Ready modules https://www.seeedstudio.com/Mesh-Bee-Open-Source-Zigbee-Pro-Module-with-MCU-JN516-p-1751.html
- DIY module + TTL adapter
E75-2G4M10S Zigbee JN5169 aliexpress.com/item/32973411720.html

## Road map

- [x] Stability, queuing APDU, waiting for data and using SQN
- [ ] deduplication of all incoming messages by sqn https://github.com/fairecasoimeme/ZiGate/issues/346 or use [build custom firmware](https://github.com/G1K/ZiGate/actions/runs/479754512) 
- [x] implementation sendZclFrameToGroup
- [ ] supported GP - (requires stable firmware with GreenPower support and 242 endpoint ) _What is being assembled now limits the resources of the APDU or does not start after the firmware_
- [ ] OTA check, not sure if it will work
- [ ] more correct error handling and re-sending with route finding
- [ ] * Backup and restore

If you have the opportunity, please join the development and testing. I don't have much development experience especially with uart and ZiGate

Perhaps this will solve the requests
https://github.com/Koenkk/zigbee2mqtt/issues/743
https://github.com/Koenkk/zigbee2mqtt/issues/409

Hedda:
Does https://github.com/Koenkk/zigbee-herdsman/pull/241 and https://github.com/Koenkk/zigbee2mqtt/pull/4737 being merged mean that zigbee-herdsman and zigbee2mqtt now works with all modules with ZiGate firmware including the original ones from https://zigate.fr by the @fairecasoimeme project from Frédéric Dubois?

Do not own a ZiGate or other NXP Zigbee coordinator myself but know it is supported by ZHA for Home Assistant via [zigpy-zigate](https://github.com/zigpy/zigpy-zigate/)

### Known working Zigbee radio modules
- [ZiGate USB-TTL](https://zigate.fr/produit/zigate-ttl/)
[...]

PR #288, opened 2021-01-12
[zigate-adapter #242 ] fix removeDevice

G1K:
as it turned out, you need to use a different type of command. removeDevice is also suitable, but the target should be the parent of the device, not the device itself. it may have worked for some of the cases the routers

associated with #242


Koenkk:
Thanks!

PR #292, opened 2021-01-13
Allow passing custom adapter class

### LINKED ISSUE #291: Custom adapter option
## My problem

It looks like you can set a custom adapter in the `SerialPortOptions` to one of the adapters defined in this repo. I would like to write an adapter for a device but don't think my implementation is 100% spec compliant so would like to experiment with my implementation before upstreaming it for use in the community.

## Proposed solution

Change `SerialPortOptions.adapter` to be of type `'zstack' | 'deconz' | 'zigate' | Adapter`. I could then pass in my custom adapter implementation (that has extended the abstract Adapter class)

tom-sherman:
Closes #291

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -1,3 +1,5 @@
+import Adapter from "./adapter";
+
 interface NetworkOptions {
     panID: number;
     extendedPanID?: number[];
@@ -10,7 +12,7 @@ interface SerialPortOptions {
     baudRate?: number;
     rtscts?: boolean;
     path?: string;
-    adapter?: 'zstack' | 'deconz' | 'zigate';
+    adapter?: 'zstack' | 'deconz' | 'zigate' | typeof Adapter;
 }
 
 interface AdapterOptions {
--- a/.babelrc.js
+++ b/.babelrc.js
@@ -1,6 +1,13 @@
 module.exports = {
     presets: [
-      '@babel/env',
+      [
+        '@babel/preset-env',
+        {
+          targets: {
+            node: 'current',
+          },
+        },
+      ],
       '@babel/typescript',
     ],
     plugins: [
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -37,11 +37,19 @@ abstract class Adapter extends events.EventEmitter {
         const {ZStackAdapter} = await import('./z-stack/adapter');
         const {DeconzAdapter} = await import('./deconz/adapter');
         const {ZiGateAdapter} = await import('./zigate/adapter');
-        type AdapterImplementation = typeof ZStackAdapter | typeof DeconzAdapter | typeof ZiGateAdapter;
+        type AdapterImplementation = (
+            | typeof ZStackAdapter
+            | typeof DeconzAdapter
+            | typeof ZiGateAdapter
+            | typeof Adapter
+        ) & {
+            autoDetectPath?: () => Promise<string>,
+            isValidPath?: (path: string) => Promise<boolean>
+        };
 
         let adapters: AdapterImplementation[];
         const adapterLookup = {zstack: ZStackAdapter, deconz: DeconzAdapter, zigate: ZiGateAdapter};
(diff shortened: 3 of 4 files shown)


tom-sherman:
I'm trying to add a test to assert that the custom adapter is instantiated when passing it to `Adapter.create()`, but I'm getting an error:

```
 FAIL  test/adapter/custom-adapter.test.ts
  ● Test suite failed to run

    ReferenceError: regeneratorRuntime is not defined

       96 |                 debug(`Failed to validate path: '${error}'`);
       97 |             }
    >  98 |         }
          |          ^
       99 | 
      100 |         return new adapter(networkOptions, serialPortOptions, backupPath, adapterOptions);
      101 |     }

      at src/adapter/adapter.ts:98:53
      at src/adapter/adapter.ts:308:6
      at Object.<anonymous> (src/adapter/adapter.ts:312:2)
      at Object.<anonymous> (test/adapter/custom-adapter.test.ts:1:1)
```

I don't see any reason why this would be the case, any help is appreciated.

tom-sherman:
Fixed the above by adding some extra config to @babel/preset-env

src/adapter/tstype.ts
```diff
@@ -10,7 +12,7 @@ interface SerialPortOptions {
     baudRate?: number;
     rtscts?: boolean;
     path?: string;
-    adapter?: 'zstack' | 'deconz' | 'zigate';
+    adapter?: 'zstack' | 'deconz' | 'zigate' | typeof Adapter;
```
Koenkk, line 15: Somehow it feel strange the the adapter is a simple string or an entire javascript class. How do you pass this option from Zigbee2MQTT? Did you also evaluate other solutions?

tom-sherman: I agree that it's a little strange, but it adds the functionality in without a breaking change. The alternative would be to have this type as just `typeof Adapter` and so the consumer would need pass the adapter class in, but again this is a breaking change.

I'm going to propose that Zigbee2MQTT expose a node API for configuring and starting the service, instead of relying on cloning the repo and an external yaml file (which I find very unidiomatic). I think there is already an issue for this.

Koenkk: So how do you currently set it? Wouldn't it be easier to create a temp fork where you add support for this adapter and can be merged later?

tom-sherman: I personally prefer being able to extend the library is valuable, there may be some adapter that you won't accept a PR for for example. Forks are difficult to manage as well.

Koenkk: > Forks are difficult to manage as well especially when this proposed extension to the API works in a backwards compatible way.

I think this can never be assured as the adapter interface can change, so there is always some maintenance here.

I'm still wondering how this can be used from zigbee2mqtt, iobroker.zigbee etc.

tom-sherman: I'm not familiar with iobroker.zigbee. For zigbee2mqtt, a Node api would be added that would be a separate install path to the current git cloning or docker image. You could then do something to this effect:

```javascript
const zibee2mqtt = require('zibee2mqtt');

zigbee2mqtt.start({
  // Other options here
  adapter: MyCustomAdapter
})
```

Koenkk: What if we keep `adapter` as a string and allow an adapter to be loaded from another package?

e.g. `adapter: 'custom/my-own-adapter'`

```
if (serialPortOptions.adapter.startsWith('custom/')) {
  let package = serialPortOptions.adapter.replace('custom/');
  adapter =require(package);
} else {
 .. current logic ...
}
```

peterbb: Sorry for coming in from the side line. I've been looking at the code this weekend, and for me it seems that the tension is created because the class `Controller` implements Controller-logic in addition to interpreting configuration (the properties), creating and injecting dependencies, persisting the database, and, serving as the API for consumers of the library.

Im going to tinker with splitting the internals up while keeping the exposed controller API unchanged. I've joined the discord, so you can reach me there. :)

PR #299, opened 2021-01-25
Adding controls for the behaviour when the device is powered on

alphazulu:
Adding controls for the behaviour when the device is powered on
Moes MS-104BZ
Attribute 8002


Koenkk:
Thanks!

Issue #285, opened 2021-01-10
ZIGATE: Cannot read property 'length' of undefined


biau:
## What happened
Using zigate I've an error after every module pairing:
```
info  2021-01-10 00:31:56: Configuring 'PriseChambre'
error 2021-01-10 00:32:00: Failed to configure 'PriseChambre', attempt 1 (TypeError: ConfigureReporting 0x7cb03eaa00b1a4a8/3 genOnOff([{"attribute":"onOff","minimumReportInterval":0,"maximumReportInterval":3600,"reportableChange":0}], {"timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null}) failed (Cannot read property 'length' of undefined)
    at Function.fromBuffer (/app/node_modules/zigbee-herdsman/dist/zcl/zclFrame.js:151:20)
    at ZiGateAdapter.<anonymous> (/app/node_modules/zigbee-herdsman/dist/adapter/zigate/adapter/zigateAdapter.js:524:50)
    at Generator.next (<anonymous>)
    at fulfilled (/app/node_modules/zigbee-herdsman/dist/adapter/zigate/adapter/zigateAdapter.js:24:58))

```
## What did you expect to happen
Pairing without error

## How to reproduce it (minimal and precise)
Just pair a device

## Debug info
Zigbee2MQTT version: 1.17.0
Adapter hardware: Zigate
Adapter firmware version: 3.1d

Koenkk:
Looks like a duplicate of https://github.com/Koenkk/zigbee-herdsman/issues/252#issuecomment-723579075

Issue #252, opened 2020-11-08
Not correct join Gaz Detector SGMHM-I1 and Siren SRHMP-I1 via Zigate in Iobroker...


itProfi:
Good day! I try join device Gaz Detector SGMHM-I1 via Zigate in Iobroker...and have some errors..
How i can correct this situation?
![Скриншот 08-11-2020 123830](https://user-images.githubusercontent.com/6728198/98462074-5b0eca00-21c2-11eb-952e-0ddc97fc3bef.jpg)
![Скриншот 08-11-2020 124206](https://user-images.githubusercontent.com/6728198/98462075-5c3ff700-21c2-11eb-8b0a-32fa3e32b687.jpg)
![Скриншот 08-11-2020 123927](https://user-images.githubusercontent.com/6728198/98462077-5c3ff700-21c2-11eb-99ac-7ba2213961bd.jpg)
Аfter restart adapter in Iobroker, device like ok, but not work properly...
[iobroker.2020-11-08.log](https://github.com/Koenkk/zigbee-herdsman/files/5506125/iobroker.2020-11-08.log)

itProfi:
Same problem with smart siren HS2WD-E (it's not "vendor":"HEIMAN", it's russian market Tricolor)
![Скриншот 08-11-2020 130842](https://user-images.githubusercontent.com/6728198/98462215-8f36ba80-21c3-11eb-9d4e-7545d5c2e3cb.jpg)
![Скриншот 08-11-2020 130619](https://user-images.githubusercontent.com/6728198/98462217-8fcf5100-21c3-11eb-9426-5832099f924d.jpg)
[iobroker.2020-11-08_siren.log](https://github.com/Koenkk/zigbee-herdsman/files/5506136/iobroker.2020-11-08_siren.log)

Koenkk:
With what adapter is this?

itProfi:
JN5169 with zigate coordinator firmware

Koenkk:
The `Cannot read property length of undefined` indicates an error in the adapter code. @G1K or @kirovilya can you look at this?

itProfi:
@Koenkk In my case, i see, that error is in OTA Update...how i can disable temporary  this?

Koenkk:
This functionality cannot be disabled currently.

itProfi:
> This functionality cannot be disabled currently.

And how i can correct this error?

Koenkk:
[...]

PR #301, opened 2021-01-28
[zigate-adapter] Improved stability and functionality

G1K:
Most importantly, the command sending has been improved. The work is not yet 100% complete, but what has turned out is already much better and more stable. And let's solve #285 #252.

Unfortunately, I ran into some difficulties that I described here https://github.com/Koenkk/zigbee-herdsman/issues/252#issuecomment-759430863
The worst thing is that in the latest firmware when communicating with devices, APDU instances always ended https://github.com/fairecasoimeme/ZiGate/issues/348 and further communication was impossible. As a result, mating did not go through to the end.  
> Now it seems that there are commits fixing this, but I could not verify that the resulting firmware did not work on my module and the code before the change is completely identical to that in the basic example and works successfully.

Therefore, it was decided to try a clean example from NXP and move the edits a little to support raw messages and LQI. The test results show a more stable firmware version. It lacks a lot in comparison with ZiGATE, and the default settings are designed for a network of 10-20 devices, but the advantages that we managed to get.

- APDUs do not leak
- GP is supported (but so far it has been commented out if you use ZIGate firmware without .. join permit fails)
- Some messages that did not reach the uart, but were received when sniffed
- Group messages are sent via raw (maybe it is also related to 1 problem)
- The firmware is built together with GP and debug messages, in this combination ZIGate does not start on my module

Therefore, I recommend installing this firmware
https://github.com/openlumi/JN-ZigbeeNodeControlBridge-firmware/releases/latest

#242

### DIFF
--- a/src/adapter/zigate/adapter/zigateAdapter.ts
+++ b/src/adapter/zigate/adapter/zigateAdapter.ts
@@ -591,7 +591,7 @@ class ZiGateAdapter extends Adapter {
                 targetShortAddress: 0xFFFD,
                 sourceEndpoint: sourceEndpoint,
                 destinationEndpoint: endpoint,
-                profileID: sourceEndpoint === 242 ? 0xa1e0 : 0x0104,
+                profileID: /*sourceEndpoint === 242 ? 0xa1e0 :*/ 0x0104,
                 clusterID: zclFrame.Cluster.ID,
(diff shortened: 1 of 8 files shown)


Koenkk:
Thanks!

PR #304, opened 2021-02-02
Livolo support: Adding method writeUndiv to send dimmer level

whitepail:
This method added to support Livolo malformed packets. See Koenkk/zigbee2mqtt#5516


src/controller/model/endpoint.ts
```diff
@@ -310,6 +310,48 @@ class Endpoint extends Entity {
...
+
+        const log = `WriteUndiv ${this.deviceIeeeAddress}/${this.ID} ` +
+            `${cluster.name}(${JSON.stringify(attributes)}, ${JSON.stringify(options)})`;
+        debug.info(log);
+
+        try {
+            const frame = Zcl.ZclFrame.create(
+                Zcl.FrameType.GLOBAL, options.direction, options.disableDefaultResponse,
+                options.manufacturerCode, options.transactionSequenceNumber ?? ZclTransactionSequenceNumber.next(),
+                'writeUndiv', cluster.ID, payload, options.reservedBits
```
Koenkk, line 338: Seems this is the only diff compared to the `write()`. Please add an option to `options` called `writeUndiv`, if true writeUndiv will be used. This avoid a lot of code duplication. Also write a test for this in `controller.test.ts` since new code has to be covered by tests (otherwise build fails).

Koenkk:
Note to self: after this check https://github.com/Koenkk/zigbee-herdsman-converters/pull/2191/files

src/controller/model/endpoint.ts
whitepail: I'm afraid that livolo uses not only "write" (Id: 02) and "writeUndiv" (Id: 03) command types. 
In sniff log I found also "readRsp" (Id: 01) command.

Is it ok to add and option commandKey: number | string with default value "write", that would accept any value declared in foundation.ts?

Koenkk: About the `readRsp`, can't this be done with: https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/endpoint.ts#L352 ?

whitepail: At glance readResponse() can be reused, not tested yet as those packets are still need to be researched.
Do I need to rework commandKey option (number|string) to writeUndiv (boolean) or current implementation (see updated pull request) is ok?

Koenkk: Please refactor to boolean (since other commandKey make no sense)

whitepail: Done, reworked and added extra coverage test for writeUndiv.

PR #308, opened 2021-02-13
Missing locking in touchlink factoryResetFirst

peterbb:
There seems to be a missing lock in the beginning of `factoryResetFirst`. All the other methods in this class acquire then releases, except this one, which only releases.

I don't have any zigbee hardware, so I have not verified that it works.

image


Koenkk:
Thanks!

(lol @ the picture 😄 )

PR #310, opened 2021-02-14
Add missing  mandatory `options` attribute to lightingColorCtrl

sjorge:
image

Do we also define the bitmaps somewhere? If so where, I couldn't find those.


Koenkk:
> Do we also define the bitmaps somewhere? If so where, I couldn't find those.

What do you mean here? The option per bit?

sjorge:
Well given it a bit map with now names/values I though we might be storing them too somewhere. But that doesn’t seem to be the case.

Koenkk:
That's not the case indeed.

PR #312, opened 2021-02-15
Enable typescript's strictNullCheck

peterbb:
This PR enables typescript's stricter compile-time null and undefined
checks.  With this check, typescript forces you to be explicit about
when undefined and null are allowed.

All new type errors that occur because of this are marked as ignored
in this PR: there are no changes in the code. This means that there
is no change what-so-ever in how the library works.

By enabling the strict flag in this PR, it opens up for gradually
marking the use of nulls and undefines more explicit.


peterbb:
Hmm, I think this might be premature. Maybe something which should be done later.

PR #324, opened 2021-03-09
Viesmann

Koenkk:


src/zcl/definition/cluster.ts
```diff
@@ -1989,6 +1999,7 @@ const Cluster: {
             tempDisplayMode: {ID: 0, type: DataType.enum8},
             keypadLockout: {ID: 1, type: DataType.enum8},
             programmingVisibility: {ID: 2, type: DataType.enum8},
+            viessmannCustom10: {ID: 0x4000, type: DataType.enum8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
```
sjorge, line 2002: @Koenkk can you drop this one, I double checked the pcap. My note were wrong as this was on hvacThermostat not hvacUserInterfaceCfg.

So this is covered by ln 1899.

Koenkk: Done, let me know once this can be merged.

sjorge: I double checked the other ones, ID and DataType matches the ones in the pcap. So I guess it's good to go.

I also think I got window_open working so that good. (Wish I had some docs on them though)

PR #327, opened 2021-03-12
Narrow buffalo-types

peterbb:
Several occurences of the any-in-disguise Value-type in Buffalo
can be replaced by their actual type without any change in
the code that is executed.

Fixes small bugs revealed by this.


Koenkk:
Thanks!

PR #329, opened 2021-03-14
rename Viessmann attrib vc3 and vc9 to something descriptive

sjorge:
Took a bit but me and jhm47 are fairly certain we discovered what 2 attributes are.

### viessmannWindowOpenForce (ex viessmanCustom3)
It's a writable boolean and when set it forced window_open to be triggered (OO on display) and the corresponding enum in viessmannCustom0 gets set to 3.

Unlike when the window_open is triggered by the device itself, it does not disable itself after some time. However the TRV will start heating again after about 30 minutes and viessmannCustom0 will be in 4.
In the corresponding zhc PR a getter and setter are added, we also expose this.

### viessmannAssemblyMode (ex viessmanCustom9)
It's a read-only boolean, it's true when the TRV is in assembly mode (-- on display) as per the manual. The gateway only seems to read it after deviceAnnounce until it is reported as installed. Presumably by the app to display installation instructions.

In the corresponding zhc PR a getter for this is added. This is not expose as it offers no real useful information.


sjorge:
Please hold with merging until I've retested these separate PRs, this was tested in my viessman dev branch with other debugging stuff.

Koenkk:
Thanks!

Issue #333, opened 2021-03-20
Support `writeStruct`


aesedepece:
This is a follow-up to #330, focusing namely on ZCL's global command `writeStruct`.

The `writeStruct` (_Write Attributes Structured_) global command enables a client to write one or more specific elements into one or more attributes of a server. In comparison with the regular `write` (_Write Attributes_), this command is intended and specially convenient for when we need to update specific fields inside complex data structures such as an Array, Structure, Set or Bag.

> The Write Attributes Structured command is generated when a device wishes to change the values of one or more attributes located on another device. Each write attribute record SHALL contain the identifier and the actual value of the attribute, or element thereof, to be written.

Selection of which attributes and elements will be updated—and what their new value will be—is performed by using a _Write Attribute Record_. Each `writeStruct` command can carry any number of these records.

| Octets: variable | Variable                 | ... | Variable                 |
|-------------------|--------------------------|-----|--------------------------|
| ZCL header        | Write attribute record 1 | ... | Write attribute record n |

Write attribute record:

| Octets: 2            | Variable | 1                   | Variable        |
|----------------------|----------|---------------------|-----------------|
| Attribute identifier | Selector | Attribute data type | Attribute value |

> The _selector_ field specifies whether the whole of the attribute  value is to be  written, or only an individual element of it. An individual element may only be written to attributes with types of Array, Structure, Set or Bag.

| Octets: 1     | 2       | ... | 2       |
|---------------|---------|-----|---------|
| Indicator (m) | Index 1 | ... | Index m |

> When writing an element to an array or structure, the Indicator subfield indicates the number of index fields that follow it. This number is limited to the range 0 -15 (i.e., the upper 4 bits of the Indicator field are set to zero). It may be further limited by an application.
>
> If this subfield is 1 or greater, the index fields indicate which element is to be written, nested to a depth of m. For example, if the attribute is an array of arrays (or structures), then if m = 2, index 1 = 5 and index 2 = 3, the third element of the fifth element of the attribute will be written.
> Note that elements are numbered from 1 upwards for both arrays and structures.
>
> The zeroth element of an array or structure has type 16 bit unsigned integer, and holds the number of elements in the array or structure. The zeroth element of an array may optionally be written (this is application dependent) and has the effect of changing the number of elements of the array. If the number is reduced, the array is truncated. If the number is increased, the content of new elements is application dependent.
>
[...]

PR #336, opened 2021-03-22
Support `writeStruct` foundation command

aesedepece:
This:
- Implements the needed data structures and writer functions for producing valid ZCL frames that use the `writeStruct` foundation command.
- Adds test for `writeStruct` ZCL frames. I created the positive vector tests by hand, following the specification. As per the malformed Livolo packet test, it has been extracted from a capture of a Livolo official gateway.
- Adds a convenient `writeStruct`method to the `Endpoint` class. Even if one could use the `Endpoint.command` method instead, it wouldn't work without a non-trivial refactor to that method that would allow for overriding the frame type and cluster-specificness of the command. So, in line with the rest of the commands, I implemented its own convenient method here.
- Potentially unlocks supporting advanced features both for compliant and non-compliant devices.
- Add tests for `Endpoint.writeStruct`, assuming that we don't care about responses to this command (as `writeStructResp`is not implemented)

Solves #333

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -593,6 +593,8 @@ class Endpoint extends Entity {
                 this.deviceIeeeAddress, this.deviceNetworkAddress, this.ID, frame, options.timeout,
                 options.disableResponse, options.disableRecovery, options.srcEndpoint
             );
+
+            // TODO: support `writeStructuredResponse`
         } catch (error) {
             error.message = `${log} failed (${error.message})`;
             debug.error(error.message);
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -573,7 +573,7 @@ class Endpoint extends Entity {
         }
     }
 
-    public async writeStruct(clusterKey: number | string, payload: KeyValue, options?: Options): Promise<void> {
+    public async writeStructured(clusterKey: number | string, payload: KeyValue, options?: Options): Promise<void> {
         const cluster = Zcl.Utils.getCluster(clusterKey);
         options = this.getOptionsWithDefaults(
             options, true, Zcl.Direction.CLIENT_TO_SERVER, cluster.manufacturerCode);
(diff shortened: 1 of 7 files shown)


src/controller/model/endpoint.ts
```diff
@@ -573,6 +573,33 @@ class Endpoint extends Entity {
         }
     }
 
+    public async writeStruct(clusterKey: number | string, payload: KeyValue, options?: Options): Promise<void> {
```
Koenkk, line 576: Please rename this to `writeStructured` (as this is how it is called in the ZCL)

src/controller/model/endpoint.ts
```diff
@@ -573,6 +573,33 @@ class Endpoint extends Entity {
...
+
+        const log = `WriteStruct ${this.deviceIeeeAddress}/${this.ID} ` +
+            `${cluster.name}(${JSON.stringify(payload)}, ${JSON.stringify(options)})`;
+        debug.info(log);
+
+        try {
+            await Entity.adapter.sendZclFrameToEndpoint(
+                this.deviceIeeeAddress, this.deviceNetworkAddress, this.ID, frame, options.timeout,
+                options.disableResponse, options.disableRecovery, options.srcEndpoint
+            );
```
Koenkk, line 595: In the ZCL I read there is a "Write Attributes Structured Response Command", does your device send this back on write?

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee-herdsman-converters/pull/2407

src/controller/model/endpoint.ts
aesedepece: Although I guess there's very little that one can do with a "Write Attributes Structured Response" anyway, as with "Write Attributes Response", beyond letting the user know that the action failed, under the assumption that the attributes always exist and are writable :shrug:

Koenkk: > The response is quite exotic and simply acknowledges the attribute and value that has been set.

What response do you get?

src/controller/model/endpoint.ts
aesedepece: Updated!

src/controller/model/endpoint.ts
Koenkk: I reviewed the PR, no comments besides this one. Still wondering if we need something like https://github.com/Koenkk/zigbee-herdsman/blob/00f91b303a7e49d6763de6d2033a20baee15ecd9/src/controller/model/endpoint.ts#L304 here.

aesedepece: I wouldn't know what to put in there that wouldn't break the tests or coverage, as at that point the value of `Entity.adapter.sendZclFrameToEndpoint` is `null`, so I can't simply do this:

```ts
if (!options.disableResponse) {
    this.checkStatus(result.frame.Payload);
}
```

I can either leave it as it is, or go all in and implement the correct response path (decoding of `WriteStructuredResponse`). What do you think is best?

Koenkk: If your device sends a `WriteStructuredResponse` back please add it, if not add `// TODO`

aesedepece: I simply added the `// TODO`. Thanks!

PR #337, opened 2021-03-23
Tuya active reports

drzony:
Added a type for active status reports, in a PR for converters I'm adding devices using it


Koenkk:
Thanks!

PR #346, opened 2021-04-04
Use builtin number parser from Buffer class

peterbb:
Use the built in read/write methods in Buffer for 3 byte integers.

Moved `isMore` further up the class hierarchy and used it other places.


Koenkk:
Thanks!

PR #347, opened 2021-04-04
enable typescript option: noImplicitThis

peterbb:
Enables the typescript check "noImplicitThis". Seems like it caught a bug.


Koenkk:
Thanks!

PR #349, opened 2021-04-05
Small refactor. Introduce Buffalo.getWritten().

peterbb:
New method `Buffalo.getWritten()` returns a buffer with with everything up to the current position.

Simplified `Buffalo.readIeeeAddr()`.

Factored out common code in `Waitress`.


Koenkk:
Thanks!

PR #353, opened 2021-04-26
Added command lookup for getUserStatusRsp

jpg0:
This is part of the door lock cluster.

Will submit PR to zigbee-herdsman-converters to use it.


Koenkk:
Thanks!

PR #363, opened 2021-05-17
Support Wiser Smart Europe attributes and commands

pklokke:
This adds the necessary commands and attributes for my supporting PR Koenkk/zigbee-herdsman-converters#2589 for the converters

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -1912,6 +1912,12 @@ const Cluster: {
             viessmannCustom13: {ID: 0x404c, type: DataType.enum8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             viessmannCustom14: {ID: 0x404d, type: DataType.bitmap8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             viessmannCustom15: {ID: 0x404e, type: DataType.bitmap8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
+            wiserSmartPriorityLevel: {ID: 0xe000, type: DataType.uint8},
+            wiserSmartMasterShortAddress: {ID: 0xe001, type: DataType.uint16},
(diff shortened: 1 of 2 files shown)


pklokke:
Ah thanks! Do you typically prefer these fixups squashed before merge?  Still new to the workflow here. :-)

src/zcl/definition/cluster.ts
```diff
@@ -1912,6 +1912,12 @@ const Cluster: {
             viessmannCustom13: {ID: 0x404c, type: DataType.enum8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             viessmannCustom14: {ID: 0x404d, type: DataType.bitmap8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             viessmannCustom15: {ID: 0x404e, type: DataType.bitmap8, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
+            wiserSmartPriorityLevel: {ID: 0xe000, type: DataType.uint8},
```
Koenkk, line 1915: Please add the `manufacturerCode` to these attributes (like the line above).

Koenkk:
Note to self: after this merge https://github.com/Koenkk/zigbee-herdsman-converters/pull/2589

src/zcl/definition/cluster.ts
pklokke: I just tested this, unfortunately these devices don't set the MS bit in their attributes/reports, so this ends up breaking the converter. I could instead decode these as numbers directly in the converter if that is the preferred way for non-standard behaviour.

Koenkk: This is preferred indeed, manufacturer attributes are only allowed anymore when carrying a manufacturerCode

pklokke: Thanks for the guidance, I have moved the attribute decode now into the converters, so these attribute conversions aren't required any longer.

PR #365, opened 2021-05-19
Danfoss Ally Hive TRV custom command

robertalexa:
Add custom command to set heating setpoint to be instant rather than PID controlled.

This PR is required by https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592 in Converters.

I am hoping my approach is the correct one. If not, I would be happy to be corrected.


Koenkk:
Thanks!

PR #369, opened 2021-05-20
Respond on requested endpoint

pklokke:
- Ensures that whichever endpoint on the coordinator that the endpoint is
that which it is bound to

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -202,7 +202,8 @@ class Device extends Entity {
                 }
 
                 try {
-                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, response);
+                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, response,
+                        {srcEndpoint: dataPayload.dstEndpoint});
                 } catch (error) {
                     debug.error(`Read response to ${this.ieeeAddr} failed`);
                 }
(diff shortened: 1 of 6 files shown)


pklokke:
I only have a zstack adapter on hand to test with, in principle it should work with other devices (assuming they expose endpoints beyond just 1, and fill in their respective fields correctly).

src/controller/model/device.ts
```diff
@@ -202,7 +202,8 @@ class Device extends Entity {
                 }
 
                 try {
-                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, response);
+                    await endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, response,
+                        {srcEndpoint: dataPayload.dstEndpoint});
```
Koenkk, line 206: Looks good, can you rename `srcEndpoint` to `sourceEndpoint` and `dstEndpoint` to `destinationEndpoint`?

Out of interest, was there a problem you ran into because of this?

pklokke: Changed `dstEndpoint` to `destinationEndpoint`, no problem.  `srcEndpoint `is part of the options flag (currently implemented and used today), and has a pretty wide impact if we change it.

> Out of interest, was there a problem you ran into because of this?

Some of the Schneider Wiser-Smart devices I am integrating in Koenkk/zigbee-herdsman-converters#2589 have a quirk of needing to be bound to endpoint 11 on the coordinator.  I do some minor tricks in the configure section to get the bind right, but without this fix, they're getting responses back on an endpoint they weren't bound to.  Most devices won't need this (as they would by default bind to endpoint 1) but it makes sense that at this level we don't hardcode it at least.

Koenkk:
Thanks!

PR #370, opened 2021-05-20
Add Danfoss specific clusters

robertalexa:
This PR is required by https://github.com/Koenkk/zigbee-herdsman-converters/pull/2592

As per our discussion I have added Danfoss/Hive specific clusters. Please let me know if I need to do anything else on this

Thanks


src/zcl/definition/cluster.ts
```diff
@@ -1900,6 +1900,17 @@ const Cluster: {
             viessmannWindowOpenForce: {ID: 0x4003, type: DataType.boolean, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             viessmannAssemblyMode: {ID: 0x4012, type: DataType.boolean, manufacturerCode: ManufacturerCode.VIESSMAN_ELEKTRO},
             schneiderWiserSpecific: {ID: 0xe110, type: DataType.enum8, manufacturerCode: ManufacturerCode.SCHNEIDER},
+            danfossWindowOpenInternal: {ID: 0x4000, type: DataType.enum8, manufacturerCode: ManufacturerCode.DANFOSS},
```
sjorge, line 1903: @Koenkk should these be sorted on attribute id?

Koenkk: Yes, but grouping all the danfoss attributes also makes sense so this is fine.

Issue #379, opened 2021-06-06
Unable to attempt a 2nd start if the first herdsman start fails


asgothian:
Hi

Sorry for yet another issue, but the backup refactor has raised a number of issues to the point that i needed to track down an error which has been causing issues for a while now.

Consider the following scenario:
- A Herdsman Instance is created with a set of parameters.
- An attempt is made to start this instance, but during start an error causes this to not work. Important detail here is that the serial port is actually opened in this scenario.
- The startup procedure throws an error.

At this point, the client application is in an unrecoverable state:
- attempting to stop the Herdsman will throw an error at Database save (the log of this error is available below)
- If this exception is unhandled the client application will be terminated.
- If this exception is caught, the client application will remain alive, with parts of the herdsman still running (e.g. the serial port is still open)
- I could not find an option to close the connection to the hardware
- I could not find an option to update the settings and attempt a "start" again

This is causing a few headaches with the iobroker.zigbee adapter - sending it into a reconnect loop which it cannot break free of as the serial port is never closed and then throwing a confusing error which has nothing to do with the actual cause of the issue.

I see two of options to deal with this on the iobroker side, unfortunately, neither of them is ideal:
- Always take down the client application and force a full restart of the client. This is bad for cases where the error should be recoverable, e.g. is caused due to a temporary disconnect (serial port/network port)
- Distinguish between fatal errors and recoverable errors and only take down the adapter when a fatal error occurs - this requires a complex parsing of error messages not designed for this (see issue #376)

Outside of this, the remaining option would be to ensure that the 'stop' command on the herdsman side always shuts down all subsystems.

```
TypeError: Cannot read property 'getEntries' of null
    at Function.loadFromDatabaseIfNecessary (/opt/iobroker/node_modules/zigbee-herdsman/dist/controller/model/device.js:220:55)
    at Function.all (/opt/iobroker/node_modules/zigbee-herdsman/dist/controller/model/device.js:240:16)
    at Controller.databaseSave (/opt/iobroker/node_modules/zigbee-herdsman/dist/controller/controller.js:262:45)
    at Controller.<anonymous> (/opt/iobroker/node_modules/zigbee-herdsman/dist/controller/controller.js:243:18)
    at Generator.next (<anonymous>)
    at /opt/iobroker/node_modules/zigbee-herdsman/dist/controller/controller.js:27:71
    at new Promise (<anonymous>)
[...]

PR #381, opened 2021-06-07
Stop error fix

asgothian:
Fixes Issue #379 by encapsulating calls in try/catch.


.remote-sync.json
```diff
@@ -0,0 +1,25 @@
+{
+  "logger": {
+    "title": "Remote Sync"
+  },
+  "uploadOnSave": false,
+  "useAtomicWrites": false,
+  "deleteLocal": false,
+  "hostname": "ioTinker.lan",
+  "port": "22",
+  "target": "/opt/iobroker/node_modules/zigbee-herdsman",
```
Koenkk, line 10: Please .gitignore this file.

package.json
```diff
@@ -73,7 +73,7 @@
   "name": "zigbee-herdsman",
   "repository": {
     "type": "git",
-    "url": "git+https://github.com/Koenkk/zigbee-herdsman.git"
+    "url": "git+https://github.com/Asgothian/zigbee-herdsman.git"
```
Koenkk, line 76: Please revert this change

src/controller/controller.ts
```diff
@@ -281,7 +283,9 @@ class Controller extends events.EventEmitter {
 
         clearInterval(this.backupTimer);
         clearInterval(this.databaseSaveTimer);
-        await this.backup();
+        try {
+            await this.backup();
+        } catch (e) {}
         await this.adapter.stop();
```
Koenkk, line 289: In what situation does the backup fail but the adapter stop doesn't? I would expect the adapter.stop to also be in the try catch.

src/controller/model/endpoint.ts
```diff
@@ -647,7 +647,7 @@ class Endpoint extends Entity {
         } catch (error) {
             error.message = `${log} failed (${error.message})`;
             debug.error(error.message);
-            throw error;
```
Koenkk, line 650: This should not be removed.

src/controller/controller.ts
```diff
@@ -265,7 +265,9 @@ class Controller extends events.EventEmitter {
 
     public async stop(): Promise<void> {
         this.stopping = true;
-        this.databaseSave();
+        try {
```
Koenkk, line 268: Is this one really needed? (I don't see how this can fail unless e.g. wrong permission are setup). 

As an alternative we can move it below the adapter.stop().

package.json
asgothian: Sorry, this change slipped through, will be changed back.

src/controller/controller.ts
asgothian: This is **the** key item of this PR.
If the start of the adapter fails before the network is initialised, but after the serial port was opened, some internal structures are not initialised leading to a number of errors. The exact error depends on how far the initialisation has come and possibly also on timing. I was not able to pinpoint the exact placement, which is why i went for the 'catch all' approach.

```
TypeError: Cannot read property 'getEntries' of null
    at Function.loadFromDatabaseIfNecessary (/opt/iobroker/node_modules/zigbee-herdsman/dist/controller/model/device.js:220:55)
```

Moving it behind the adapter.stop() might be possible, but would only solve half the issue. The Herdsman would still fail due to its internal structures not being initialised in a situation where this should not pose an issue to begin with.

src/controller/controller.ts
asgothian: Interestingly enough, the adapter.stop() does not throw an exception in any of the cases i was able to duplicate, while the this.backup still references uninitialised data structures. I could imagine placing it in the try/catch, but am hesitant to do so as i would want an error thrown on adapter.stop to be passed to the client application, while the error on backup would only be valid for cases where the backup should succeed.

Since the error messages are not currently parseable, this becomes impossible to determine, so i opted for catching the errors from being unable to perform a backup while keeping the errors on stop for the client application to handle.

src/controller/model/endpoint.ts
asgothian: I was not aware this was still in the code. It is a remnant of an older test and definitely not wanted.

src/controller/controller.ts
Koenkk: Can you try without this try/catch and changing https://github.com/Koenkk/zigbee-herdsman/blob/941613e8f75047f529a8f0c87d57192508c9b78d/src/controller/controller.ts#L114 to 


```js
        // Database
        this.database = Database.open(this.options.databasePath);
        Entity.injectDatabase(this.database);

        // Adapter
        this.adapter = await Adapter.create(this.options.network,
            this.options.serialPort, this.options.backupPath, this.options.adapter, this.logger);
        debug.log(`Starting with options '${JSON.stringify(this.options)}'`);
        const startResult = await this.adapter.start();
        debug.log(`Started with result '${startResult}'`);
        Entity.injectAdapter(this.adapter);
```

asgothian: I'll check this later today.

src/controller/controller.ts
asgothian: I managed to duplicate the backup failure tonight too. The failure originated in the test if use of backup was supported. On the zStack Adapter (the only one where i managed to recreate it), a crash was caused here:
```
    public async supportsBackup(): Promise<boolean> {
        return this.version.product !== ZnpVersion.zStack12;
    }
```
by `this.version` being `undefined`.

Due to other possible side effects i decided to keep this try/catch in the PR, but added an error message to be posted for information.

src/controller/controller.ts
asgothian: tested - works.

src/controller/controller.ts
Koenkk: What if you change `return this.version.product !== ZnpVersion.zStack12;` to `return this.version && this.version.product !== ZnpVersion.zStack12;`

asgothian: It will remove this one error, but i am hesitant to go that route, as it is just one of many errors i have seen which i am actually able to recreate. I am not convinced that i will not hit yet another error when this.version is populated but deeper in the backup functionality something is still uninitialized.

Koenkk: @asgothian then we can fix that later, I think it's better than putting a try/catch for now.

asgothian: Ok.. i will run a test tomorrow - awaiting guests for tonight

asgothian: Updated. Was unable to test, as i could not get the error to reoccur. But i am sufficiently certain that it will work.

Koenkk:
Thanks!

PR #392, opened 2021-07-21
fix: handle socket vs serial when timeout occurs in deconz

ammmze:
This issue was brought up in Koenkk/zigbee2mqtt#4883 but went stale and unresolved. I ran into this in my kubernetes cluster where I am using Akri to run a pod that runs ser2net on the node that has the conbee ii stick. But if I move the USB stick, then the service gets recreated and gets a new IP address and we hit these timeouts but then we blow up before we get back to the code where we open the socket again, so effectively zigbee2mqtt needs to be restarted to start working again


Koenkk:
Thanks!

PR #393, opened 2021-07-22
ZNP Security Manager Table Occupancy Function Fix

castorw:
This PR fixes the security manager table occupancy function which caused APS keys not to be backed up properly. Some Z-Stack firmware versions use different empty states of the security manager entries, this caused the backup mechanism to identify an security manager entry for the first entry in address manager and not backup its key from TCLK table since there was an indication of arbitrary key being present in the APS link key table.

**What to expect?**
* If you have a network restored without APS keys where the routers expect encrypted APS communication, you will have to re-commission your network or shutdown and re-pair routers one-by-one,
* If you have a network that has been created freshly (not from corrupt backup), you can apply this patch safely and should get a correct backup afterwards.

### DIFF
--- a/src/adapter/z-stack/structs/entries/security-manager-table.ts
+++ b/src/adapter/z-stack/structs/entries/security-manager-table.ts
@@ -1,7 +1,8 @@
+/* eslint-disable max-len */
 /* eslint-disable @typescript-eslint/explicit-module-boundary-types */
 /* eslint-disable @typescript-eslint/explicit-function-return-type */
 import {Table} from "../table";
-import {securityManagerEntry} from "./security-manager-entry";
+import {SecurityManagerAuthenticationOption, securityManagerEntry} from "./security-manager-entry";
 import {StructMemoryAlignment} from "../struct";
 
 /**
@@ -14,7 +15,7 @@ export const securityManagerTable =
     (dataOrCapacity?: Buffer | Buffer[] | number, alignment: StructMemoryAlignment = "unaligned") => {
         const table = Table.new<ReturnType<typeof securityManagerEntry>>()
             .struct(securityManagerEntry)
-            .occupancy(e => ![0xfffe, 0xffff].includes(e.ami))
+            .occupancy(e => ![0xfffe, 0xffff].includes(e.ami) && !(e.ami === 0x0000 && e.authenticationOption === SecurityManagerAuthenticationOption.Default))
             .inlineHeder();
         return typeof dataOrCapacity === "number" ?
             table.build(dataOrCapacity) :
--- a/src/adapter/z-stack/structs/entries/security-manager-table.ts
+++ b/src/adapter/z-stack/structs/entries/security-manager-table.ts
@@ -16,7 +16,7 @@ export const securityManagerTable =
         const table = Table.new<ReturnType<typeof securityManagerEntry>>()
             .struct(securityManagerEntry)
             .occupancy(e => ![0xfffe, 0xffff].includes(e.ami) && !(e.ami === 0x0000 && e.authenticationOption === SecurityManagerAuthenticationOption.Default))
-            .inlineHeder();
+            .inlineHeader();
         return typeof dataOrCapacity === "number" ?
             table.build(dataOrCapacity) :
             table.build(dataOrCapacity, alignment);
--- a/src/adapter/z-stack/structs/table.ts
+++ b/src/adapter/z-stack/structs/table.ts
@@ -140,7 +140,7 @@ export class Table<R extends BuiltStruct> implements SerializableMemoryObject {
      * 
      * @param hasInlineHeader Sets whether table has record count header.
      */
-    public inlineHeder(hasInlineHeader = true): this {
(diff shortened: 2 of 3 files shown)


sjorge:
> * If you have a network restored without APS keys where the routers expect encrypted APS communication, you will have to re-commission your network,

You can sort of get around it, by using join via device and re-pairing all the routers 1 by 1, it's slow AF but it's an option if you have stuff behind a wall for example. You can't use the generic permit join as then it's whack-a-mole via which router you end up going, which can fail if it's one of those without keys

castorw:
@sjorge Yeah you are right. You can shutdown routers and re-pair them one by one. That should do the job and keep the rest of the network working.

sjorge:
> @sjorge Yeah you are right. You can shutdown routers and re-pair them one by one. That should do the job and keep the rest of the network working.

No need to power down, as you can spicy the router to use when enabling joining. It recently got fixed by Koen again.

castorw:
@sjorge Okay then. @Koenkk we may provide information ok this in release notes.

sjorge:
> @sjorge Okay then. @Koenkk we may provide information ok this in release notes.

Yeah I ended up snaking out form the coordinator, did 2 bulbs nearby, then use those to get the in wall switch, then the hallways bulb, ... not fun but better than building from scratch as I got to keep all the pesky end devices online :)

Koenkk:
I can confirm this indeed fixes the issue.

This is a sniff of pairing a device via a router after a reflash:

Screenshot 2021-07-23 at 14 37 12

The `Update Device` is now followed up by a `Transport Key` which allows to device to join, great work @castorw !

Koenkk:
Added the following to the release notes:

Screenshot 2021-07-23 at 14 49 30

castorw:
@Koenkk thanks!

Just to shed a bit more light on this. The primary reason this happens is the fact that having a R21+ spec coordinator and router causes the devices to negotiate an APS key and therefore both these devices may internally want to communicate in APS encrypted fashion. The routers from sniffs seen before indicated that the coordinator was sending only network-encrypted frames while the router joining the device kept pushing APS-encrypted payloads which weren't decipherable by the coordinator since it lost the APS link key after restore from backup - which was broken - because of the stuff fixed by this PR.

sjorge:
@Koenkk you might want to make it clear in the notes they need to update z2m and at least have it started/stopped once **before** updating the coordinator with the new firmware. If they do it the other way around they lose the APS keys.

sjorge:
Perhaps swap both bullet points too?

Koenkk:
@sjorge done, thanks

copystring:
My stick is already on 20210708. After updating to latest dev and restarting z2m do I have to re-flash the stick or is it ok like this?

sjorge:
> My stick is already on 20210708. After updating to latest dev and restarting z2m do I have to re-flash the stick or is it ok like this?

You probably have to remove and re-add all your routers, you should be able to do them one by one and use the join via device feature (you need to start at the coordinator obviously)

copystring:
> > My stick is already on 20210708. After updating to latest dev and restarting z2m do I have to re-flash the stick or is it ok like this?
> 
> You probably have to remove and re-add all your routers, you should be able to do them one by one and use the join via device feature (you need to start at the coordinator obviously)

Hm. I think you misunderstood my question.

I understand the part about having to re-add/re-pair all my routers. In the notes from Koen it says I have to flash 20210708 after updating to the latest z2m and restarting it. My stick is already on 20210708. So do I have to flash 20210708 again after updating and restarting?

Let me know if I'm unclear about this...

sjorge:
> > > My stick is already on 20210708. After updating to latest dev and restarting z2m do I have to re-flash the stick or is it ok like this?
> > 
> > 
> > You probably have to remove and re-add all your routers, you should be able to do them one by one and use the join via device feature (you need to start at the coordinator obviously)
> 
> Hm. I think you misunderstood my question.
> 
> I understand the part about having to re-add/re-pair all my routers. In the notes from Koen it says I have to flash 20210708 after updating to the latest z2m and restarting it. My stick is already on 20210708. So do I have to flash 20210708 again after updating and restarting?
> 
> Let me know if I'm unclear about this...

No if you're already on 20210708 you should be good.

Koenkk:
I noticed texas instruments released a new SDK, would be good to include this in the next fw update:

tested this firmware and looks good, @sjorge can you also give it a try?

- https://github.com/Koenkk/Z-Stack-firmware/tree/develop/coordinator/Z-Stack_3.x.0/bin

sjorge:
Did you forget to push? It’s still the one from 2 weeks ago.

~ sjorge

Koenkk:
@sjorge I've withdrawn this firmware since it caused memory errors (https://github.com/Koenkk/zigbee-herdsman/pull/393#issuecomment-886021164)

ahmaddxb:
> 
> 
> Perhaps swap both bullet points too?

where can i find this? could you paste the url, I would like to read the rest.

Koenkk:
These are the release notes of the next release, not publicly available yet.

PR #396, opened 2021-07-30
Adding to manuSpecificSinope parameters for dm2500 dimmer

onixx:
These are parameters that will work with Sinope DM2500ZB and SW2500ZB to control to status led brightness.
Also, minimumBrightness is to control the minimum output of DM2500ZB. set to avoid flickers on some lightbulbs


Koenkk:
Thanks!

Issue #376, opened 2021-06-02
Zigbee-Herdsman error handling and error reporting


asgothian:
The GitHub user @Hedda opened an issue [(Issue #1110 at iobroker.zigbee)](https://github.com/ioBroker/ioBroker.zigbee/issues/1110) to verify if/how the iobroker.zigbee adapter can use the current efforts towards ZNP Adapter Backup for interoperability between zigbee2mqtt and iobroker.zigbee and a possible chance for a near zero configuration approach for iobroker.zigbee.

I have run a check on how the iobroker.zigbee adapter could interface with these systems, and have come up with one detail which makes the data exchange of configuration data and issues difficult. As it stands now, the Zigbee Herdsman will raise issues with messages when something does not go as planned. This includes (but is not limited to) errors during setup and configuration mismatch. I could not find a method to query the nvbackup information without attempting to use these to start zigbee network.

Additionally, reacting correctly to the errors raised by the herdsman is made increasingly complex due to the need to parse the error messages raised by the herdsman.

I would propose to introduce a set of numeric error codes into the errors raised by the herdsman in order to allow for calling  system to clearly identify the error in detail. 
As we currently are not able to identify the cause of the issue there are currently just two possible reactions to the raised errors:
- catch the error and continue operation
- attempt to halt the zigbee-herdsman by calling herdsman.stop, and attempt a reinitialization after a pause.
As a result we see a number of followup errors. One issue we see is a serial port which is locked which makes reinitialization attempts fail by default. Another issue is uncaught errors from incompletely initialized structures within the zigbee-herdsman.

I am working on tracking down and eliminating these issues. This work would be made a lot easier by having unique error ID's paired with a table defining the state of the herdsman when this error is raised and how to proceed to correctly reset and possibly reinitialize and restart the zigbee-herdsman.

Is this something which can be integrated in the herdsman ? Is this of interest for the zigbee-herdsman project ? I can assist with some of the work, but my toolchain does not currently allow for developing / testing typescript code.

A.

Hedda:
Ping @castorw who made PR https://github.com/Koenkk/zigbee-herdsman/pull/303 as well as @puddly the indirectly related https://github.com/zigpy/zigpy/issues/557 format discussion for interoperability.

Koenkk:
[...]

PR #401, opened 2021-08-05
forceStartWithInconsistentAdapterConfiguration option to ignore configuration error

kirovilya:
Continuing the topic I started https://github.com/Koenkk/zigbee-herdsman/issues/376, I suggest adding the runInconsistent parameter to start the adapter in such critical cases.
In our ioBroker.zigbee project, there are many cases when it is not possible to update the stick firmware, although previous versions of the adapter worked successfully on the current firmware.
It will be for such users that they will be asked to set this parameter until they switch to a new firmware or another stick.
The user will always receive messages about configuration errors, but will be able (on their own responsibility) to launch the application and continue working.

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -17,6 +17,7 @@ interface AdapterOptions {
     concurrent?: number;
     delay?: number;
     disableLED: boolean;
+    runInconsistent?: boolean;
 }
 
 interface CoordinatorVersion {
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -17,7 +17,7 @@ interface AdapterOptions {
     concurrent?: number;
     delay?: number;
     disableLED: boolean;
-    runInconsistent?: boolean;
+    forceStartWithInconsistentAdapterConfiguration?: boolean;
 }
 
 interface CoordinatorVersion {
--- a/src/adapter/z-stack/adapter/manager.ts
+++ b/src/adapter/z-stack/adapter/manager.ts
@@ -222,7 +222,12 @@ export class ZnpAdapterManager {
                         this.logger.error(`Please update configuration to prevent further issues.`);
                         this.logger.error(`If you wish to re-commission your network, please remove coordinator backup at ${this.options.backupPath}.`);
                         this.logger.error(`Re-commissioning your network will require re-pairing of all devices!`);
-                        throw new Error("startup failed - configuration-adapter mismatch - see logs above for more information");
+                        if (this.options.adapterOptions.runInconsistent) {
+                            this.logger.error(`Run despite a configuration error.`);
+                            return "startup";
+                        } else {
+                            throw new Error("startup failed - configuration-adapter mismatch - see logs above for more information");
+                        }
                     } else {
                         /* Backup does not match adapter state */
                         this.debug.strategy("(stage-4) adapter state does not match backup");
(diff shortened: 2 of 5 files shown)


Koenkk:
I think `forceStartInconsistentAdapterConfiguration` would be a better name.

@castorw do you see any nasty side effect for this? I don't plan to allow enabling this for z2m users.

src/adapter/z-stack/adapter/manager.ts
```diff
@@ -222,7 +222,12 @@ export class ZnpAdapterManager {
                         this.logger.error(`Please update configuration to prevent further issues.`);
                         this.logger.error(`If you wish to re-commission your network, please remove coordinator backup at ${this.options.backupPath}.`);
                         this.logger.error(`Re-commissioning your network will require re-pairing of all devices!`);
-                        throw new Error("startup failed - configuration-adapter mismatch - see logs above for more information");
+                        if (this.options.adapterOptions.runInconsistent) {
+                            this.logger.error(`Run despite a configuration error.`);
```
castorw, line 226: Please update the message to `Running despite adapter configuration mismatch as configured. Please update the adapter to compatible firmware and recreate your network as soon as possible.`.

src/adapter/tstype.ts
```diff
@@ -17,6 +17,7 @@ interface AdapterOptions {
     concurrent?: number;
     delay?: number;
     disableLED: boolean;
+    runInconsistent?: boolean;
```
castorw, line 20: Rename to `forceStartWithInconsistentAdapterConfiguration`. Make it long so people see its bad.

test/adapter/z-stack/adapter.test.ts
```diff
@@ -1586,6 +1586,35 @@ describe("zstack-adapter", () => {
...
+        expect(result).toBe("resumed");
+        expect(mockLoggerError.mock.calls[0][0]).toBe("Configuration is not consistent with adapter state/backup!");
+        expect(mockLoggerError.mock.calls[1][0]).toBe("- PAN ID: configured=124, adapter=123");
+        expect(mockLoggerError.mock.calls[2][0]).toBe("- Extended PAN ID: configured=00124b0009d69f77, adapter=00124b0009d69f77");
+        expect(mockLoggerError.mock.calls[3][0]).toBe("- Network Key: configured=01030507090b0d0f00020406080a0c0d, adapter=01030507090b0d0f00020406080a0c0d");
+        expect(mockLoggerError.mock.calls[4][0]).toBe("- Channel List: configured=21, adapter=21");
+        expect(mockLoggerError.mock.calls[5][0]).toBe("Please update configuration to prevent further issues.");
+        expect(mockLoggerError.mock.calls[6][0]).toMatch(`If you wish to re\-commission your network, please remove coordinator backup at ${backupFile}`);
+        expect(mockLoggerError.mock.calls[7][0]).toBe("Re-commissioning your network will require re-pairing of all devices!");
+    });
```
castorw, line 1616: Please add expect section to validate the last message generated by the inconsistent run check.

castorw:
@Koenkk It's indeed nasty but if required by other implementations, I think we shall add it. I hope we can keep Z2M users out of this option. Let's just not provide it.

@kirovilya I have added few comments, please incorporate them and I think its good to go. However I would recommend forcing the users to reflash their adapters and recreate their networks as soon as possible.

kirovilya:
@Koenkk @castorw Done

kirovilya:
thank you

PR #413, opened 2021-09-04
Add attribute for pulse Develco specific

robertalexa:
I do not own this device. However I am trying to help with this issue https://github.com/Koenkk/zigbee2mqtt/issues/8638. I will try my best to get this tested with users before having a final PR in zhc.

This attribute is in accordance with the develco documentation (link available in the issue)
image


Koenkk:
Thanks!

PR #416, opened 2021-09-10
support tuya 0xE002

nukusinji:
support tuya 0xE002


Koenkk:
For what devices are these attributes used? I'm wondering wether these attributes should be named since I can imagine e.g. `53258` can be used for more attributes then just `alarm_temperature_max`, did you get a protocol description from TuYa?

nukusinji:
> For what devices are these attributes used? I'm wondering wether these attributes should be named since I can imagine e.g. `53258` can be used for more attributes then just `alarm_temperature_max`, did you get a protocol description from TuYa?

The device is https://www.aliexpress.com/item/1005002540793253.html?spm=a2g0s.9042311.0.0.27424c4dX7QekW

I don't know, whether another devices use this protocol.

Koenkk:
Thanks!

PR #417, opened 2021-09-10
Fix Third Reality devices don't change zoneState

Tediore:
Like Tuya devices, Third Reality devices (contact sensor and motion sensor which I will add support for soon) appear to not change zoneState after enroll which leads to a failed interview even though the devices work properly. I added a similar fix for this based on the fix for Tuya devices.

See https://github.com/Koenkk/zigbee2mqtt/issues/4655

Full disclosure: I'm not familiar with Javascript/Typescript (only have experience with Python), so I based this change on the change added for Tuya devices. It does make the interview complete, but hopefully there's nothing wrong with my syntax.

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -406,22 +406,15 @@ class Device extends Entity {
             return true;
         }
 
-        // Similar to TuYa devices, Third Reality devices don't change zoneState after enroll
-        if this.modelID?.match('^3R.*?Z') {
-            this._powerSource = this._powerSource || 'Battery';
-            this._interviewing = false;
-            this._interviewCompleted = true;
-            this.save();
-            debug.log(`Interview - quirks matched for Third Reality end device`);
-            return true;
-        }
-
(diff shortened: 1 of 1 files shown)


src/controller/model/device.ts
```diff
@@ -406,6 +406,16 @@ class Device extends Entity {
             return true;
         }
 
+        // Similar to TuYa devices, Third Reality devices don't change zoneState after enroll
+        if this.modelID?.match('^3R.*?Z') {
```
Koenkk, line 410: Instead of adding here, can you try adding it to the lookup below (line 424):

```
'^3R.*?Z': {type: 'EndDevice', powerSource: 'Battery'}
```

Tediore: Yep, that works. Thanks! I figured what I did probably wasn't ideal lol

Koenkk: Can you update this pr with that change?

Tediore: Done!

Koenkk:
Thanks!

PR #423, opened 2021-09-20
default_bind_group coordinator (901) for ikea devices

G1K:


Koenkk:
Thanks!

Note: if I understand correctly the coordinator has to be in the group to forward commands to herdsman, if yes, it would be better to change the coordinator firmware to forward all messages (even when not in the group)

G1K:
> Note: if I understand correctly the coordinator has to be in the group to forward commands to herdsman, if yes, it would be better to change the coordinator firmware to forward all messages (even when not in the group)

Yes, that's right. we encountered it on the ikea on/off switch E1743. So far, only this solution has turned out. I would be happy to modify the firmware, but I don't know how to do it in C. Although we already use the modified firmware to forward raw messages. https://github.com/openlumi/JN-ZigbeeNodeControlBridge-firmware

PR #430, opened 2021-10-05
Add 0xe001 Tuya cluster

bogdal:
This PR adds `0xe001` Tuya cluster that allows changing the switch type of the latest Tuya modules e.g: https://aliexpress.com/item/1005003051876791.html

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4023,14 +4023,6 @@ const Cluster: {
             }
         },
     },
-    manuSpecificTuyaSwitch: {
-        ID: 0xe001,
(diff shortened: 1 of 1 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -4023,6 +4023,14 @@ const Cluster: {
             }
         },
     },
+    manuSpecificTuyaSwitch: {
```
Koenkk, line 4026: Can you rename this to `manuSpecificTuya_3` (I expect this cluster will be used for more than just a switch)

Koenkk:
Thanks!

PR #440, opened 2021-10-28
Extend sendWhenActive transmission window

kennylevinsen:
sendWhenActive queues a request to be sent when data is received from an
end device. All current entries in the pending queue are sent whenever
data is received.

If requests were sent in response to the received data, these would be
end up queued for when the device sent data next time. Instead, define a
configurable window from last queue flush within which the device is
still considered active, bypassing the pending queue.

Tasks:
- [ ] Evaluate if default window size is sane


kennylevinsen:
Superseded by https://github.com/Koenkk/zigbee-herdsman/pull/447.

PR #442, opened 2021-10-29
Move led control to firmware

Koenkk:
As discussed in https://github.com/Koenkk/Z-Stack-firmware/issues/203#issuecomment-907634474 the LED control responsibility should move to the firmware. With the current implementation herdsman sets the LED on/off when joining is enabled, this breaks LEDs for custom firmwares.

With this change `setLED()` and `supportsLED()` are removed from the controller, herdsman will no longer control the LEDs when joining is enabled/disabled.

For ZStack adapters:
- For firmwares having `revision >= 20211029`:
  - `ledControl` will not be called when joining is enabled/disabled
  - If `disable_led: true` is set LED control is called with `{ledid: 0xFF, mode: 5}`. When `ledControl` is called with `mode: 5` the firmware should disable all leds.

- For firmwares having `revision < 20211029`, the behaviour isn't changed `ledControl` is still called when joining is enabled/disabled. (this is to avoid a breaking change on firmwares not turning on leds when joining is enbaled/disabled)
- I will provide new firmwares which have internal LED control when joinig is enabled/disabled:
  - Z-Stack_Home_1.2: https://github.com/Koenkk/Z-Stack-firmware/tree/2671dbaa484b4a1ac724e77cc71f2155fbbf7628/coordinator/Z-Stack_Home_1.2
  - Z-Stack_3.x.0: TODO
- For Zigbee2MQTT this requires: https://github.com/Koenkk/zigbee2mqtt/pull/9404

Please review: @egony @kirovilya

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -108,10 +108,6 @@ abstract class Adapter extends events.EventEmitter {
 
     public abstract reset(type: 'soft' | 'hard'): Promise<void>;
 
-    public abstract supportsLED(): Promise<boolean>;
-
-    public abstract setLED(enabled: boolean): Promise<void>;
-
     public abstract supportsBackup(): Promise<boolean>;
 
     public abstract backup(): Promise<Models.Backup>;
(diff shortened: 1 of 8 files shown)


egony:
Hi!
Description looks reasonable, I think all is right...

Issue #445, opened 2021-11-01
Request timing for Sleepy End Devices


kennylevinsen:
*This summarizes information from https://github.com/Koenkk/zigbee-herdsman-converters/pull/3240*

zigbee-herdsman currently treat Sleepy End Devices (SEDs) the same as any other, assuming the device will always be responsive with some simple retry logic. To fix certain SEDs that have shown problems with this behavior, a sendWhenActive option is utilized in a few places to delay a send attempt until next time data is received, which seem to work but with caveats.

Instead, we should have some general handling of Sleepy End Devices that can be enabled for all such end devices.

### Polling intervals

Sleepy End Devices (SEDs) turn off their radio, making them unable to receive messages most of the time. They instead poll their parent for messages using a MAC data request. The parent is only required to hold a *single* message for up to 7.68 seconds.

SEDs have two polling intervals:

- Long polling, defaulting to once every 5 seconds. This polling mode is the normal mode of operation.

- Short polling, defaulting to once every 0.5 seconds. Short polling is intended for whenever the device wishes to be responsive, or when "fast poll" is requested through the Poll Control Cluster.

Long polling for battery-powered devices that do not need reliable Rx at all times *may* be much longer than the 7.68 second TTL for messages queued with their parent - which means that there is no guarantee that a request sent at an arbitrary point in time to a sleepy end device will be successfully delivered.

We are not directly informed of what polling mode a device is in, nor of when the device polls.

The Sleepy End Device must poll faster than the End Device Poll Timeout in order to not be aged out of the network. This timeout is communicated through an End Device Timeout Request.

### Poll control cluster

The Poll Control Cluster also provides us with the following:

- A check-in command, sent by default once every 1 hour. This *should* be longer than the long poll interval.

   When a client (which in this case would be zigbee2mqtt) binds to this cluster, we should receive this command at the specified interval, and should respond with whether we want the server to enter fast poll, and for how long. We also have the ability to cancel a fast poll early with a fast poll stop command. While the server is in fast poll, it will use the short polling interval.

- The ability to configure and query check-in, long poll and short poll intervals, as well as minimums and timeouts.

## The problem

[...]

PR #446, opened 2021-11-03
Check-in command support

kennylevinsen:
This implements support for receiving and responding to check-in commands. If there are any pending messages, fast poll is requested and the queue is emptied. Otherwise, fast poll is declined.

Notes:
- The device configure is still responsible for binding the cluster, as the messages will otherwise never arrive.
- Implicit check-in is disabled if we receive a check-in message, but maybe it would be nicer if it was controlled together with the binding setup (device.enablePollClusters())?
- The check-in command still bubbles up as a message, so if the device lacks a genPollCtrl commandCheckIn converter, zigbee2mqtt will log an error. We could let the message be consumed, but the "checkin_presence" converter currently relies on it.

Comments welcome.

Related to https://github.com/Koenkk/zigbee-herdsman/issues/445

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -429,7 +429,7 @@ class Controller extends events.EventEmitter {
 
         device.updateLastSeen();
         this.emit(Events.Events.lastSeenChanged, {device: device} as Events.LastSeenChangedPayload);
-        device.sendPendingRequests();
+        device.implicitCheckin();
 
         if (device.networkAddress !== payload.networkAddress) {
             debug.log(`Device '${payload.ieeeAddr}' announced with new networkAddress '${payload.networkAddress}'`);
@@ -532,7 +532,7 @@ class Controller extends events.EventEmitter {
 
         device.updateLastSeen();
         this.emit(Events.Events.lastSeenChanged, {device: device} as Events.LastSeenChangedPayload);
-        device.sendPendingRequests();
+        device.implicitCheckin();
 
         if (!device.interviewCompleted && !device.interviewing) {
             const payloadStart: Events.DeviceInterviewPayload = {status: 'started', device};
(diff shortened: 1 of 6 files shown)


Koenkk:
> Implicit check-in is disabled if we receive a check-in message, but maybe it would be nicer if it was controlled together with the binding setup (device.enablePollClusters())?

Good point, we can do this in the `configure()` and save this in the database, now `useImplicitCheckin` is not persisted across restarts.

> The check-in command still bubbles up as a message, so if the device lacks a genPollCtrl commandCheckIn converter, zigbee2mqtt will log an error. We could let the message be consumed, but the "checkin_presence" converter currently relies on it.

Can be solved easily by adding `genPollCtrl` to https://github.com/Koenkk/zigbee2mqtt/blob/b3e9afaf3ee55e2d62f6775dc0c62e8550b8fef1/lib/extension/receive.ts#L117


I see the tests are currently failing, can you update these and make sure the new code is covered? Use `npm tst` and `npm run test-with-coverage`.

kennylevinsen:
> Good point, we can do this in the configure() and save this in the database, now useImplicitCheckin is not persisted across restarts.

Take a look at `configurePollControl` and see what you think. I find the interface slightly awkward, but on the other hand I feel like it would be very manual to use poll control if people had to bind and flip implicitCheckin themselves.

> I see the tests are currently failing, can you update these and make sure the new code is covered? Use npm tst and npm run test-with-coverage.

Done, although tests fail on master right now.

Koenkk:
Sorry for the failing tests, I've fixed this now in master.
Reviewed the code, looks good to me so if you rebase and CI checks pass I will merge this.

kennylevinsen:
> Sorry for the failing tests, I've fixed this now in master.

It happens - I was going to submit a fix, but you beat me to it. :)

Rebased. Now we're just failing a lint on a file that hasn't been changed.

Koenkk:
> Rebased. Now we're just failing a lint on a file that hasn't been changed.

It fails in `src/zcl/zclFrame.ts` which has been changed

kennylevinsen:
> It fails in `src/zcl/zclFrame.ts` which has been changed

... Apologies, I must have forgotten to put my eyeballs back in. You're right, and fixed.

Koenkk:
Haha 👀 , thanks!

PR #447, opened 2021-11-03
endpoint: Try immediate send before queing pending

kennylevinsen:
sendWhenActive always queues the request for next time the pending queue
was flushed. This has the unfortunate side-effect of delaying sends
until next time the end device talks, even if the device is currently
active.

Instead, try to send the request immediately, and only fall back to the
pending queue if that failed or if something was already queued.

Notes:
- The detection for transaction expired is a little rough. We also still go through normal retry first, which might be a bit wasteful.
- We will still fail the entire pending queue if we try to send it at a bad time, instead of detecting this and putting them back in the queue.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -278,7 +278,7 @@ class Endpoint extends Entity {
     }
 
     private async queueRequest<Type>(func: () => Promise<Type>): Promise<Type> {
-        debug.info(`Writing to ${this.deviceIeeeAddress}/${this.ID} when active`);
+        debug.info(`Sending to ${this.deviceIeeeAddress}/${this.ID} when active`);
         return new Promise((resolve, reject): void =>  {
             this.pendingRequests.push({func, resolve, reject});
         });
@@ -299,7 +299,7 @@ class Endpoint extends Entity {
             }
         }
 
-        // If we got an expired transaction, the device is likely sleeping.
+        // If we got a failed transaction, the device is likely sleeping.
         // Queue for transmission later.
         return this.queueRequest(func);
     }
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -265,26 +265,43 @@ class Endpoint extends Entity {
         return this.pendingRequests.length > 0;
     }
 
-    public sendPendingRequests(): void {
-        [...this.pendingRequests].forEach(async (r) => {
-            this.pendingRequests.splice(this.pendingRequests.indexOf(r), 1);
+    public async sendPendingRequests(): Promise<void> {
+        while (this.pendingRequests.length > 0) {
+            const r = this.pendingRequests.shift();
             try {
                 const result = await r.func();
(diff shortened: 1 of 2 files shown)


src/controller/model/endpoint.ts
```diff
@@ -261,26 +261,37 @@ class Endpoint extends Entity {
...
-            });
-        } else {
-            return func();
+    private async sendRequest<Type>(func: () => Promise<Type>, sendWhenActive: boolean): Promise<Type> {
+        if (this.pendingRequests.length === 0) {
+            try {
+                return await func();
+            } catch(error) {
+                // TODO: Improve check for expired transaction
+                if (!sendWhenActive || error.code !== 240) {
```
Koenkk, line 282: I would not rely on the error code here, they might be different across different adapters

kennylevinsen: Should we formalize an error type, or should we just throw it in pending regardless of the error when sendWhenActive is set?

Koenkk: - sendWhenActive == false -> throw error
- sendWhenActive == true -> dont throw error

kennylevinsen: Done, and expanded the tests to cover the new cases.

src/controller/model/endpoint.ts
```diff
@@ -261,26 +261,36 @@ class Endpoint extends Entity {
...
 
-    public async sendRequest<Type>(func: () => Promise<Type>, sendWhenActive: boolean): Promise<Type> {
-        if (sendWhenActive) {
-            return new Promise((resolve, reject): void =>  {
-                this.pendingRequests.push({func, resolve, reject});
-            });
-        } else {
-            return func();
+    private async sendRequest<Type>(func: () => Promise<Type>, sendWhenActive: boolean): Promise<Type> {
+        if (this.pendingRequests.length === 0) {
```
Koenkk, line 277: Why is the request only send when pendingRequests are empty? I would expect that a request with `sendWhenActive = false` would always be send directly.

kennylevinsen: Hmm, true. I'll change that.

The idea was that if things had started queuing, we'd keep queuing for some amount of ordering, but it does behave a bit odd of you're asking for no queueing.

kennylevinsen:
Note: I'm currently missing test coverage, will get that sorted tomorrow.

kennylevinsen:
Test coverage fixed. Had some trouble getting the timing right in the test (sending data to wake things up *after* the request has been queued, but before we end up stalling on an await), but found a solution that's good enough.

Koenkk:
@kennylevinsen I've rebased it after https://github.com/Koenkk/zigbee-herdsman/pull/446 but now the coverage is failing, could you check this?

kennylevinsen:
Yeah, the new checkin test had to be adjusted slightly for the sendWhenActive behavior change. Fixed!

Koenkk:
Awesome, thanks! 🏅

PR #448, opened 2021-11-05
Add support for empty network address in unified backup

castorw:
Some coordinators (other than Z-Stack) do not allow for backup of network addresses of devices currently known to coordinator. For this reason the open coordinator backup format now allows for `null` network address. As per https://github.com/zigpy/open-coordinator-backup/pull/8 the network addresses may be replaced with invalid node address `0xFFFE` in restore time without affecting functionality of the network.


Koenkk:
Thanks!

PR #454, opened 2021-11-11
device: Enable poll control automatically

kennylevinsen:
Automatically bind the poll control cluster on all endpoints where it is
available. Legacy checkin behavior is disabled if any endpoint supported
the cluster.

This also expands the poll control tests slightly.


Koenkk:
Thanks!

PR #458, opened 2021-11-16
Add 3 missing attributs in seMetering cluster

ralmn:
Add 3 missing attributs in seMetering cluster according page 715 and 716 of https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf


Koenkk:
Thanks!

PR #461, opened 2021-11-26
manuSpecificTuya_1 as 0xE000 (57344) cluster

kirovilya:
girier module cluster
image

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4113,6 +4113,12 @@ const Cluster: {
         commands: {},
         commandsResponse: {},
     },
+    manuSpecificTuya_1: {
+        ID: 0xE000,
+        attributes: {},
+        commands: {},
+        commandsResponse: {},
+    },
     manuSpecificTuya_3: {
         ID: 0xe001,
         attributes: {


src/zcl/definition/cluster.ts
```diff
@@ -4113,6 +4113,12 @@ const Cluster: {
         commands: {},
         commandsResponse: {},
     },
+    manuSpecificTuya_1: {
+        ID: 0xE000,
+        attributes: {},
+        commands: {},
+        commandsResponse: {},
```
Koenkk, line 4120: Since there are no attributes/commands/commandReponses how is this cluster used?

kirovilya:
not used yet, but crashes if it cannot find such a cluster

Koenkk:
> not used yet, but crashes if it cannot find such a cluster

what crashes?

kirovilya:
sorry for bad picture (not my):
image

but it is not last herdsman...

kirovilya:
image

kirovilya:
this error appears 10 seconds after pairing a new device. apparently a message comes with this cluster - it falls like in the picture.
but if I add an empty cluster description, it will work fine.

I did not look in to the incoming message - this is happening for the user, not for me.

Koenkk:
I believe this issue is already fixed in the latest herdsman due to https://github.com/Koenkk/zigbee-herdsman/commit/110b67a44bcefbafc3747ad0d8a0938eaa0dc40f , could you check that?

kirovilya:
you are right. it works.

PR #483, opened 2021-12-22
Add support for additional data point values in Tuya payloads

### LINKED ISSUE #478: TuYa dataResponse: Support for device that sends multiple datapoints at once
I am trying to enable support for a TuYa WLS-100z water leak detector. As expected, it uses the `manuSpecificTuya` cluster.

Using the Tuya IoT Platform, I found out that it reports DPs 1 and 4. However, `zigbee-herdsman` only shows DP 1 data and no DP 4 data at all.

I reckon that the device sends two CmdData Units on the [serial protocol](https://developer.tuya.com/en/docs/iot/tuya-zigbee-module-uart-communication-protocol?id=K9ear5khsqoty#title-14-Send%20commands) and these end up in a single `dataResponse` ZCL frame:

```
npm[4523]: 2021-12-18T17:16:06.285Z zigbee-herdsman:adapter:zStack:znp:AREQ <-- AF - incomingMsg - {"groupid":0,"clusterid":61184,"srcaddr":33320,"srcendpoint":1,"dstendpoint":1,"wasbroadcast":0,"linkquality":45,"securityuse":0,"timestamp":1347288,"transseqnumber":0,"len":18,"data":{"type":"Buffer","data":[9,6,1,0,1,1,4,0,1,0,4,2,0,4,0,0,0,90]}}
```

The received DP data actually contains both data points:

|Data|Value|
|---|---|
|0,1 | Sequence number|
|1 | DP 1|
|4| Datatype: enum|
|0,1| Length: 1 byte|
|0| Value: water detected|
|4| DP 4|
|2| Type: integer|
|0,4| Length: 4 bytes|
|0,0,0,90| Integer value: 90 (% battery)|

However, zigbee-herdsman apparently only expects a single DP in the data, see https://github.com/Koenkk/zigbee-herdsman/blob/902363b95c2a9a21a80992e4b2838a52adb9de79/src/zcl/definition/cluster.ts#L4115

I have no knowledge of zigbee-herdsman (or zigbee-hardsman-converters). So, I would love if someone with more knowledge could have a look at this or could give me some hints on how to support such payloads/devices.

gmbnomis:
Although most TuYa devices send only one data point value in a data
report/command, it is possible to include values for multiple data points in
such a payload.

Extend the definitions of those receiving commands to decode all values
present.  Store them in the `dpValues` array.  Additionally, remove the `fn`
field: It is just the high byte of the length of the data buffer and isn't
used by converters.

Change the definitions of the sending commands to use an array of DP values
as well.

**NOTE:**
- This is a breaking change to the tuya library functions in zigbee-herdsman-converters.
- This breaks all existing Tuya from Zigbee converters using `dataResponse`,
  `dataReport`, or `activeStatusReport`.  They need to be adapted to use the
  first entry of `dpValues`

Closes #478

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4004,11 +4004,7 @@ const Cluster: {
                 ID: 0,
                 parameters: [
                     {name: 'seq', type: DataType.uint16},
-                    {name: 'dp', type: DataType.uint8},
-                    {name: 'datatype', type: DataType.uint8},
-                    {name: 'length_hi', type: DataType.uint8},
-                    {name: 'length_lo', type: DataType.uint8},
-                    {name: 'data', type: BuffaloZclDataType.LIST_UINT8},
+                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                 ],
             },
             /**
@@ -4040,11 +4036,7 @@ const Cluster: {
                 ID: 4,
                 parameters: [
                     {name: 'seq', type: DataType.uint16},
-                    {name: 'dp', type: DataType.uint8},
-                    {name: 'datatype', type: DataType.uint8},
-                    {name: 'length_hi', type: DataType.uint8},
-                    {name: 'length_lo', type: DataType.uint8},
-                    {name: 'data', type: BuffaloZclDataType.LIST_UINT8},
+                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                 ],
             },
 
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4104,10 +4104,7 @@ const Cluster: {
                 ID: 1,
                 parameters: [
                     {name: 'seq', type: DataType.uint16},
-                    {name: 'dp', type: DataType.uint8},
-                    {name: 'datatype', type: DataType.uint8},
-                    {name: 'fn', type: DataType.uint8},
-                    {name: 'data', type: DataType.octetStr},
+                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                 ],
             },
             /**
@@ -4117,10 +4114,7 @@ const Cluster: {
                 ID: 2,
                 parameters: [
                     {name: 'seq', type: DataType.uint16},
-                    {name: 'dp', type: DataType.uint8},
-                    {name: 'datatype', type: DataType.uint8},
-                    {name: 'fn', type: DataType.uint8},
-                    {name: 'data', type: DataType.octetStr},
+                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                 ],
             },
 
@@ -4134,10 +4128,7 @@ const Cluster: {
                 ID: 6,
                 parameters: [
                     {name: 'seq', type: DataType.uint16},
-                    {name: 'dp', type: DataType.uint8},
-                    {name: 'datatype', type: DataType.uint8},
-                    {name: 'fn', type: DataType.uint8},
-                    {name: 'data', type: DataType.octetStr},
+                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                 ],
             },
             /**
--- a/src/zcl/buffaloZcl.ts
+++ b/src/zcl/buffaloZcl.ts
@@ -73,6 +73,12 @@ interface ExtensionFieldSet {
     extField: TsType.Value[];
 }
 
+interface TuyaDataPointValue {
+    dp: number;
+    datatype: number;
+    data: Buffer;
+}
+
 class BuffaloZcl extends Buffalo {
     private readUseDataType(options: BuffaloZclOptions): TsType.Value {
         return this.read(options.dataType, options);
@@ -276,6 +282,24 @@ class BuffaloZcl extends Buffalo {
         }
     }
 
+    private readListTuyaDataPointValues(): TuyaDataPointValue[] {
+        const value = [];
+
+        while (this.isMore()) {
+            try {
+                const dp = this.readUInt8();
+                const datatype = this.readUInt8();
+                const len_hi = this.readUInt8();
+                const len_lo = this.readUInt8();
+                const data = this.readBuffer(len_lo + (len_hi << 8));
+                value.push({dp, datatype, data});
+            } catch (error) {
+                break;
+            }
+        }
+        return value;
+    }
+
     private readUInt40(): [number, number] {
         const lsb = this.readUInt32();
         const msb = this.readUInt8();
@@ -390,6 +414,8 @@ class BuffaloZcl extends Buffalo {
             return this.readListThermoTransitions(options);
         } else if (type === 'GDP_FRAME') {
             return this.readGdpFrame(options);
+        } else if (type === 'LIST_TUYA_DATAPOINT_VALUES') {
+            return this.readListTuyaDataPointValues();
         } else if (type === 'uint40') {
             return this.readUInt40();
         } else if (type === 'uint48') {
--- a/src/zcl/buffaloZcl.ts
+++ b/src/zcl/buffaloZcl.ts
@@ -300,6 +300,17 @@ class BuffaloZcl extends Buffalo {
         return value;
     }
 
+    private writeListTuyaDataPointValues(dpValues: TuyaDataPointValue[]): void {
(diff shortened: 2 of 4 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -4120,6 +4120,7 @@ const Cluster: {
                     {name: 'datatype', type: DataType.uint8},
                     {name: 'fn', type: DataType.uint8},
                     {name: 'data', type: DataType.octetStr},
+                    {name: 'additionalValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
```
gmbnomis, line 4123: In order not to update all of the existing converters, I chose to store additional DP values in an extra field.

Alternatively (and cleaner) one could just do:

```typescript
            dataResponse: {
                ID: 1,
                parameters: [
                    {name: 'seq', type: DataType.uint16},
                    {name: 'dpValues', type: BuffaloZclDataType.LIST_TUYA_DATAPOINT_VALUES},
                ],
            },
```

But this would break all existing TuYa converters (and, if I adapt these, I don't have a way to test them).

Koenkk:
@hacker-cb could you review this?

hacker-cb:
@gmbnomis where you found about multiply dp in one command?

There is no such information in [official tuya document](https://developer.tuya.com/en/docs/iot-device-dev/tuya-zigbee-universal-docking-access-standard?id=K9ik6zvofpzql#subtitle-6-Private%20cluster).

image

gmbnomis:
> @gmbnomis where you found about multiply dp in one command?

Please have a look into the corresponding issue #478. I have a water leakage sensor that sends such a payload. 

You are right, the Zigbee part does not mention it explicitly, but if you look at the [serial protocol](https://developer.tuya.com/en/docs/iot/tuya-zigbee-module-uart-communication-protocol?id=K9ear5khsqoty#title-14-Send%20commands) it is possible to send multiple DP values back-to-back. ("Sending multiple commands are supported in obj type data points").

I reckon that in my case, these back-to-back data point values are just put into the Zcl "DP data"

hacker-cb:
@gmbnomis seems you are right, I can accept version that MCU DPs data is transfered transparently via Zigbee.

I think right way is completely replace single-dp behaviour with multi-dp because may be other devices also use multi-dp format and currently we just ignore this.

Of cause, need to refactor converters in this case.
I think it will be better to use class-based dp everywhere.

After brief lookup, for incoming dp:

1. Just need to change `fromZigbee.js` and `lidl.js`.
2. ...?

For outgoing dp:

1. Add `sendMultiDataPoints()` method to `lib/tuya.js`, based on `sendDataPoint()`.
2. Call it from `sendDataPoint()` for backward-compatibility.
3. Completely replace `sendDataPoint[Value|sendDataPointBool|sendDataPointEnum|...]() ` with class-based dp.
4. ...?

gmbnomis:
> Of cause, need to refactor converters in this case. I think it will be better to use class-based dp everywhere.

@hacker-cb Makes sense. Essentially, we would convert these by changing `msg.data....` into `msg.data.dpValues[0]...`?

 
> After brief lookup, for incoming dp:
> 
> 1. Just need to change `fromZigbee.js` and `lidl.js`.

There is also `tuya_thermostat_weekly_schedule` in `legacy.js`

> 2. ...?

I would propose:

2. Get rid of the `fn`field. According to the documentation, it is just the high byte of the length and it's used nowhere but in `tuya_data_point_dump`.
3. Adapt `tuya_data_point_dump` to be able to dump multiple data points. Adapt `read_tuya_dump.py` (I have already done this)
4. Adapt the documentation https://www.zigbee2mqtt.io/advanced/support-new-devices/02_support_new_tuya_devices.html

Koenkk:
Sounds good to me, I think such change has quite a risk so once ready I will merge it after the next release on 1 January.

gmbnomis:
Great!

@hacker-cb As you have looked into the outgoing part in more detail, should we split up the work? (I can continue to work on the incoming part, which will mostly happen in other repositories I suppose)

hacker-cb:
I think at least this things of the outgoing part should be merged together with incoming to avoid double refactoring of the `zigbee-herdsman` project:

> * Add `sendMultiDataPoints()` method to `lib/tuya.js`, based on `sendDataPoint()`.
> * Call it from `sendDataPoint()` for backward-compatibility.

@Koenkk this PR should be merged very quickly to avoid situations [like this](https://github.com/Koenkk/zigbee-herdsman-converters/pull/3571#issuecomment-1000567703).

Koenkk:
Does this pr also require any changes to zigbee-herdman-converters?

gmbnomis:
> Does this pr also require any changes to zigbee-herdman-converters?

@Koenkk Yes, this PR totally breaks all "from Zigbee" Tuya converters. The necessary adaptations are at https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572.

As you said before, this change  is quite risky (although I tried to be as conservative as possible), as I can't test any of those converters. So you might want to merge it after releasing.

Additional changes will be necessary for the documentation on how to add TuYa devices. (I plan to provide these as well, but haven't started working on those yet)

This is the input side. Then there is the output (sending) side of things. @hacker-cb I am not sure if I understand your statement on the output side above. In your view, should the changes to the sending methods become part of this PR, or should this PR be merged first to get the output changes on top as quickly as possible?

hacker-cb:
@gmbnomis 

I mean that all `zigbee-herdsman` `cluster.ts` modifications of the both input/output parts should be together with input modifications in one PR.
`zigbee-herdsman-converters` `sendDataPoint()` should be modified to match new cluster and should be in one PR with inputs.

Koenkk:
@gmbnomis 

> So you might want to merge it after releasing.

I will set a reminder for 2 January, will ping you to update the zigbee-herdsman-converters pr then.

gmbnomis:
I have done the input and output side now, see also https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572.

> For outgoing dp:
> 
> 1. Add `sendMultiDataPoints()` method to `lib/tuya.js`, based on `sendDataPoint()`.

done

> 2. Call it from `sendDataPoint()` for backward-compatibility.

done

> 3. Completely replace `sendDataPoint[Value|sendDataPointBool|sendDataPointEnum|...]() ` with class-based dp.

I have not done this: Looking at the existing "to Zigbee" converters, the main use case will still be to send a single DP.
I think we should not make this main use case more complicated and keep the "convenience" functions to send a single DP value.


@hacker-cb Could you have a look? Additionally, if you have a TuYa device that accepts outgoing DP commands, could you test this with a real device? (I verified that the outgoing payloads are the same, but I can't verify with a real device)

hacker-cb:
> I have not done this: Looking at the existing "to Zigbee" converters, the main use case will still be to send a single DP.
> I think we should not make this main use case more complicated and keep the "convenience" functions to send a single DP value.

I agree with you to keep convenience functions.
Outgoing part looks nice. For the incoming part - I left comment in review.

> Additionally, if you have a TuYa device that accepts outgoing DP commands, could you test this with a real device? (I verified that the outgoing payloads are the same, but I can't verify with a real device)

I can't test on real Tuya devices because I'm not in office till end of the NY holidays. May be @Koenkk can ask somebody else...

Koenkk:
@hacker-cb no rush, will merge this after new years eve.

hacker-cb:
Also there is one more thing left to do:

I wrote comment in review of the [Processing all dp data in incoming message with for/foreach loop](https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572).

So, @gmbnomis - what about that?

gmbnomis:
@hacker-cb Sorry, I can't find a comment at https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572?

hacker-cb:
@gmbnomis 

image

gmbnomis:
@hacker-cb Ah, that looks like a "Pending" comment that is only visible to you. You will need to submit the review to make it visible.

hacker-cb:
@gmbnomis, sorry, submitted, try now :)

Koenkk:
@gmbnomis I plan to merge everything tomorrow:

- are you ready?
- Could you link all prs I need to merge? (to make sure I don't miss any)

gmbnomis:
> * are you ready?

Yes, all PRs are final from my point of view

> * Could you link all prs I need to merge? (to make sure I don't miss any)

Overall, there are three PRs:
- This one: https://github.com/Koenkk/zigbee-herdsman/pull/483/
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/3572
- https://github.com/Koenkk/zigbee2mqtt.io/pull/1095/

(I will post the changes for the actual device needing this feature after they are merged)

Koenkk:
@gmbnomis merged all, big  thanks for working on this!

PR #486, opened 2021-12-30
Add assert descriptions

mtandersson:
Running z2m i get in my not so stable setup a lot of assertions like 
```
Zigbee2MQTT:debug 2021-12-30 07:13:13: Saving state to file /app/data/state.json
Assertion failed
Assertion failed
```

This just add a hint of whats going wrong might help the user.


Koenkk:
Thanks!

PR #488, opened 2022-01-05
Improve handling of greenpower devices on larger networks

pklokke:
- Allow forwarding of unicast encapsulated GPDF if permitjoin is used on single
router/proxy
- This allows GPDs to not spam broadcast traffic in larger networks, which
can bog down certain lighting scenarios
- Permit join using all routers retains existing behaviour
- Remove GPD proxy entries from routers when removed
- Extended frames, and added mechanism for optional attributes


src/controller/greenPower.ts
```diff
@@ -47,44 +47,71 @@ class GreenPower extends events.EventEmitter {
...
+
+                const frame = Zcl.ZclFrame.create(
+                    Zcl.FrameType.SPECIFIC, Zcl.Direction.SERVER_TO_CLIENT, false,
+                    null, ZclTransactionSequenceNumber.next(), 'pairing', 33, payload
+                );
+
+                await this.adapter.sendZclFrameToAll(242, frame, 242);
+            }
+            else {
+                const coordinator = await this.adapter.getCoordinator();
```
Koenkk, line 68: Could you explain this if/else logic?

src/controller/controller.ts
```diff
@@ -576,6 +582,14 @@ class Controller extends events.EventEmitter {
         device.implicitCheckin();
         device.linkquality = dataPayload.linkquality;
 
+        // Process gpDevice as actual device in event of encapsulated notification via unicast
+        if (gpDevice) {
+            gpDevice.updateLastSeen();
+            gpDevice.implicitCheckin();
```
Koenkk, line 588: I don't think we have to call implicit checking for gp devices, this is only relevant for devices where commands can be send to.

Koenkk:
CI is failing because of the test coverage.

You can check it locally by running `npm run test-with-coverage`, coverage report can be found in coverage/lcov-report/index.html. Some existing green power tests: https://github.com/Koenkk/zigbee-herdsman/blob/8abf23aa0437c53c3d0ba32da276732dce173948/test/controller.test.ts#L3630

src/controller/greenPower.ts
pklokke: When permit join below is used towards an individual proxy (i.e. network address isn't null) I specified that the commissioning should be encapsulated as a unicast (0x2b, rather than 0x0b).  This means that when the commissioning packet is received here in this mode, it can be simply checked if it was broadcast or not.  If its not broadcast (i.e. unicast) I configure the payload for the pairing packet to instruct the proxy *which forwarded the commisioning packet* to forward messages via unicast for the given GPD.

This lets a person commissioning to choose the proxy (or if repeated proxies) they would like to use for GPDF forwarding.  This existing behaviour (groupcasting on all proxies if a packet is received) works, but scales poorly on large networks.

src/controller/controller.ts
pklokke: I can remove this, no problem.

src/controller/greenPower.ts
Koenkk: Thanks for the explanation!

pklokke:
Still chipping away at this, but a little time-poor right now, I'll let know once I've completed the new test cases

pklokke:
Took a bit longer than expected, needed to refactor some of the optional attribute things, and had to develop the test cases more.  Now it works in theory and in practice I believe.

Whenever you're ready @Koenkk , I'd appreciate a review. :-)

Koenkk:
Looks good, big thanks!

PR #492, opened 2022-01-15
Refactor sendWhen

Koenkk:
See https://github.com/Koenkk/zigbee-herdsman/commit/3ee4af3c13a2a4d186c6d79d451c1c314a4e3d0f#r63645018

Related prs:
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/3688
- https://github.com/Koenkk/zigbee2mqtt/pull/10832

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -101,13 +101,11 @@ class Device extends Entity {
     get defaultSendWhenActive(): boolean {return this._defaultSendWhenActive;}
     set defaultSendWhenActive(defaultSendWhenActive: boolean) {
         this._defaultSendWhenActive = defaultSendWhenActive;
-        for (const endpoint of this._endpoints) {
-            endpoint.defaultSendWhenActive = defaultSendWhenActive;
-        }
     }
+    get useImplicitCheckin(): boolean {return this._useImplicitCheckin;}
 
     public meta: KeyValue;
-    private useImplicitCheckin: boolean;
+    private _useImplicitCheckin: boolean;
 
     // This lookup contains all devices that are queried from the database, this is to ensure that always
     // the same instance is returned.
@@ -159,7 +157,7 @@ class Device extends Entity {
         this.meta = meta;
         this._lastSeen = lastSeen;
         this._defaultSendWhenActive = defaultSendWhenActive;
-        this.useImplicitCheckin = useImplicitCheckin;
+        this._useImplicitCheckin = useImplicitCheckin;
     }
 
     public createEndpoint(ID: number): Endpoint {
@@ -167,8 +165,7 @@ class Device extends Entity {
             throw new Error(`Device '${this.ieeeAddr}' already has an endpoint '${ID}'`);
         }
 
-        const endpoint = Endpoint.create(ID, undefined, undefined, [], [], this.networkAddress, this.ieeeAddr,
-            this.defaultSendWhenActive || false);
+        const endpoint = Endpoint.create(ID, undefined, undefined, [], [], this.networkAddress, this.ieeeAddr);
         this.endpoints.push(endpoint);
         this.save();
         return endpoint;
@@ -192,10 +189,7 @@ class Device extends Entity {
     }
 
     public implicitCheckin(): void {
-        if (!this.useImplicitCheckin) {
-            return;
-        }
-        this.endpoints.forEach(async e => e.sendPendingRequests());
+        this.endpoints.forEach(async e => e.sendPendingRequests(false));
     }
 
     public updateLastSeen(): void {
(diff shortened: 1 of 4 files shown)


Koenkk:
@kennylevinsen  this pr is not complete yet, but I hope you get the idea behind it. Could you take a look?

Some further ideas:
- Replace _useImplicitCheckin and defaultSendWhenActive in device.ts to `defaultSendWhen`
- https://github.com/Koenkk/zigbee-herdsman/pull/492/files#diff-be0e9f494ec1a1e1e2c5d8c33866b7ad6b8e2f080c85133eced5c95a249e41bfR311 still feels a bit strange, even when sendWhen == active or fastpoll the request will be send immediately, do you remember the reason for this?

src/controller/model/device.ts
```diff
@@ -515,11 +509,6 @@ class Device extends Entity {
     }
 
     private async interviewInternal(): Promise<void> {
-        // Clear defaultSendWhenActive mode when running interview
-        for (const endpoint of this._endpoints) {
-            endpoint.defaultSendWhenActive = false;
-        }
```
kennylevinsen, line 521: It's nice to get rid of this, which was an ugly workaround for interview unittests failing under sendWhenActive.

src/controller/model/endpoint.ts
```diff
@@ -261,43 +264,62 @@ class Endpoint extends Entity {
...
-            this.pendingRequests.push({func, resolve, reject});
+            this.pendingRequests.push({func, resolve, reject, queueSendWhen});
         });
     }
 
-    private async sendRequest<Type>(func: () => Promise<Type>, sendWhenActive: boolean): Promise<Type> {
+    private async sendRequest<Type>(func: () => Promise<Type>, 
+        sendWhen: 'active' | 'auto' | 'immediate'): Promise<Type> {
+        const device = this.getDevice();
+        const fastPollingAvailable = !device.useImplicitCheckin;
```
kennylevinsen, line 294: I guess we should rename `useImplicitCheckin` now that it's been repurposed.

src/controller/model/endpoint.ts
```diff
@@ -261,43 +264,62 @@ class Endpoint extends Entity {
...
+    public async sendPendingRequests(fastPolling: boolean): Promise<void> {
+        const handled: PendingRequest[] = [];
+        for (const request of this.pendingRequests) {
+            if (fastPolling || request.queueSendWhen == 'active') {
+                try {
+                    const result = await request.func();
+                    request.resolve(result);
+                } catch (error) {
+                    request.reject(error);
+                }
```
kennylevinsen, line 276: It would have been really neat if sendWhen active/fastpoll messages that failed could retry on next active/fastpoll period, instead of retrying immediately inside the adapter logic... But that's out of scope for this PR.

src/controller/model/endpoint.ts
```diff
@@ -261,43 +264,62 @@ class Endpoint extends Entity {
...
 
+        // TODO
         try {
             return await func();
         } catch(error) {
-            if (!sendWhenActive) {
+            if (queueSendWhen === 'immediate') {
                 throw(error);
             }
         }
```
kennylevinsen, line 318: > [this] still feels a bit strange, even when sendWhen == active or fastpoll the request will be send immediately, do you remember the reason for this?

It is strange, and it's a little annoying to effectively have retry logic both at this level and at adapter level...

The reason for attempting to send immediately was twofold:
1. To minimize responsiveness impact when sendWhenActive suddenly got enabled device-wide, as it previously only affected 4 individual and highly specific messages.
2. To improve sendWhenActive behavior when sending in response to a received message which was otherwise postponed till next time the queue flushed. An example would be updated load estimate from danfoss thermostats, which immediately result in sending the mean load.

The behavior was implemented in https://github.com/Koenkk/zigbee-herdsman/pull/447. Previous implementations of this used a short timeout after the queue emptied during which instantaneous sending was allowed. Maybe going back to something like that would be prettier, and when fastpoll is in use we can control exactly how long such a timeout should be.

kennylevinsen (commented):
This is overall a nice refactor, especially in the cleaner interview handling. I assume you want to address the "device with super long checkinInterval" quirk in a separate PR?

kennylevinsen:
Also note that there are still a few toZigbee converters that used sendWhenActive directly and so need to be updated.

src/controller/model/endpoint.ts
kennylevinsen: Nvm, you already said you want it to be defaultSendWhen. :)

src/controller/model/endpoint.ts
Koenkk: Makes sense, lets keep it.

Koenkk:
Completed the refactor now, zhc and z2m prs can be found in the OP

@kennylevinsen could you do a final check?

kennylevinsen (approved):
LGTM!

Koenkk:
@kirovilya @nurikk FYI this is a breaking change, `sendWhenActive` has been removed and replaced with `sendWhen: 'active'`

kirovilya:
ok. not currently using it

PR #501, opened 2022-02-12
New method "writeResponse" implementation in Endpoint class

systemiii:
Added new method "writeResponse" in Endpoint class.
Property "status" can be specified in payload otherwise it's set to 0.

My first pull request on Github :D

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -414,31 +414,32 @@ class Endpoint extends Entity {
         let payload;
         for (const [nameOrID, value] of Object.entries(attributes)) {
             if (value.hasOwnProperty('status')) {    
-	        if (cluster.hasAttribute(nameOrID)) {
-		    const attribute = cluster.getAttribute(nameOrID);
-		    if (value.status !== 0) {
-            	        payload = payloadOther;
-            	        payload.push({attrId: attribute.ID, status: value.status});
-            	    } else {
-            	        // When status is set to 0 (SUCCESS) attribute 'attrId' is omitted
-            	        payload = payloadSuccess;
-            	        payload.push({status: value.status});
-            	    }  		    
-	        } else if (!isNaN(Number(nameOrID))){
-	            if (value.status !== 0) {
-            	        payload = payloadOther;
-            	        payload.push({attrId: Number(nameOrID), status: value.status});
-            	    } else {
-            	        // When status is set to 0 (SUCCESS) attribute 'attrId' is omitted
-            	        payload = payloadSuccess;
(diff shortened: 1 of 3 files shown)


Koenkk:
Looks good! Could you add some tests? Then this is ready to merge. You can use https://github.com/Koenkk/zigbee-herdsman/blob/85cd71a7a1e400de9ec85af79375e48b66ed2b7f/test/controller.test.ts#L3039 as inspiration.

Tests can be run locally by executing `npm run test-with-coverage`.

systemiii:
Thank you @Koenkk, adding new tests was also useful to better study the 'Write Response' command, in fact I improved the previous code.
I think I have done, if there are no more steps.

src/controller/model/endpoint.ts
```diff
@@ -403,6 +403,68 @@ class Endpoint extends Entity {
...
     }
 
+    public async writeResponse(
+        clusterKey: number | string, transactionSequenceNumber: number, attributes: KeyValue, options?: Options
+    ): Promise<void> {
+        assert(!options || !options.hasOwnProperty('transactionSequenceNumber'), 'Use parameter');
+        const cluster = Zcl.Utils.getCluster(clusterKey);
+        options = this.getOptionsWithDefaults(options, true, Zcl.Direction.SERVER_TO_CLIENT, cluster.manufacturerCode);
+        const payloadOther: {status: number; attrId: number}[] = [];
+        const payloadSuccess: {status: number}[] = [];
```
Koenkk, line 413: What if something is in both payloadOther and payloadSuccess? It seems only the last one set by `payload = payloadOther;` is send.

systemiii: According to the ZCL specification, the `status = 0` and the `status != 0` must not be indicated at the same time in the payload, so this situation should never occur, if the user writes the payload correctly.

Wrong `{0x0000: {status: 0x00}, 0x4292: {status: 0x06}, 0xfff1: {status: 0x01}, 0xfff2: {status: 0x01}}`
Correct `{0x4292: {status: 0x06}, 0xfff1: {status: 0x01}, 0xfff2: {status: 0x01}}`
Correct `{0x0000: {status: 0x00}`

I just verified that 'ZclFrame.create' is able to ignore the 'attrId' attribute when the 'status' is 0, so I think it is better to simplify the method as it was before.

systemiii: Changes done.
Sorry, I made the code too complicated

systemiii:
I have tested the method with Livolo devices and It works

Koenkk:
thanks!

PR #513, opened 2022-03-21
bounds check in parseEnqueueSendDataResponse

kizmc:
Relates to https://github.com/Koenkk/zigbee2mqtt/issues/10542

Although this does not address the actual data in the unexpectedly short packet, I don't think that we should allow a rogue/undefined packet to crash zigbee2mqtt.


Koenkk:
@ChrisHae could you review this?

PR #517, opened 2022-04-03
Update cluster.ts

Phil-ibert:
Hello,

I have added the newly found OWON non-standard attributes to the seMetering definition


Koenkk:
thanks!

PR #532, opened 2022-04-29
Update Inovelli VZM35-SN activePowerReports type

nathanfiscus:
This device is not yet in production, but the most recent pre-production firmware changed this parameter to a uint8.


Koenkk:
Thanks!

PR #534, opened 2022-05-04
Add Options attribute to Level Control cluster

pyos:
This is a ZCLv7 mandatory (for lights) attribute; in particular, setting bit 0 (ExecuteIfOff) is useful because it allows sending 'move to level' commands while the light is off to change the brightness at which it will turn on with the next 'on' command (assuming OnLevel is 255)

Not sure how many manufacturers really implement this attribute, but at least IKEA does


Koenkk:
Thanks!

PR #546, opened 2022-05-30
Add new Tuya definitions

vladyspavlov:
I sniffed traffic of TS0004 and found these new attributes, which are not defined.


Koenkk:
Thanks!

Issue #555, opened 2022-07-04
Unhandled promise rejection when stopping the coordinator


ercksen:
Hey, I encountered a bug today where my Node process exited due to an unhandled promise rejection and finally found out that stopping the coordinator was the cause.

Following code can reproduce the issue:

```
const {Controller} = require("zigbee-herdsman");

const coordinator = new Controller({
    databasePath: 'test.db',
    serialPort: {path: '/dev/ttyACM0'}
});

let date = new Date();

coordinator
    .start()
    .then(() => {
        coordinator.stop()
            .then(() => console.log("Stopped", new Date() - date))
            .catch(err => console.log("Catch error (2):", new Date() - date, err));
        console.log("Controller created", new Date() - date);
    })
    .catch(err => console.log("Catch error (1):", new Date() - date, err))
```

I added dates for debugging purposes but when the app crashes no error handler is invoked. I used my stopwatch and it takes 10 seconds to crash. Output:

```
Controller created 211
Stopped 319
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

[UnhandledPromiseRejection: This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). The promise rejected with the reason "TIMEOUT".] {
  code: 'ERR_UNHANDLED_REJECTION'
}
```

I tried to get the stack using the unhandled rejection event but it was undefined, I do not know where to search. It would be nice to restart the controller without terminating Node itself. I am using CC253X with DECONZ Adapter.

Maybe this information is helpful to debug the problem. If you have any further questions, let me know. I am using v0.14.40 and the test.db file had no previous devices in it.

ercksen:
Here is the log output. I marked the point at which the `stop()` function of the controller is invoked. The lines before are from regular communication I think but I have attached some of the last lines before the stop was initiated. I forgot to mention that I am using Docker with a mounted device, if that makes any difference.

```
2022-07-06T17:39:09.679Z zigbee-herdsman:deconz:adapter checking coordinator simple descriptor
2022-07-06T17:39:09.681Z zigbee-herdsman:deconz:driver DATA_REQUEST - destAddr: 0x0 EP:0 SeqNr. 3 request id: 1
2022-07-06T17:39:09.682Z zigbee-herdsman:deconz:frameParser DATA_REQUEST RESPONSE - request id: 1 status: 0
[...]

PR #561, opened 2022-07-12
Fix unhandled promise rejection (#555)

ercksen:
Deconz driver: Clear intervals and shared arrays after close; catch all unhandled promises and forward errors to debugger


Koenkk:
Thanks!

PR #570, opened 2022-08-27
EZSP adapter: Change parameters and generate new TC link key

kirovilya:
Changed some parameters to start the adapter.
Set the mode for changing the TC link key on request - necessary for some devices that do not accept the standard key due to security (Perenio plug).
Fixed a crash when choosing pairing on a separate router (does not work, but does not crash now).


Koenkk:
thanks!

PR #577, opened 2022-09-08
EZSP adapter: correct startup status and disable set some configuration values

kirovilya:
Correct startup status when form network
Disable set some configuration values


Koenkk:
Thanks!

PR #580, opened 2022-09-18
EZSP: Trying to fix pairing process and prepare for Install code usage

kirovilya:
https://github.com/Koenkk/zigbee-herdsman/issues/319


MattWestb:
Great that you changing to using the firmware default = the maximum the firmware can handling so can using the resources better in the SOC.

If having problems with end devices leaving and rejoining and battery draining its possible changing `CONFIG_MAX_END_DEVICE_CHILDREN` to `0` after have adding at least one router and end devices cant using the coordinator and being forced using routers > more stable network function with end devices.

kirovilya:
> If having problems with end devices leaving and rejoining and battery draining its possible changing `CONFIG_MAX_END_DEVICE_CHILDREN` to `0` after have adding at least one router and end devices cant using the coordinator and being forced using routers > more stable network function with end devices.

Thanks for the info. It is currently not possible to pass additional settings for adapters. If there are problems, I will think about how to transfer such settings.
I have a problem in one more place: one of my devices cannot connect to the network, because the network uses the default TC key. I was able to resolve this issue by setting the policy TC_KEY_REQUEST_POLICY = GENERATE_NEW_TC_LINK_KEY instead of ALLOW_TC_KEY_REQUESTS. But then the pairing process becomes more difficult for other devices. Therefore, I have returned the old value of this policy for now. I do not know how to change the key only for this device.

MattWestb:
Old Zigbee HA 1.x device cant updating the TC-Link key and ZB3 is one must **if** running new stack in the device (Zigbee PRO 19 and later i think its the version its enforced if both side is having high enough Zigbee stack) and all is done by policy in the NCP and is global and can not being don by device.
I think if forcing updating TC-link key some devices is getting problem then they is not doing the handshake or miss reading it. The paring (ZB3) device is asking for node description of the coordinator and shall using the version of the Zigbee stack and if its high enough it shall requesting key update and if its lower its up to the device staying with default TC-Link key or leaving the network (policy in the device firmware).

PR #586, opened 2022-09-28
Add Manufacturer clusters for Develco MOSZB140

sjorge:
As part of https://github.com/Koenkk/zigbee2mqtt/discussions/14178 I got my hands on a MOSZB140 so I want to improve support for it, in the technical spec there are a few manufacturer specific attributes that will be useful.

- genBasic.develcoLedControl: to control the LED behavior
- ssIasZone.develcoAlarmOffDelay: to allow changing of the timeout for the occupancy alarms send over ssIasZone
- msOccupancySensing.develcoTargetLevel/develcoArmThresholdMinTemperature/develcoArmThresholdMaxTemperature: to allow logic configuration for the 3 available endpoint that host msOccupancySensing, these are more configurable and more sensitive than the ssIasZone provided one.


Koenkk:
Thanks!

src/zcl/definition/cluster.ts
```diff
@@ -50,6 +50,7 @@ const Cluster: {
             deviceEnabled: {ID: 18, type: DataType.boolean},
             alarmMask: {ID: 19, type: DataType.bitmap8},
             disableLocalConfig: {ID: 20, type: DataType.bitmap8},
+            develcoLedControl: {ID: 0x8100, type: DataType.bitmap8, manufacturerCode: ManufacturerCode.DEVELCO},
```
sjorge, line 53: When doing a read of genBasic.develcoLedControl this fails with UNSUPPORTED_ATTRIBUTE **unless** I specify the manufacturerCode in the `ep.read()` call. I this the correct behavior? I was expecting not having to specify the manufacturerCode as it is listed in the definition here. (And there are no overlapping attributes with ID 0x8100).

sjorge: Ok so for `ep.read()` and `ep.write()` I need to specify it, it gets picked up proper in the  readReposne/attributeReport. Seems to be normal behavior.

Koenkk: It is indeed picked from the options, not sure what is the most logical here given that you can read non and manufaturer specific attributes at the same time while there can be only 1 manuf code.

Code: https://github.com/Koenkk/zigbee-herdsman/blob/6561b752fc17dd6600bb2a8e4e1241b5547043c4/src/controller/model/endpoint.ts#L463

sjorge: I guess if it's a list of attributes to read/write, we could split them in groups based on manufacturerCode if non is provided via options 🤔 and do multiple read/writes.

So `['deviceEnabled', 'alarmMask', 'develcoLedControl']` would be split into ['deviceEnabled', 'alarmMask'] and `[ 'develcoLedControl']` if no manufacturerCode is specified in the options.

So it will 'just work tm', that will also make the frontend able to read/write to these because you cannot specify options for that at the moment.

Koenkk: Yes but that will make it more complex. Maybe we can do something as simple as: all attributes should either use none are the same manufacturer code (if not we just throw an error).

sjorge: That would be just as complicated right?

e.g if options is undefined, we would need to check the cluster definition in either case to see if all manufcaturerCode's are the same (be it none or a specific one). So if we're here anyway  why not split them into multiple lists to read ?

Probably not really worth doing anything, I don't thing many people will run into this, just those that are working on adding/improving device support. Although the behavior is certainly a bit unexpected.

Koenkk: Yes lets leave it as is.

PR #588, opened 2022-09-30
develcoAlarmOffDelay: correct ID from 0x8100 to 0x8001

sjorge:
image

My brain has a hard time noticing 0x8100 and 0x8001 are not the same... spend 30min scratching my head on this one.

On a happier note, led_control is working well :)


Koenkk:
Haha, well found!

PR #597, opened 2022-10-16
Cleanup Develco Attributes

sjorge:
Develco/Frient never replied, so I am removing the attribute that were returning UNSUPPORTED_ATTRIBUTE.
I also added the develcoPrimarySwVersion and develcoPrimaryHwVersion attributes properly, after do not forget to merge the referenced zhc PR also.


Koenkk:
Thanks!

PR #598, opened 2022-10-17
EZSP adapter: backup and experimental EZSPv7,v6,v5 support

kirovilya:
https://github.com/Koenkk/zigbee-herdsman/issues/319

Backup EZSP adapter config, but without restore yet... and without devices...
Support for earlier versions of EZSP (earlier than 8), but without source routing yet...


Koenkk:
Great work! 🎉

PR #604, opened 2022-11-05
EZSP: watchdog returned, added CRC checker

kirovilya:
With the release of version 1.28.1, problems were found on some gateways connected via TCP. Moreover, when you restart z2m, everything is restored.
Presumably, errors occur during the messaging process (incorrect CRC, frames with errors) and the handler did not react to such messages in any way, considering them to be normal, hoping that the next messages would be good.
In such cases, due to the disconnected watchdog, the connection was not reset and reconnected (similar to a restart).

So added frame checksum check, reset on error, returned watchdog.


Koenkk:
🚀

PR #606, opened 2022-11-07
Add doubleTapClearNotifications to Inovelli Cluster

nathanfiscus:
Added in fw 2.05 for the VZM31-SN

Added to converter here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/4910


Koenkk:
Thanks!

PR #615, opened 2022-11-22
Add Sinope attributes

faucherc:
Adding some Sinope attributes in the existing clusters.  Most of these attributes is used by the Sinope TH1400zb that I'm currently enhancing the converter.


Koenkk:
Thanks!

PR #616, opened 2022-11-22
Add DawonDNS specific attributes

nukusinji:
For DawonDNS Plug device , B530, B540 
DawonDNS has not specific manufacture ID.

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -3432,6 +3432,7 @@ const Cluster: {
             owonAccumulativeEnergyThreshold: {ID: 0x5006, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
             owonReportMode: {ID: 0x5007, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
             owonPercentChangeInPower: {ID: 0x5007, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
+            dawonspecific: {ID: 0x0099, type: DataType.uint8},
         },
         commands: {
             getProfile: {


src/zcl/definition/cluster.ts
```diff
@@ -3432,6 +3432,7 @@ const Cluster: {
             owonAccumulativeEnergyThreshold: {ID: 0x5006, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
             owonReportMode: {ID: 0x5007, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
             owonPercentChangeInPower: {ID: 0x5007, type: DataType.uint8,manufacturerCode: ManufacturerCode.OWON},
+            dawonspecific: {ID: 0x0099, type: DataType.uint8},
```
Koenkk, line 3435: Can we add a `manufacturerCode` here?

nukusinji: > 

They use manufacturerCode "manufId":4098 and manufId":4602".  So i can not added it.
 I think they has not manufacturer Code because they not 'zigbee alliance'.

**Can we use manufacturerName??**
**They use manufName":"DAWON_DNS"**

Koenkk: Then it may be better not to add it, this will cause problems when other manufacturers also have a attribute with the same ID.

nukusinji: > 

I wanna add a tozigbee function. 
const payload = {0x0099: {value: 0x00, type: 0x20}}; // INT8U
                await entity.write('seMetering', payload);
but device is support attribute 0x0099, but it occur error 'UNSUPPORTED_ATTRIBUTE'.

Publish 'set' 'energy_reset' to 'hair_dryer_plug' failed: 'Error: Write 0x00158d0004f669d1/1 seMetering({"153":{"value":0,"type":32}}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":true,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (Status 'UNSUPPORTED_ATTRIBUTE')'

How to I use this attribute without add in cluster.ts .

Koenkk: If it doesn't work this way, I don't expect it to work when added to the cluster.ts file, can you double check this?

nukusinji: > If it doesn't work this way, I don't expect it to work when added to the cluster.ts file, can you double check this?

Sorry, I mean if can not add this PR, How to I use this attribute.
 because **it occur error 'UNSUPPORTED_ATTRIBUTE' without this PR.**

Koenkk: Can you provide the herdsman debug log when setting it with:

```js
const payload = {0x0099: {value: 0x00, type: 0x20}}; // INT8U
                await entity.write('seMetering', payload);
```

and when it is added to `cluster.ts`?

nukusinji: i need more time for modify cluster.ts in synology docker.

PR #622, opened 2022-12-05
philips: define FC03/2

zegl:
As a part of https://github.com/Koenkk/zigbee-herdsman-converters/pull/5049

Define that 0xFC03 / 0x0002 is used as the state for Philips Hue Gradients.


Koenkk:
Thanks!

PR #623, opened 2022-12-06
Added new manufacturer specific commands for Bosch

danieledwardgeorgehitchcock:
Resolves https://github.com/Koenkk/zigbee2mqtt/issues/13520#issuecomment-1337186900

I am in unknown territory here so, this will definitely need a review!

This should add the functionality to replicate this command:

Bosch Tamper Test


src/zcl/definition/cluster.ts
```diff
@@ -5030,6 +5030,21 @@ const Cluster: {
         commandsResponse: {
         },
     },
+    manuSpecificBosch: {
+        ID: 0x05,
```
Koenkk, line 5034: This should be `0x500`

Koenkk:
Can you provide a screenshot of what is send with this code?

danieledwardgeorgehitchcock:
> Can you provide a screenshot of what is send with this code?

That's pretty much it - there's a pcap file in the linked post

Koenkk:
Thats with the original gateway right? I meant what is send when using this command in z2m.

danieledwardgeorgehitchcock:
Ah sorry! I'll get right on it!

danieledwardgeorgehitchcock:
Not sure why I am getting a build error but, will look into that later.

I have added the command to the ssIasZone cluster as suggested above and can now invoke it via the frontend:

Bosch Test Tamper

[Here is the pcap file](https://drive.google.com/file/d/1tXRS0BHmH7MNk3wwBBx_7ynvesoXmDPq/view?usp=sharing) as requested - a couple of points I have observed:

- Manufacturer code is missing (I expected this as it isn't included anywhere in the command but, how do I do that?)
- The frame control field seems to be different (Z2M `0x001`, Bosch controller `0x005`)
- The data payload is missing - Not sure why this would be the case..?

Even given the above, I feel that this is progress in the right direction - Z2M and the device don't seem to be grumbling about the command, but at the same time, it doesn't functionally change anything on the device...

Koenkk:
> Manufacturer code is missing (I expected this as it isn't included anywhere in the command but, how do I do that?)

This is currently not possible via the frontend. Also regarding the next point, when you set the manufacturerCode `Manufacturer Specific: True` will be set automatically. The code to send this command is (double check `manufacturerCode`) 

```js
await entity.command('ssIasZone, 'boschTestTamper', {data: 2}),{manufacturerCode: 0x1133});
```

> The data payload is missing - Not sure why this would be the case..?


Can you provide the network key of your sniff? Just a screenshot of the send command is also enough (and easier). But first check with the above command.

Koenkk:
Thanks!

> I can see that there is a default response being returned too - I guess I'd populate the commandResponse section for that?

No, a defaultResponse is a standard ZCL command which is already being handled by zigbee-herdsman out-of-the-box.

PR #626, opened 2022-12-08
Add support for indicator for Clipsal Wiser device

PramodAryalGithub:
Indicator states. 0: On when load is on; 1: Always on; 2: on when load is off; 3: always off


Koenkk:
Thanks!

PR #634, opened 2022-12-24
Add attributes to manuSpecificSinope

MattL0:
I have read/poll every attribute from 0 t0 1200 of manuSpecificSinope cluster on a TH1123ZB-G2, with a small  bash script.

A nice (I will start an issue/conversation on this on the zigbee-herdman-converters git) part is that some work can be started on the eco mode. When writting to any of ecoMode or ecoMode1-2, this set a sort of offset on triac/valve ( I mean you need to set an setpoint higher/lower than usual to get heat. At the same time playing with those attributes, I saw that you can send more ''juice'' to the heater ( in watts) even when the thermostat is set to the maximum setpoint). 
- ecoMode can be set from -100 to 100
- ecoMode1 can be set from 0 to 99
- ecoMode2 can be set from 0 to 100
- Theses create variance in currentLoad atttribute

Three issues remain ( this devices uses zcl version 8) ( Not sure how many issues i need to create here for theses)

- There is an issue where the manufacturerCode needs to be set again in converters, for Groups commands that have at least 1 Sinope G2  (clusters manuSpecificSinope and hvacThermostat ( attributes: SinopeOccupancy, SinopeMainCycleOutput, SinopeBacklight)
- Theses 3 lasts attributes need manufacturerCode to be set when reading ( repports that the attribute is not readable if not set), and same for writing.  
- Sinope TH1123ZB-G2 tries to read dst time every 5 sec or less, and this is spammimg the logs. I think nothing is sent back.  I think there needs to be a ways to update that time ''handshake'' for this device. 

- ( for references only) This model seems to be low on memory, or the firmware use to much. we need to find some way to unload that, so we can add more attributes reporting (this g2 stats seems to be able to repport anything)

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4075,7 +4075,7 @@ const Cluster: {
             // attribute ID :1's readable
             keypadLockout: {ID: 2, type: DataType.enum8},
             // 'firmware number': {ID: 3, type: DataType.unknown},
-            firmwareVersion: {ID: 4, type: DataType.charStr },
+            firmwareVersion: {ID: 4, type: DataType.charStr},
             outdoorTempToDisplay: {ID: 16, type: DataType.int16},
             outdoorTempToDisplayTimeout: {ID: 17, type: DataType.uint16},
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #639, opened 2023-01-02
fix wrong datatype Sinope

MattL0:
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5195#issuecomment-1369207694


Koenkk:
Thanks!

PR #640, opened 2023-01-04
Add support for incomingNetworkStatusHandler

fuzzy01:
Fix for https://github.com/Koenkk/zigbee2mqtt/issues/15724
Based on latest UG100 reference guide 0x00C4 has been redefined.

### DIFF
--- a/src/adapter/ezsp/driver/commands.ts
+++ b/src/adapter/ezsp/driver/commands.ts
@@ -2224,12 +2224,12 @@ export const FRAMES: {[key: string]: EZSPFrameDesc} = {
             // gpdCommandPayload: LVBytes
         },
     },
-    changeSourceRouteHandler: {
+    incomingNetworkStatusHandler: {
         ID: 0x00C4,
         request: null,
         response: {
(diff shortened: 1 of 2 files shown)


Koenkk:
@kirovilya can you review this?

kirovilya:
image
image

I see that the changes appeared in the new version of the stack 7.2.0
so it's strange to see that these changes affect the bug https://github.com/Koenkk/zigbee2mqtt/issues/15724 where stack 6.10.3 was used

And another question will be how to combine work with old stacks, where there was another frame under the code 0xC4? It is necessary to make some bypass to support both stacks.

MattWestb:
In GSDK 3.2 for EZSP 6.10.X is not allocated `0x00C4 --unassigned-- ` so very likely its taken away between EZSP 6.0.X and 6.9.x and very likely in 6.7.X that was getting the V8 framing.
Perhaps it was implanted but not documented. Interesting if 6.7.X (= stable version of EZSP) is using the new or old command for the frame or not using it at all.
`

PR #651, opened 2023-01-30
added more debugging info for LQI/routingTable

TomTonic:
Sorry, I don‘t know TypeScript and I don‘t have a development environment.
Please regard this as a feature request that might be rather close to an actual implementation 😇

### DIFF
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -564,7 +564,7 @@ class ZStackAdapter extends Adapter {
                 );
                 const result = await response.start().promise;
                 if (result.payload.status !== ZnpCommandStatus.SUCCESS) {
-                    throw new Error(`LQI for '${networkAddress}' failed`);
+                    throw new Error(`LQI for '${networkAddress}' failed with status code '${result.payload.status}' ('${ZnpCommandStatus[result.payload.status]}')`);
                 }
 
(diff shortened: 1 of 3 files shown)


TomTonic:
However, I‘m not sure if this will always yield in the same „with status code '1' ('FAILURE')“ in real wold scenarios. In this case, the „feature“ would be pretty useless.

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -564,7 +564,8 @@ class ZStackAdapter extends Adapter {
                 );
                 const result = await response.start().promise;
                 if (result.payload.status !== ZnpCommandStatus.SUCCESS) {
-                    throw new Error(`LQI for '${networkAddress}' failed`);
+                    throw new Error(`LQI for '${networkAddress}' failed with status code '${
```
Koenkk, line 567: I think this is a good improvement, but could you change the format to:

```js
LQI for '${networkAddress}' failed with error: '${ZnpCommandStatus[result.payload.status]}' (${result.payload.status})
```

TomTonic: ok, done. I tried to add the same functionality to the ezsp adapter, yet the error code lookup did not work - so I skipped it for the ezsp adapter.

Koenkk:
Thanks!

PR #653, opened 2023-01-31
Add Yale Module Clusters

danieledwardgeorgehitchcock:
enhances work done in Koenkk/zigbee2mqtt#4778

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -5166,7 +5166,217 @@ const Cluster: {
             },
         },
         commandsResponse: {},
-    },    
+    },
+    manuSpecificAssaDoorLock: {
+        ID: 0xEACC,
+        attributes: {
+            autoLockTime: {ID: 0x0012, type: DataType.uint8},
+            wrongCodeAttempts: {ID: 0x0013, type: DataType.uint8},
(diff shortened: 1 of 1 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -5166,7 +5166,262 @@ const Cluster: {
             },
         },
         commandsResponse: {},
-    },    
+    },
+    manuSpecificAssaDoorLock: {
+        ID: 0xFC00,
```
Koenkk, line 5171: Could we add a manufacturer code here like; https://github.com/Koenkk/zigbee-herdsman/blob/064662f37da1c0538e52e09f80bc3f724eb693a9/src/zcl/definition/cluster.ts#L4141

danieledwardgeorgehitchcock: I looked for one in the decrypted driver and sniffs but couldn't find one. Will dig a bit deeper

danieledwardgeorgehitchcock: Just double checked the sniff, and there isn't any `manufacturerCode` that is sent along with the packets..?

image

Koenkk: Thanks for checking!

Koenkk:
Thanks!

PR #654, opened 2023-02-01
Add Big-endian uint type and new Tuya definitions

mayongze:
resolve  Koenkk/zigbee2mqtt#16404

### DIFF
--- a/src/zcl/definition/dataType.ts
+++ b/src/zcl/definition/dataType.ts
@@ -32,6 +32,7 @@ enum DataType {
     int32 = 43,
     enum8 = 48,
     enum16 = 49,
+    beUint24 = 50,
     singlePrec = 57,
     doublePrec = 58,
     octetStr = 65,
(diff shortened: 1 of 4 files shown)


src/zcl/definition/dataType.ts
```diff
@@ -32,6 +32,7 @@ enum DataType {
     int32 = 43,
     enum8 = 48,
     enum16 = 49,
+    beUint24 = 50,
```
Koenkk, line 35: Could you add this to https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zcl/definition/buffaloZclDataType.ts instead (since this is not an official Zigbee type)

mayongze: The changes have been made and can be reviewed now. Thank you for bringing this to my attention

Koenkk:
Thanks!

PR #664, opened 2023-02-12
Deconz Adapter: add timeout to WaitForDataRequest and react to it accordlingly (replaces hardcoded 60 second timeout)

ercksen:


Koenkk:
Thanks!

PR #669, opened 2023-02-20
EZSP: additional cleaning of uart packets and elimination of the address request during the ZDO request

kirovilya:
Skipping XOFF and XON byte in uart package.
Adding address pair to mapping on bind/unbind/remove request.


Koenkk:
Thanks!

PR #670, opened 2023-02-21
Throw an error object for APS TIMEOUT error.

yehzhang:
Fixes Koenkk/zigbee2mqtt#16756. 

Alternatively, the catcher of the error can handle both error objects and strings. It is just easier to always throw an error object.


Koenkk:
Thanks!

PR #671, opened 2023-02-24
Rename PM25 cluster, add support for Ikea type

### LINKED ISSUE #667: PM2.5 cluster is not actually heiman specific
Hi, this issue is related to another thread (https://github.com/Koenkk/zigbee2mqtt/issues/16717). We're trying to get the new IKEA air quality sensor up and running.

Currently, zigbee cluster 0x42A is specified as `heimanSpecificPM25Measurement`. However it isn't manufacturer specific at all - it is a normal cluster as specified in the ZCL spec (chapter 4.13). That's a problem now though, since there is a "manufacturerCode" given for this cluster that prevents messages with the IKEA manufacturer code coming through.

https://github.com/Koenkk/zigbee-herdsman/blob/2d1c453faf889afe57533778abf30fb9f074a1e9/src/zcl/definition/cluster.ts#L4589

My naive approach would be:
* simply renaming this cluster to "pm25Measurement" and renaming all references in zigbee-herdsman-convertes as well
* removing the manufacturerCode spec, since it isn't manufacturer specific at all

I'm not too familiar with zigbee-herdsman's architecture though. Are those two steps sufficient (would be happy to provide a PR then) or will we break something by renaming the cluster and removing the manufacturer code? 

Cheers!

peterkappelt:
Fixes #667 

The PM25 cluster is not specific to heiman, but a general purpose cluster defined by the ZCL spec.

Also, I've added the "measuredValueIkea" property that is of non-standard type. This is the required base for IKEA Vindstyrka support (https://github.com/Koenkk/zigbee2mqtt/issues/16717)

Should be merged at the same time as https://github.com/Koenkk/zigbee-herdsman-converters/pull/5520, in order to match the name change there. Let me know if there's anything left!

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -2630,6 +2630,9 @@ const Cluster: {
     pm25Measurement: {
         ID: 0x042a,
         attributes: {
+            //IKEA Vindstyrka: measured value is reported as float
+            measuredValueIkea: {ID: 0x0000, type: DataType.singlePrec, manufacturerCode: ManufacturerCode.IKEA_OF_SWEDEN},
+            //default cluster spec: values reported as uint16
             measuredValue: {ID: 0x0000, type: DataType.uint16},
             measuredMinValue: {ID: 0x0001, type: DataType.uint16},
             measuredMaxValue: {ID: 0x0002, type: DataType.uint16},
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #674, opened 2023-03-04
Add Ikea VOC index cluster

0ip:
Required by
https://github.com/Koenkk/zigbee-herdsman-converters/pull/5546

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4034,9 +4034,9 @@ const Cluster: {
         ID: 0xfc7e,
         manufacturerCode: ManufacturerCode.IKEA_OF_SWEDEN,
         attributes: {
-            measuredValue: {ID: 0x0000, type: DataType.uint16},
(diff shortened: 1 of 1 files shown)


sjorge:
Does this IKEA one also output a resolution value? The custom cluster used by Develco does and I think they are using the same sensirion sensors, you can check if it has an additional uint16 as attribute 3.

For reference:
https://github.com/Koenkk/zigbee-herdsman/blob/1644537c0366db7875c8b9176093e0733341dbfa/src/zcl/definition/cluster.ts#L4766-L4777

0ip:
@sjorge I have tested it and the attribute does not seem to be the present

sjorge:
> @sjorge I have tested it and the attribute does not seem to be the present

Give your reply in the zhc PR, if this is a relative index like value it might makes sense there is none as it probably already accounts for some inaccuracies.

Koenkk:
Thanks!

PR #676, opened 2023-03-05
Improve request queue behavior for sleepy end devices

slugzero:
The current queue logic for sleepy end devices (sendWhen='active' and 'fastpoll') leads to inconsistent behavior of the controller in case of transmission failures: Before a failure occurs, messages are sent immediately. As soon as a failure occurs (for whatever reason, we do not check for the cause), the controller queues all messages and only sends them upon explicit or implicit checkin. Furthermore, messages in the queue are not aged out. So if the error is unrecoverable, the only way to empty the message queue and return to the original behavior is to restart the controller. 
This makes it really hard to debug sleepy end devices, because the controller behaves differently seemingly out of the blue (i.e., after any failed transmission). Also, as soon as a failed request is queued, the device seems unresponsive or only responsive with significant delay. 

This PR fixes the following:
* Consistent behavior of request queueing: On each sendRequest, attempt to send the whole queue, no matter how large the queue is. New requests are always queued at the back to keep the ordering, and if everything goes well, are sent immediately.
* Aging of requests: remove requests if they are in the queue longer than the checkinInterval of the end device. 
*  Clean up entire queue after a "fastpoll" transmission of the entire queue. I put this into a separate commit because I have no device to test this.  

Note that the new send behavior still may lead to requests being held up in the queue for some time because a unrecoverable error is "clogging" the queue. But due to message aging, this will be only temporary (at most for the checkinInterval or until the next checkin from the end device). A solution could be to only queue those requests that failed due to the end device being asleep (i.e. timeout?). 

This will help to fix some of the issues with unresponsive end devices, e.g. https://github.com/Koenkk/zigbee2mqtt/issues/13478, or at least make the issues less persistent. However I do not believe that this will fully fix *all* issues, because there seems to be an additional  firmware-related issue.

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -705,16 +705,21 @@ class Device extends Entity {
 
         // Bind poll control
         try {
+            let retentionTimeout : number;
             for (const endpoint of this.endpoints.filter((e): boolean => e.supportsInputCluster('genPollCtrl'))) {
                 debug.log(`Interview - Poll control - binding '${this.ieeeAddr}' endpoint '${endpoint.ID}'`);
                 await endpoint.bind('genPollCtrl', coordinator.endpoints[0]);
                 const pollPeriod = await endpoint.read('genPollCtrl', ['checkinInterval']);
+                retentionTimeout = pollPeriod.checkinInterval;
                 if (pollPeriod.checkinInterval <= 2400) {// 10 minutes
                     this.defaultSendRequestWhen = 'fastpoll';
                 } else {
                     this.defaultSendRequestWhen = 'active';
                 }
             }
+            this.endpoints.forEach((e) => {
+                e.retentionTimeout = retentionTimeout;
+            });
         } catch (error) {
             /* istanbul ignore next */
             debug.log(`Interview - failed to bind genPollCtrl (${error})`);
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -55,6 +55,7 @@ class Device extends Entity {
     private _deleted: boolean;
     private _defaultSendRequestWhen?: SendRequestWhen;
     private _lastDefaultResponseSequenceNumber: number;
+    private _pollCheckingInterval: number;
 
     // Getters/setters
     get ieeeAddr(): string {return this._ieeeAddr;}
@@ -104,6 +105,8 @@ class Device extends Entity {
     set defaultSendRequestWhen(defaultSendRequestWhen: SendRequestWhen) {
         this._defaultSendRequestWhen = defaultSendRequestWhen;
     }
+    get pollCheckingInterval(): number {return this._pollCheckingInterval;}
+    set pollCheckingInterval(pollCheckingInterval: number) {this._pollCheckingInterval = pollCheckingInterval;}
 
     public meta: KeyValue;
 
@@ -132,7 +135,7 @@ class Device extends Entity {
         manufacturerID: number, endpoints: Endpoint[], manufacturerName: string,
         powerSource: string, modelID: string, applicationVersion: number, stackVersion: number, zclVersion: number,
         hardwareVersion: number, dateCode: string, softwareBuildID: string, interviewCompleted: boolean, meta: KeyValue,
-        lastSeen: number, defaultSendRequestWhen: SendRequestWhen,
+        lastSeen: number, defaultSendRequestWhen: SendRequestWhen, pollCheckingInterval: number
     ) {
         super();
         this.ID = ID;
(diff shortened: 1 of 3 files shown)


Koenkk:
Could you take a look at the test coverage? To see the missing paths execute `npm run test-with-coverage` and check `coverage/lcov-report/index.html`.

slugzero:
Yep, will do. But it might take a little while, I am not really familiar with jest yet.

src/controller/model/device.ts
```diff
@@ -705,16 +705,21 @@ class Device extends Entity {
...
-                const pollPeriod = await endpoint.read('genPollCtrl', ['checkinInterval']);
+                const pollPeriod = await endpoint.read('genPollCtrl', ['checkinInterval'], {sendWhen: 'immediate'});
+                retentionTimeout = pollPeriod.checkinInterval;
                 if (pollPeriod.checkinInterval <= 2400) {// 10 minutes
                     this.defaultSendRequestWhen = 'fastpoll';
                 } else {
                     this.defaultSendRequestWhen = 'active';
                 }
             }
+            this.endpoints.forEach((e) => {
```
Koenkk, line 720: - Every device has only 1 checkinInterval, therefore there is no need to store a value per endpoint. I suggest to put this property on the device instead of endpoint. (same as `defaultSendRequestWhen`). You can provide a default value for this property `fromDatabaseEntry` (also similar to how `defaultSendRequestWhen` is hanlded). This reduces the need to check for existence of this property in the endpoint.
- I suggest to rename this to `pollCheckingInterval`. This value may be useful for other use cases and then `retentionTimeout` might be a confusing name.

slugzero: Thanks for the suggestions. I renamed and removed the variable as you suggested. I did not set the default value in `defaultSendRequestWhen`, however. I thought about this before and I decided not to do so, because that value would also be written back to the database later, where it becomes indistinguishable from actual device data. Since this is just a somewhat arbitrary value, I think this should not go to the database. What do you think?

src/controller/model/endpoint.ts
```diff
@@ -264,48 +267,73 @@ class Endpoint extends Entity {
...
 
     private async queueRequest<Type>(func: () => Promise<Type>, sendWhen: 'active' | 'fastpoll'): Promise<Type> {
         debug.info(`Sending to ${this.deviceIeeeAddress}/${this.ID} when active`);
         return new Promise((resolve, reject): void =>  {
-            this.pendingRequests.push({func, resolve, reject, sendWhen});
+            // Remove request from queue after a timeout. We use the checkinInterval from genPollCtrl if available.
+            // If not, the default is 24h, which practically replicates old behavior where messages never expired
+            // From a standard perspective, as little as 7.68 seconds would be sufficient if the device didn't tell
+            /// us otherwise.
+            const expires = (this.retentionTimeout || 86400000) + Date.now();
```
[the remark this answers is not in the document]
slugzero: changed it, thanks

Koenkk:
Many thanks!

sjorge:
@Koenkk as per our conversation wrt issues with configuring end devices after join and reconfiguring them later on on discord, here is the herdsman debug log captured on my dev setup.

In this case I tried to pair random TRV from my random devices box, it starts out OK but (re)configure stopped after the device send a few message. To not make it sleep I periodically pushed the button to keep it awake but that does not help.

Every time I wake the device it just keeps increasing the pending messages:
```
2023-03-25T13:40:05.214Z zigbee-herdsman:controller:endpoint sendRequest: remaining # of pending requests after sending: 22 active: 0
```

[zhdebug.log](https://github.com/Koenkk/zigbee-herdsman/files/11069030/zhdebug.log)

Koenkk:
@slugzero could you take a look at this?

Koenkk:
@slugzero could you take a look at this? Seems to be some regression.

Koenkk:
Temporarily revert the PR (https://github.com/Koenkk/zigbee-herdsman/pull/687)

slugzero:
@Koenkk I had a look at this and opened a new issue with the findings from my analysis. See https://github.com/Koenkk/zigbee2mqtt/issues/17177

tl;dr: I believe the same behavior was already there before, but it was not visible directly after startup. A way to reproduce this on the latest stable build is described in the ticket. @sjorge would be great if you could confirm that the same behavior can also be provoked on the stable build by sending an invalid request to the device.

sjorge:
I've not been able to trigger this on stable at all, but not sure there are broken/invalid zigbee frames being sent. 

Even after 2 days of uptime i can still successfully leave/join/configure a device.

sjorge:
I read a few attribute via the expose data, then I tried to read genOnOff.onOff which the device does not support, this returns UNSUPPORT_ATTRIBUTE as it should.

I then hit reconfigure.... this also works fine afterwards

[zhdebug-master.log](https://github.com/Koenkk/zigbee-herdsman/files/11114387/zhdebug-master.log)

PR #687, opened 2023-03-28
Revert "Improve request queue behavior for sleepy end devices"

Koenkk:
@slugzero I will temporarily revert your PR because the next z2m release is soon (1 April) and  it has introduced some issues:

- https://github.com/Koenkk/zigbee-herdsman/pull/676#issuecomment-1483828764
- https://github.com/Koenkk/zigbee2mqtt/issues/17160


slugzero:
Alright, I actually wanted to suggest this, too. I think the queue needs some more redesign and I'd like to discuss first what it should look like.

PR #697, opened 2023-04-15
EZSP: Attempt to handle correctly and reconnect on errors.

kirovilya:
Separation of command processing 0x00C4 before version 9 and after.
Limiting the UART queue to one parallel command.
Clearing the waiting queue on close and reset.


Koenkk:
Thanks!

PR #700, opened 2023-04-24
Improve Request Queue Behavior #2

slugzero:
This is the second PR to implement https://github.com/Koenkk/zigbee2mqtt/issues/17177. 
It is mostly a refactor to make future PRs more easily understandable. Apart from one fix to the previous PR, it does not change the current behavior (i.e. tests are only adapted to reflect the changes in the type definitions).

Changes:   
* Fix the send behavior for response messages: With the previous commit, response messages were not queued any more on send failure. But if the request queue was not empty, messages with sendWhen='fastpoll' and 'active'  were still queued. Now response messages are always sent immediately and never get queued. 
* Make `PendingRequest` a class and rename it to `Request`. The ZclFrame is taken out of the func() as a separate member of the Request class, to make it accessible for message filtering. The class also includes some member functions to e.g. handle reject/resolve callbacks from multiple send requests.
* Change `pendingRequests` from `Array` to `Set`, which handles asynchronous addition and removal of elements better, and allows to get rid of copying the pendingRequests array back and forth during sending. Apart from that, Set and Array should behave identically for us. A Set will not allow multiple identical elements, but queued requests should always differ, at least in their frame counter. 
* Introduce the new sendPolicies with defaults (but do not apply the policies yet).

I was not sure where the best place to put the new classes and types would be. So lmk if there is a better location and I will move them there.


src/controller/model/endpoint.ts
```diff
@@ -70,10 +96,56 @@ interface ConfiguredReporting {
     reportableChange: number,
 }
 
-interface PendingRequest {
-    /* eslint-disable-next-line @typescript-eslint/no-explicit-any*/
-    func: () => Promise<any>, resolve: (value: any) => any, reject: (error: any) => any,
-    sendWhen: 'active' | 'fastpoll', expires: number, lastError: Error
+/* eslint-disable-next-line @typescript-eslint/no-explicit-any*/
+export class Request<Type = any> {
```
Koenkk, line 100: I propose to move this to a separate file, `endpoint.ts` is getting quite big. What about `src/controller /helpers/request.ts`?

slugzero: sounds good! I moved it there, together with the defaultSendPolicy as a static member

Koenkk:
Thanks!

Something to be taken to the next pr: there is no need to prefix private class members with a `_` (since in TS you use the private modifier)

PR #706, opened 2023-05-04
Added "commandSchneiderWiserThermostatBoost"

llaissus:
Device: Schneider Electric CCTFR6400

Got the log: 2023-05-04T17:26:58.496Z zigbee-herdsman:controller:log Skipping command 'schneiderWiserThermostatBoost' because it is missing from the lookup


Koenkk:
@llaissus do you plan to use this command in a converter?

llaissus:
That's correct, I'm writing a converter for Schneider Electric CCTFR6400.
The idea is to drive a dry towel (heat and boost mode)

Koenkk:
I see, thanks!

PR #707, opened 2023-05-04
Improve request queue behavior #3: Introduce request filtering

slugzero:
This is the 3rd PR to improve the request queue behavior for sleepy end devices as discussed in https://github.com/Koenkk/zigbee2mqtt/issues/17177 

Changes:
- Introduce sendPolicy option (for override only, default send policies are used if not set)
- Introduce request filtering and apply filtering according to the send policy. By default, all read/write requests are filtered (read requests: multiple identical read requests are combined to one. write requests: only latest payload is sent, all older requests are rejected. responses are always sent immediately). If this is explicitly not wanted, the sender now has to specify the 'bulk' or 'queue' send policy. 

@Koenkk I saw your comment regarding the '_' in private members, but I will leave this for a final cleanup PR to not clutter the PRs too much with refactoring.

### DIFF
--- a/src/controller/helpers/request.ts
+++ b/src/controller/helpers/request.ts
@@ -54,6 +54,13 @@ class Request<Type = any> {
         this._lastError = lastError ?? Error("Request rejected before first send");
     }
 
+    moveCallbacks(from: Request <Type>) : void {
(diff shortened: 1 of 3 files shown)


Koenkk:
Looks good, thanks!

PR #721, opened 2023-06-21
[EZSP] Fix group binding

tomasbedrich:
Koenkk/zigbee2mqtt#18065


Koenkk:
Cool, thanks for the fix!

PR #727, opened 2023-06-26
EZSP: Removed the queue for sending commands to devices

kirovilya:
This allows to send multiple commands without waiting for a response from the device.
Useful when the signal from the device to the coordinator is bad.


Koenkk:
Thanks!

Koenkk:
@matt-oakes for z-stack it is needed to not ddos the coordinator.

matt-oakes:
Thanks for the reply! Is there not also a queue inside `znp` which should mean that only one message at a time is sent? https://github.com/Koenkk/zigbee-herdsman/blob/cb5dd54382359c3fbf73339fef5b72f2319a898c/src/adapter/z-stack/znp/znp.ts#L302 Is there a need for two queues?

I beleive EZSP also has this queue when it comes to actually sending commands https://github.com/Koenkk/zigbee-herdsman/blob/cb5dd54382359c3fbf73339fef5b72f2319a898c/src/adapter/ezsp/driver/ezsp.ts#L553

This makes sense to me as you want to serial port to be used serially, but I'm not not sure what the "outer" queue in z-stack is doing.

kirovilya:
For ezsp, I removed the queue when sending a command to a device, so as not to wait for the successful completion of the previous command for the same device.
The need was that sometimes delivering a command to the device and confirming delivery can take a long time (up to 10 seconds or more) and in case of a bad signal, it will not even reach the device. And at this time the user’s device “freezes” and does not react at all, but after a timeout, all his sent commands will come out with a timeout error.
If you remove the waiting queue, then all commands will be sent (some of them will reach the device) and only those commands that did not receive a response due to a timeout will be returned with an error. This way we keep the device available in case of communication problems.

PR #739, opened 2023-08-04
Added support for Control4 6 button keypads. (C4-KP6 model only)

ArcadeMachinist:
See https://github.com/Koenkk/zigbee2mqtt/issues/15361 for howto and converters.

Receving button presses would work with either Z-Stack or EZSP. 
However sending commands to the keypad (to set LEDs) works only with EZSP 
(eg. Sonoff ZDongle-E and other SiliconLabs based devices)

Feel free to hide this behind some "set config flags to enable feature"


Koenkk:
> However sending commands to the keypad (to set LEDs) works only with EZSP

Why doesn't this work with zstack?

ArcadeMachinist:
Because ZNP api missing calls required to set profile ID independent of endpoint config.

https://github.com/Koenkk/zigbee2mqtt/issues/15361#issuecomment-1365030097

ArcadeMachinist:
I propose we add it for everyone. As ZNP can't control LEDs, but the keypad works great.
Capable of counting sending multi clicks too.
It just won't have LED control for ZNP guys, that's all.

For example Tuya TS004F keypas also has LEDs, but there is no way to control them at all.
But still it is a "supported" device.

Just my thoughts. The decision is yours.

Also please point me to some examples on accessing low level EZSP apis from an extenstion.

Koenkk:
I missed the fact that only the led control was not working for ZNP. 

> Also please point me to some examples on accessing low level EZSP apis from an extenstion.

Here is an example of directly calling the adapter from zigbee-herdsman-converters:
https://github.com/Koenkk/zigbee-herdsman-converters/blob/ef5b19b7748ec0cd6632408acecd36262266b50d/src/devices/schneider_electric.js#L169

ArcadeMachinist:
Another question.
What is the best way to define "virtual" endpoints in the converter?
For example keypad has 6 LEDs, from what I understand the converter should expose 6 (virtual) endpoints, so the LEDs are controllable independently.
Then it would handle each virtual endpoint and construct/send real control packet.

Or there is a better approach?

Koenkk:
If you are not settings `meta: {multiEndpoint: true}` in your converter, you can add a toZigbee converter with `key: ['led_1', 'led_2', 'led_3`]`.

DCandE:
> I missed the fact that only the led control was not working for ZNP.
> 
> > Also please point me to some examples on accessing low level EZSP apis from an extenstion.
> 
> Here is an example of directly calling the adapter from zigbee-herdsman-converters: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ef5b19b7748ec0cd6632408acecd36262266b50d/src/devices/schneider_electric.js#L169

I know it's been a while but these devices are still in the wild and could be very useful to me. is there a reason the comit didn't go through to support the button presses and device.

ArcadeMachinist:
I guess it got pretty long behind the main branch.
I might take some time to update my code, I'm still runnning old zigbee-herdsman, to keep my keypads operational.
I love them and can not find anything close to those 6 button RGB led wall switches.

PR #746, opened 2023-08-15
fix: Don't remove devices with linkkey from backup if they are still present in the database

Koenkk:
Related issue: https://github.com/Koenkk/zigbee2mqtt/issues/18426


sjorge:
I guess with this in place, we could theoretically also re-add devices if they go missing but we have them in the backup?

Koenkk:
Yes, but we first have to check when exactly devices go missing.

Issue #717, opened 2023-06-07
Improve endpoint->configureReporting to handle manufacturer specific clusters/attributes


sjorge:
https://github.com/Koenkk/zigbee-herdsman/pull/712#issuecomment-1580084643

We should include some extra info so we correctly identify which manufacturer cluster and or attribute was configured.

sjorge:
@Koenkk is the TS refactor of zhc done?

Koenkk:
No not yet, I became a father a few weeks ago so I have less time to spend (so don't expect it to be done soon)

sjorge:
> No not yet, I became a father a few weeks ago so I have less time to spend (so don't expect it to be done soon)

First of all, Congratulations! Can we add the do not stale label to this issue, so it doesn't get closed.

PR #748, opened 2023-08-20
fix: Fix incorrect configured reporting cluster returned when cluster is manufacturer specific

Koenkk:
Fixes: https://github.com/Koenkk/zigbee-herdsman/issues/717

@sjorge please review the updated `controller.test.ts`


sjorge:
I’m not at home today, I have set a reminder for tomorrow so I don’t forget.

sjorge:
LGTM, I also applied the change locally and checked the H1 which was messed up before and all of the other Ubisys devices I have.

Screenshot 2023-08-20 at 22 47 48

PR #749, opened 2023-08-22
fix: emit disconnect event when zigate unplugged

yateeshc:


yateeshc:
Controller needs to know when an adapter is disconnected so that it can take appropriate clean up actions such as emitting offline event in case of Zigbee2mqtt.  This behavior is already in place for Zstack.

Koenkk:
Thanks!

PR #763, opened 2023-09-14
mdns discovery fix. adapter always zstack

Tarik2142:
Mdns discovery only worked with zstack adapters. This PR fixes this bug and the adapter will be taken from the mdns configuration

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -111,9 +111,14 @@ abstract class Adapter extends events.EventEmitter {
                             serialPortOptions.adapter = mdnsAdapter;
                             serialPortOptions.baudRate = mdnsBaud;
                             if (serialPortOptions.adapter && serialPortOptions.adapter !== 'auto') {
-                                adapter = adapterLookup[serialPortOptions.adapter];
+                                if (adapterLookup.hasOwnProperty(serialPortOptions.adapter)) {
+                                    adapter = adapterLookup[serialPortOptions.adapter];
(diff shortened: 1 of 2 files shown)


Tarik2142:
checking for **auto** makes no sense because in this case the **hasOwnProperty(serialPortOptions.adapter)** check will still fail

Tarik2142:
I apologize for the push spam, but only the **test** script works on Windows. All the others don't work, probably due to file path issues, in any case I couldn't get them to run even after trying to fix the paths
image

src/adapter/adapter.ts
```diff
@@ -110,7 +110,18 @@ abstract class Adapter extends events.EventEmitter {
                             serialPortOptions.path = `tcp://${mdnsIp}:${mdnsPort}`;
                             serialPortOptions.adapter = mdnsAdapter;
                             serialPortOptions.baudRate = mdnsBaud;
-                            resolve(new adapter(networkOptions, serialPortOptions, backupPath, adapterOptions, logger));
+                            
+                            if (adapterLookup.hasOwnProperty(serialPortOptions.adapter) 
+                            && serialPortOptions.adapter !== 'auto') {
```
Koenkk, line 115: `adapterLookup` doesn't contain `auto` so the `serialPortOptions.adapter !== 'auto'` statement is not needed

Tarik2142: I understand that, but typescript doesn't
image
image

Koenkk:
Thanks!

PR #764, opened 2023-09-17
Added manuSpecificSinope attributes

alray31:
Specific to Sinope, Calypso RM3500ZB:

- tankSize: {ID: 19, type: DataType.enum8}

Specific to Sinope thermostats:

- setpoint: {ID: 260, type: DataType.int16}
- floorTemperature: {ID: 263, type: DataType.int16}, roomTemperature: {ID: 269, type: DataType.int16}
- hvacMode: {ID: 278, type: DataType.enum8}

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -4187,6 +4187,7 @@ const Cluster: {
             outdoorTempToDisplay: {ID: 16, type: DataType.int16},
             outdoorTempToDisplayTimeout: {ID: 17, type: DataType.uint16},
             secondScreenBehavior: {ID: 18, type: DataType.enum8}, // auto:0,setpoint:1,outside:2
+            tankSize: {ID: 19, type: DataType.enum8}, //1=40gal, 2=50gal, 3=60gal, 4=80gal
             currentTimeToDisplay: {ID: 32, type: DataType.uint32},
             ledIntensityOn: {ID: 82, type: DataType.uint8},
             ledIntensityOff: {ID: 83, type: DataType.uint8},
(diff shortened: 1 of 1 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -4187,6 +4187,7 @@ const Cluster: {
             outdoorTempToDisplay: {ID: 16, type: DataType.int16},
             outdoorTempToDisplayTimeout: {ID: 17, type: DataType.uint16},
             secondScreenBehavior: {ID: 18, type: DataType.enum8}, // auto:0,setpoint:1,outside:2
+            tankSize: {ID: 19, type: DataType.enum8}, //1=40gal, 2=50gal, 3=60gal, 4=80gal
```
Koenkk, line 4190: Can you check if it still works when adding the manufacturer code? 

E.g. change `tankSize: {ID: 19, type: DataType.enum8},` to `tankSize: {ID: 19, type: DataType.enum8, manufacturerCode: ManufacturerCode.SINOPE_TECH},`

alray31: Honestly, I'm not sure how I can test it and if I had to test it before making this PR? I was under the impression that adding these attributes in cluster.ts was the first step so we can then test them from the dev console? Sorry if this was not the right way to do it. I would be happy to test what you're asking but would need some guidance, I don't know how I can poll these attributes if they aren't available in the dev console...

I'm confident the attributes from this PR are right, they come from Claude Gelinas, he's supporting Sinopé devices on ZHA. He's doing a great job gathering all the specific attributes for this brand in a table [here](https://github.com/claudegel/sinope-zha#readme).

Koenkk: You can manually edit the `cluster.js` file (which is under the `node_modules/zigbee-herdsman/dist` folder) for testing purposes.

alray31: Sorry for the delay,
Looks like herdsman won't start when using 
`tankSize: {ID: 19, type: DataType.enum8, manufacturerCode: ManufacturerCode.SINOPE_TECH},`

image


It's working fine when using  
`tankSize: {ID: 19, type: dataType_1.default.enum8},`

ANd after also adding the attribute to zigbee2mqtt/frontend, it was selectable from the dev console but trying to read it only returned an error (unsupported attribute). I'm waiting for Claude, the guy who's supporting it on ZHA. He'll switch back his unit from the manufacturer hub to ZHA and try to poll this attribute (it is defined in ZHA) to confirm if it's the right one.  I'll get back to you asap

PR #766, opened 2023-09-22
Sinope cluster attribute for RM3500ZB Calypso

claudegel:
I've added attribute 19 in Sinope cluster for tankSize parameter of RM3500ZB Calypso


claudegel:
Sorry just saw that it is already in another open PR

PR #774, opened 2023-10-12
feat(ignore):  Enable incremental TSC builds

FabianMangold:
As requeted in [zigbee-herdsman-converters PR#6273](https://github.com/Koenkk/zigbee-herdsman-converters/pull/6273) 😄 

For the sake of consistency with the other components (i.e zigbee-herdsman-converters), a cleanup is performed prior packing (i.e pre-pack hook).
For that reason, the build pipeline itself doesn't take advantage of the incremental build speed improvement.

If desired, we could still optimize the CI pipelines in all of the projects by removing the prepack hooks, but it is best practice to do a "clean build" prior packing / publishing.
Hope that implementation is OK.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ dist/
 .idea
 temp/
 .remote-sync.json
+tsconfig.tsbuildinfo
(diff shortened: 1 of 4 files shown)


Koenkk:
Thanks!

PR #777, opened 2023-10-12
Update cluster.ts

smarts4home:
added command 0xf0 for cluster 0x0502 required for Bosch Outdoor Siren

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -2872,6 +2872,12 @@ const Cluster: {
                     {name: 'strobelevel', type: DataType.uint8},
                 ],
             },
+            boschOutdoorSiren: {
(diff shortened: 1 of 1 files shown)


src/zcl/definition/cluster.ts
```diff
@@ -2872,6 +2872,12 @@ const Cluster: {
                     {name: 'strobelevel', type: DataType.uint8},
                 ],
             },
+            boschOutdoorSiren: {
+                ID: 0xf0,
+                parameters: [
+                    {name: 'data', type: DataType.uint8},
```
Koenkk, line 2878:
```suggestion
                    {name: 'data', type: DataType.uint8, manufacturerCode: ManufacturerCode.Bosch},
```

Can you check if it also works with this?

PR #780, opened 2023-10-15
add Inovelli VZM35 cluster parameters

nathanfiscus:
Adds support for the recently released VZM35 fan controller.  Renamed Manufacturer specific cluster to remove the device name since it used by multiple devices.


Koenkk:
Thanks!

PR #786, opened 2023-10-25
Fix reporting for Elko Super TR

arnerek:
Remove the manufacturer code to fix the reporting for Elko Super TR. See https://github.com/Koenkk/zigbee2mqtt/issues/18619


Koenkk:
Thanks!

PR #787, opened 2023-10-26
fix: AttributeReport: Erroneous attribute names

FabianMangold:
Reparse cluster attributes names using the target device's manufacturer ID, as certain devices fail to set the manufacturerSpecific and customerCode data in the ZCL Frame header.

Hi @Koenkk,
with reference to the issue mentioned here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6333
What do you think? Is this acceptable ?

Cheers

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -683,7 +683,7 @@ class Controller extends events.EventEmitter {
             if (frame.isGlobal()) {
                 if (frame.isCommand('report')) {
                     type = 'attributeReport';
-                    data = ZclFrameConverter.attributeKeyValue(dataPayload.frame);
+                    data = ZclFrameConverter.attributeKeyValue(dataPayload.frame, device.manufacturerID);
                 } else if (frame.isCommand('read')) {
                     type = 'read';
                     data = ZclFrameConverter.attributeList(dataPayload.frame);
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -686,15 +686,15 @@ class Controller extends events.EventEmitter {
                     data = ZclFrameConverter.attributeKeyValue(dataPayload.frame, device.manufacturerID);
                 } else if (frame.isCommand('read')) {
                     type = 'read';
-                    data = ZclFrameConverter.attributeList(dataPayload.frame);
+                    data = ZclFrameConverter.attributeList(dataPayload.frame, device.manufacturerID);
                 } else if (frame.isCommand('write')) {
                     type = 'write';
-                    data = ZclFrameConverter.attributeKeyValue(dataPayload.frame);
+                    data = ZclFrameConverter.attributeKeyValue(dataPayload.frame, device.manufacturerID);
                 } else {
                     /* istanbul ignore else */
                     if (frame.isCommand('readRsp')) {
                         type = 'readResponse';
-                        data = ZclFrameConverter.attributeKeyValue(dataPayload.frame);
+                        data = ZclFrameConverter.attributeKeyValue(dataPayload.frame, device.manufacturerID);
                     }
                 }
             } else {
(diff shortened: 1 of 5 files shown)


src/controller/helpers/zclFrameConverter.ts
```diff
@@ -1,19 +1,26 @@
-import {ZclFrame} from '../../zcl';
+import {ZclFrame, Utils as ZclUtils} from '../../zcl';
+import Debug from "debug";
 
 interface KeyValue {[s: string]: number | string}
 
-function attributeKeyValue(frame: ZclFrame): KeyValue {
+function attributeKeyValue(frame: ZclFrame, manufacturerID?: number): KeyValue {
```
Koenkk, line 6: Please make this argument mandatory

src/controller/helpers/zclFrameConverter.ts
```diff
@@ -1,19 +1,26 @@
...
     const payload: KeyValue = {};
+    let cluster = frame.Cluster;
```
Koenkk, line 8: The if below can be removed, just use: `let cluster = ZclUtils.getCluster(frame.Cluster.ID, manufacturerID);` here

src/controller/helpers/zclFrameConverter.ts
```diff
@@ -1,19 +1,26 @@
...
+        Debug.log(`Attempting to resolve manufacturer specific (${manufacturerID}) cluster attributes.`);
+        cluster = ZclUtils.getCluster(frame.Cluster.ID, manufacturerID);
+    }
 
     for (const item of frame.Payload) {
         try {
-            const attribute = frame.Cluster.getAttribute(item.attrId);
+            const attribute = cluster.getAttribute(item.attrId);
             payload[attribute.name] = item.attrData;
+            Debug.log(`Attr: ${item.attrId} (${attribute.name}) -> ${item.attrData}}`);
```
Koenkk, line 19: I propose to not add any debug logging here, it will pollute the log files a lot.

Koenkk:
Looks good!

Could you make similar changes to `attributeList` (it suffers from the same issue)

src/controller/controller.ts
```diff
@@ -683,7 +683,13 @@ class Controller extends events.EventEmitter {
             if (frame.isGlobal()) {
                 if (frame.isCommand('report')) {
                     type = 'attributeReport';
-                    data = ZclFrameConverter.attributeKeyValue(dataPayload.frame);
+                    // Certain devices (e.g. Legrand/4129) fail to set the manufacturerSpecific flag and
```
Koenkk, line 686: Please move these comments to `zclFrameConverters.ts`, the only change needed here is:

```js
data = ZclFrameConverter.attributeKeyValue(dataPayload.frame, device.manufacturerID);
```

FabianMangold:
Thanks for your feedback @Koenkk.
I've updated the PR accordingly, but that brought changes to 2 additional models (Endpoint and Device).
The tests are failing, i'll take a look shortly.

Koenkk:
@FabianMangold no need to hurry, I will merge this after the next z2m release so we have enough time to test this in dev.

FabianMangold:
Hello @Koenkk,
No worries, it is already fixed. I have integrated all of your suggestion, but had to implement some null checks due to the bogus frames used in the tests!
Hope that is OK now, let me know.

Koenkk:
Looks good!

Would you mind adding a test case to `controller.test.ts` to test the specific case for the Legrand device?

FabianMangold:
Of course, let me check how this can be mocked best, not sure how to do this yet :).

FabianMangold:
Hi @Koenkk,

I've added 4 new tests to the suite, to validate all use-cases.
Let me know if this is OK
Thanks a lot.

Koenkk:
Thanks!

PR #793, opened 2023-11-05
fix(typo): change to `commission`

Rotzbua:
Not sure if this rename has side effect on other repos because it is a keyword.


Koenkk:
There are also reference to this in zigbee-herdsman-converters, please also create a PR for that.

Koenkk:
Thanks!

PR #536, opened 2022-05-11
Use standard Zcl parser for green power frames

filips:
As far as I can tell, the Green Power Indication received should contain the full ZCL frame for the green power message. Thus, we could get a more correct and complete parsing of the contents by using the zcl parser. 

I have not tested this, as I do not have a Conbee II at my disposal.

Related issue: https://github.com/Koenkk/zigbee2mqtt/issues/12414

PR #795, opened 2023-11-06
Deconz ZGP Fixes

konistehrad:
Following in the footsteps of https://github.com/Koenkk/zigbee-herdsman/pull/536 I'm trying to spruce up ZGP handling a bit between Deconz and Z2M. Namely: I'm trying to reduce the amount of bespoke processing happening in the deconz driver and move more of it to the controller. This is the first push, which specifically fixes "options" handling.

I could really use some guidance on the frame parsing side, though. https://github.com/dresden-elektronik/deconz-serial-protocol/issues/13#issuecomment-992586453 indicates that this should be just a raw ZGP frame pushed down the serial pipe, but for GP commissioning notifications data is in places I simply don't expect it to be. Why is options at offset 14, and frameCounter 36 for example? I'm sure there's something I'm missing here.

### DIFF
--- a/src/adapter/deconz/adapter/deconzAdapter.ts
+++ b/src/adapter/deconz/adapter/deconzAdapter.ts
@@ -1069,9 +1069,8 @@ class DeconzAdapter extends Adapter {
     private checkReceivedGreenPowerIndication(ind: gpDataInd) {
         ind.clusterId = 0x21;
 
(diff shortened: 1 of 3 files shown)


Koenkk:
> I could really use some guidance on the frame parsing side

@ChrisHae is this something you could help with?

PR #797, opened 2023-11-11
fix: Restrict attribute name remapping (PR#787)

FabianMangold:
* Restrict to Legrand only following regressions

Reference:
https://github.com/Koenkk/zigbee-herdsman/commit/d8e4b5bac01478b54f43190e775cb14f72c17b43#commitcomment-132312842

### DIFF
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -1,5 +1,6 @@
 import {ZclFrame, Utils as ZclUtils} from '../../zcl';
 import {Cluster} from '../../zcl/tstype';
+import ManufacturerCode from '../../zcl/definition/manufacturerCode';
 
 interface KeyValue {[s: string]: number | string}
 
@@ -12,7 +13,7 @@ interface KeyValue {[s: string]: number | string}
 function getCluster(frame: ZclFrame, deviceManufacturerID: number): Cluster {
     let cluster = frame.Cluster;
 
-    if (!frame?.Header?.manufacturerCode && frame?.Cluster && deviceManufacturerID == 4129) {
+    if (!frame?.Header?.manufacturerCode && frame?.Cluster && deviceManufacturerID == ManufacturerCode.LegrandNetatmo) {
         cluster = ZclUtils.getCluster(frame.Cluster.ID, deviceManufacturerID);
     }
     return cluster;
--- a/src/controller/helpers/zclFrameConverter.ts
+++ b/src/controller/helpers/zclFrameConverter.ts
@@ -12,8 +12,7 @@ interface KeyValue {[s: string]: number | string}
 function getCluster(frame: ZclFrame, deviceManufacturerID: number): Cluster {
     let cluster = frame.Cluster;
 
-    if (!frame?.Header?.manufacturerCode && frame?.Cluster &&
-         Number.isInteger(deviceManufacturerID) && deviceManufacturerID == 4129) {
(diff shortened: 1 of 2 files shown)


src/controller/helpers/zclFrameConverter.ts
```diff
@@ -8,10 +8,12 @@ interface KeyValue {[s: string]: number | string}
 // This leads to incorrect reported attribute names.
 // Remap the attributes using the target device's manufacturer ID
 // if the header is lacking the information.
+// Note: Limit this feature to Legrand / 4129 only following regressions with Tuya devices
 function getCluster(frame: ZclFrame, deviceManufacturerID: number): Cluster {
     let cluster = frame.Cluster;
 
-    if (!frame?.Header?.manufacturerCode && frame?.Cluster && Number.isInteger(deviceManufacturerID)) {
+    if (!frame?.Header?.manufacturerCode && frame?.Cluster &&
+         Number.isInteger(deviceManufacturerID) && deviceManufacturerID == 4129) {
```
Koenkk, line 16: I think there is no need to do the integer check here, just `deviceManufacturerID == 4129` should be good enough.

FabianMangold:
The INT check is required due to some bogus tests, those may fail otherwise.
That check was already part of the first PR if I recall correctly 

On 12 Nov 2023, at 08:49, Koen Kanters ***@***.***> wrote:

﻿


@Koenkk commented on this pull request.

--------------------------------
In src/controller/helpers/zclFrameConverter.ts:


>  function getCluster(frame: ZclFrame, deviceManufacturerID: number): Cluster {
     let cluster = frame.Cluster;
 
-    if (!frame?.Header?.manufacturerCode && frame?.Cluster && Number.isInteger(deviceManufacturerID)) {
+    if (!frame?.Header?.manufacturerCode && frame?.Cluster &&
+         Number.isInteger(deviceManufacturerID) && deviceManufacturerID == 4129) {


I think there is no need to do the integer check here, just deviceManufacturerID == 4129 should be good enough.

src/controller/helpers/zclFrameConverter.ts
FabianMangold: The INT check is required due to some bogus tests, those may fail otherwise.
That check was already part of the first PR if I recall correctly.

Koenkk: - How many tests are failing? Could those be fixed?
- I don't see a reason to revert #787 as that will break the Legrand use case?

FabianMangold: HI @Koenkk,
Sorry for the delay in getting back, i am currently on a business trip.

* There are 4 occurrences of missing manufacturerID at a quick glance. Those might be legitimate, I am not yet sure which tests are issuing those requests.
* I've removed the integer check, as you are right, it is definitly not required. I've gone through the history, and I really cannot recall why i have introduced that check really (quite sure it must have had a good reason!)..
So sorry for that, it is removed now!

* Yes, it would break the Legrand use case, if implementing the devices as specified by the Legrand documentation. Having that said, the documentation is inconsistent too.
As mentioned by @sjorge, the fix helps on "certain" devices, and may "break" others, depending on the firmware.
Would it make sense to add a new "meta" annotation, that could be enabled on a device basis in order to explicitly enable the remapping of the attributes instead of hardcoding the manufacturerID here ?

Koenkk:
Looks good now, thanks!

sjorge:
All good again for my ubisys devices after this follow up! 👍

PR #800, opened 2023-11-12
Add Profalux Manufacturer Cluster

ghoz:
Profalux covers have 2 custom clusters 0xfc20 and 0xfc21

0xfc21 seems to be used for configuration
Attribute 0 stores the configuration for the type of cover, allowing us to detect if tilt needs to be exposed  by https://github.com/Koenkk/zigbee-herdsman-converters/pull/6459


Koenkk:
Thanks!

PR #806, opened 2023-11-20
Add tunnel cluter

xsp1989:
I resubmitted the code, and by the way, I would like to explain the modifications to the buffaloZcl.ts file. When I read it, all the remaining data was returned. I checked all the codes, and it seemed that no code would use BuffaloZclDataType.BUFFER in the response. Data type should have no effect.


Koenkk:
Thanks!

PR #813, opened 2023-12-01
Revert "feat: Add `tradfriButton` cluster 

thatslolo:
This reverts commit 6e454d86823d5c0c47c454a71a96f1062e4f077c.

the new clusters broke the dots on symfonisk remote.
Another implementation proposed here: Koenkk/zigbee-herdsman-converters#6574

Alternatively we need one more event `commandAction6` and a rework of the symfonisk implementation

### DIFF
--- a/src/controller/events.ts
+++ b/src/controller/events.ts
@@ -137,12 +137,6 @@ const CommandsLookup: {[s: string]: MessagePayloadType} = {
     
     // Schneider
     'schneiderWiserThermostatBoost':'commandSchneiderWiserThermostatBoost',
-
-    // Tradfri
-    'action1': 'commandAction1',
-    'action2': 'commandAction2',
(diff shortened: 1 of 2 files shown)


Koenkk:
I think it's better to update the symfonisk integration, could you make a PR for this?

thatslolo:
I see two options: 

1. 
- Add double click to the tradfriButton cluster
- create new integration for symfonisk and somrig based on the new cluster
2. 
- revert the commit
- keep using symfonisk integration
- reuse symfonisk integration for somrig

I think option 2 would mitigate risk and is less work. 

I won't have the time to change both integrations at least for this  week.

Koenkk:
I will fix this in https://github.com/Koenkk/zigbee2mqtt/discussions/20003

PR #817, opened 2023-12-04
Improve Request Queue Behavior #5

slugzero:
This is the 5th PR in a series to improve the request queue behavior for sleepy end devices, as discussed in https://github.com/Koenkk/zigbee2mqtt/issues/17177.

This PR introduces the following change: If an attempt to send a message after an implicit checkin fails, the message is not deleted immediately. Instead, all messages are now kept until one of the following is met: 
- message is delivered successfully
- send attempt fails after an explicit checkin
- the message has expired
- message is overwritten by a new message with the same payload

The PR also introduces another small fix that an incoming genPollCtrl message does not trigger an implicit checkin any more. The proper behavior is to send a checkin response and wait for the device to enter fastpoll, then send all messages. The implicit checkin is unnecessary and may lead to bad side effects like messages being lost or sent out of sequence. 

PS: this one took me a little bit longer because I did not have sufficient time to test it in the last months. Next one will follow more quickly :-)


Koenkk:
Thanks!

PR #824, opened 2023-12-09
Update Inovelli cluster definition

futurexdesign:
Update Inovelli cluster definition to add quick start parameters for `VZM35` and `VZM36` devices.


Koenkk:
Thanks!

PR #829, opened 2023-12-16
feat: add displayUnit and airQuality attributes to aqaraOpple

sjorge:
For https://github.com/Koenkk/zigbee-herdsman-converters/pull/6709, so we can switch those to modernExtend (enum)Lookups.

I guess we should technically rename this to something like lumiManufacturerSpecific because that seems to be what it is, but that seems like a huge rename I do not want to attempt.

Just some small improvements to VOCKQJK11LM is quickly growing into a big refactor I fear :(


Koenkk:
Thanks!

PR #835, opened 2023-12-22
zcl: update ubisys hvacThermostat custom attributes

sjorge:
I am unable to find any references to manuSpecificUbisysHeatingRegulatorThermostat in zhc at the moment, I will do a PR to fix the Ubisys H1 that needs to be merged after this.

I did verify the attribute exist in the H10 technical reference though, so I'd like to keep them around at least if someone does do more work on the H10.

This also solves our attribute collision issue for now, as ubisys moved them to avoid a collision after feedback I provided.


sjorge:
Hmmm interesting, I'm getting INVALID_TYPE now but I copied the type for there docs which matched the old type I was using.

sjorge:
I also noticed some of the attribute ID's were off by 1 (as per https://www.ubisys.de/wp-content/uploads/ubisys-h10-technical-reference.pdf, page 30)

Koenkk:
Thanks!

PR #836, opened 2023-12-23
Add calibrate valve command for Bosch BTH-RA

DerDreschner:
The Bosch BTH-RA has the ability to trigger the valve calibration remotely instead of pushing the button on the device itself. This pull request adds the necessary command to the `hvacThermostat` cluster.


Koenkk:
Thanks!

PR #839, opened 2023-12-23
feat: add Amazon Work With All Hubs cluster

sjorge:
I've seen the WWAH cluster on a few of my devices, mostly if they have the 'works with amazon echo' sticker on the box.

I never found a attribute list but Ubisys was kind enough to provide it when I asked. I have no plans on use this, but it looks like it might be interesting to collect device metrics and perhaps with the auto generation for unsupported devices.

otaMaxOfflineDuration might be useful for the OTA process too, that should indicate how long we need to wait before assuming the device is back.

### DIFF
--- a/src/zcl/definition/cluster.ts
+++ b/src/zcl/definition/cluster.ts
@@ -5701,6 +5701,38 @@ const Cluster: {
         commands: {},
         commandsResponse: {},
     },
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Issue #803, opened 2023-11-17
Cannot read statuses and control devices. Adapter: ezsp


JakubGoliasz:
### What happened?

I cannot control Zigbee devices or read their status. There is also a crash at the end.

*Adapter: ezsp*

### What did you expect to happen?

Ability to receive messages from devices and control them.

### How to reproduce it (minimal and precise)

Just run the Zigbee2MQTT 1.33.2 add-on in HA. Rollback to 1.33.1-1 helps.

*Adapter: ezsp*

### Zigbee2MQTT version

1.33.2

### Adapter firmware version

6.10.3.0 build 297

### Adapter

SONOFF Zigbee 3.0 USB Dongle

### Debug log

```
Zigbee2MQTT:debug 2023-11-02 07:31:08: Publishing 'set' 'state' to 'Lampa'
Zigbee2MQTT:error 2023-11-02 07:31:08: Publish 'set' 'state' to 'Lampa' failed: 'Error: Command 0x04cf8cdf3c7e4d4c/1 genOnOff.off({}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":false,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (sendZclFrameToEndpointInternal error)'
Zigbee2MQTT:debug 2023-11-02 07:31:08: Error: Command 0x04cf8cdf3c7e4d4c/1 genOnOff.off({}, {"sendWhen":"immediate","timeout":10000,"disableResponse":false,"disableRecovery":false,"disableDefaultResponse":false,"direction":0,"srcEndpoint":null,"reservedBits":0,"manufacturerCode":null,"transactionSequenceNumber":null,"writeUndiv":false}) failed (sendZclFrameToEndpointInternal error)
    at EZSPAdapter.sendZclFrameToEndpointInternal (/app/node_modules/zigbee-herdsman/src/adapter/ezsp/adapter/ezspAdapter.ts:471:19)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at Queue.executeNext (/app/node_modules/zigbee-herdsman/src/utils/queue.ts:32:32)
```

```
[...]

PR #841, opened 2023-12-26
Missing parentheses to call function

Nerivec:
Some errors seem to fail at the wrong time when serial port never actually opens properly because of this line ([fails here instead](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/ezsp/driver/ezsp.ts#L550)).

https://github.com/Koenkk/zigbee-herdsman/issues/803#issuecomment-1823539146
https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/issues/546


Koenkk:
Thanks! 🎄

PR #842, opened 2023-12-26
Add breezemode to Inovelli Cluster

futurexdesign:
Add "Breeze Mode"  parameter to the Inovelli cluster for VZM36.   Parameter is new in latest beta firmware.


Koenkk:
Thanks!

PR #844, opened 2023-12-27
endpoint.configureReporting should use manufacturerCode from ZCL definition when available

### LINKED ISSUE #843: endpoint.configureReporting does not take into account attribute specific manufacturerCode ?
https://github.com/Koenkk/zigbee-herdsman/blob/599b3e96afe9ace3cb0231ff069fc3ff70e43b2e/src/controller/model/endpoint.ts#L619

I'm debugging a weird issue I'm seeing on a Ubisys H1, it looks like when I send:
```
mosquitto_pub -t zigbee2mqtt/bridge/request/device/configure_reporting -m '{"attribute":"ubisysVacationMode","cluster":"hvacThermostat","id":"trv/office/1","maximum_report_interval":3600,"minimum_report_interval":0,"reportable_change":0}'
```

It ends up configuring `occupiedHeatingSetpoint` for reporting not `ubisysVacationMode`, it seems it somehow ignores the manufactureringCode specified in the definitions.ts for the ubisysVacationMode attribute. 

I'm not sure yet why or how though.

sjorge:
I ran into this when trying to update the Ubisys H1, unless you manually specify the options (the frontend does not), we try to configure it without specifying the manufacturerCode. In this case resulting in a weird INVALID_DATA_TYPE error.

When our ZCL Cluster definition has an attribute manufacturerCode override specified, we should use it.

I worked backwards from what I usually do... which means I probably did it in the correct order 😬 
I wrote a test first that configures hvacThermostat.ubisysVacationMode and swap the device in the mock to one with the Ubisys manufacturerCode. And update the manufacturerCode, attributeID and attributeDataType in the expected data to match those of ubisysVacationMode.

```

 FAIL  test/controller.test.ts (24.601 s)
  ● Controller › Endpoint configure reporting for manufacturer specific attribute from definition

    expect(received).toStrictEqual(expected) // deep equality

    - Expected  - 2
    + Received  + 2

    @@ -73,14 +73,14 @@
          "commandIdentifier": 6,
          "frameControl": Object {
            "direction": 0,
            "disableDefaultResponse": true,
            "frameType": 0,
    -       "manufacturerSpecific": true,
    +       "manufacturerSpecific": false,
            "reservedBits": 0,
          },
    -     "manufacturerCode": 4338,
    +     "manufacturerCode": null,
          "transactionSequenceNumber": 11,
        },
        "Payload": Array [
          Object {
            "attrId": 18,

      2752 |         expect(call[1]).toBe(129);
      2753 |         expect(call[2]).toBe(1)
    > 2754 |         expect({...deepClone(call[3]), Cluster: {}}).toStrictEqual({"Cluster":{},"Command":{"ID":6,"name":"configReport","parameters":[{"name":"direction","type":32},{"name":"attrId","type":33},{"conditions":[{"type":"directionEquals","value":0}],"name":"dataType","type":32},{"conditions":[{"type":"directionEquals","value":0}],"name":"minRepIntval","type":33},{"conditions":[{"type":"directionEquals","value":0}],"name":"maxRepIntval","type":33},{"conditions":[{"type":"directionEquals","value":0},{"type":"dataTypeValueTypeEquals","value":"ANALOG"}],"name":"repChange","type":1000},{"conditions":[{"type":"directionEquals","value":1}],"name":"timeout","type":33}],"response":7},"Header":{"commandIdentifier":6,"frameControl":{"direction":0,"disableDefaultResponse":true,"frameType":0,"manufacturerSpecific":true,"reservedBits":0},"manufacturerCode":4338,"transactionSequenceNumber":11},"Payload":[{"attrId":18,"dataType":16,"direction":0,"maxRepIntval":10,"minRepIntval":1,"repChange":1}]});
           |                                                      ^
      2755 |
      2756 |         expect(endpoint.configuredReportings.length).toBe(1);
      2757 |         expect({...endpoint.configuredReportings[0], cluster: undefined}).toStrictEqual({"attribute":{"ID":18,"type":16,"manufacturerCode":4338,"name":"ubisysVacationMode"},"minimumReportInterval":1,"maximumReportInterval":10,"reportableChange":1, "cluster": undefined});

      at toStrictEqual (test/controller.test.ts:2754:54)
      at call (test/controller.test.ts:2:1)
      at Generator.tryCatch (test/controller.test.ts:2:1)
      at Generator._invoke [as next] (test/controller.test.ts:2:1)
      at asyncGeneratorStep (test/controller.test.ts:2:1)
      at asyncGeneratorStep (test/controller.test.ts:2:1)
```

That test failed (as expected), after updating endpoint.configureReporting, the new test now passes without manually specifying the manufacturerCode option.

`if (attribute.hasOwnProperty('manufacturerCode')) options.manufacturerCode = attribute.manufacturerCode;` doesn't feel very TypeScript to me so there may be a cleaner way to write the same line of code ?

This closes #843, I had hoped to be able to add the new features to the H1 before end of year, but since I ran into this it probably going to be for early next year.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -633,7 +633,9 @@ class Endpoint extends Entity {
                     const attribute = cluster.getAttribute(item.attribute);
                     dataType = attribute.type;
                     attrId = attribute.ID;
-                    if (attribute.hasOwnProperty('manufacturerCode')) options.manufacturerCode = attribute.manufacturerCode;
+                    if (attribute.hasOwnProperty('manufacturerCode')) {
+	                options.manufacturerCode = attribute.manufacturerCode;
+		    }
                 }
             }
 
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -634,8 +634,8 @@ class Endpoint extends Entity {
                     dataType = attribute.type;
                     attrId = attribute.ID;
                     if (attribute.hasOwnProperty('manufacturerCode')) {
-	                options.manufacturerCode = attribute.manufacturerCode;
-		    }
+                        options.manufacturerCode = attribute.manufacturerCode;
+                    }
                 }
             }
 
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -633,6 +633,7 @@ class Endpoint extends Entity {
                     const attribute = cluster.getAttribute(item.attribute);
                     dataType = attribute.type;
                     attrId = attribute.ID;
+                    if (attribute.hasOwnProperty('manufacturerCode')) options.manufacturerCode = attribute.manufacturerCode;
                 }
             }
 
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -2733,6 +2733,30 @@ describe('Controller', () => {
         expect({...endpoint.configuredReportings[0], cluster: undefined}).toStrictEqual({"attribute":{"ID":16384,"type":48,"manufacturerCode":4641,"name":"viessmannWindowOpenInternal"},"minimumReportInterval":1,"maximumReportInterval":10,"reportableChange":1, "cluster": undefined});
     });
 
+    it('Endpoint configure reporting for manufacturer specific attribute from definition', async () => {
+        await controller.start();
+        await mockAdapterEvents['deviceJoined']({networkAddress: 129, ieeeAddr: '0x129'});
+        const device = controller.getDeviceByIeeeAddr('0x129');
+        device._manufacturerID = 0x10f2;
+        const endpoint = device.getEndpoint(1);
+        mocksendZclFrameToEndpoint.mockClear();
+        await endpoint.configureReporting('hvacThermostat', [{
+            attribute: 'ubisysVacationMode',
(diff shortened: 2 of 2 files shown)


sjorge:
Edit, it looks like there is a type error now. I think the type might be wrong though, as the zcl cluster definition were we get this data from does have an optional manufacturerCode.

src/controller/model/endpoint.ts
```diff
@@ -633,6 +633,7 @@ class Endpoint extends Entity {
                     const attribute = cluster.getAttribute(item.attribute);
                     dataType = attribute.type;
                     attrId = attribute.ID;
+                    if (attribute.hasOwnProperty('manufacturerCode')) options.manufacturerCode = attribute.manufacturerCode;
```
Koenkk, line 636: Could you try if just `options.manufacturerCode = attribute.manufacturerCode;` also works here?

sjorge: That would override the one from https://github.com/Koenkk/zigbee-herdsman/blob/9d122f3a531f6487f4e7f3b737ebf70549ded5c0/src/controller/model/endpoint.ts#L623 by undefined if the attribute doesn't have one set

Koenkk:
Thanks again!

Issue #847, opened 2023-12-27
endpoind.read() should also use attribute specific manufacturerCode when available


sjorge:
Related to #843 we should also just use the information when doing a read.

However this does not seems as straight forward as with the reporting configuration.

Since `endpoint.read()` can take a array of attributes. As long as they have the same manufacturerCode requirement this is not an issue but in theory it could be a mixed, with and without (I doubt it we will see with different ones in one read)! For example one could read `localTemp` + `ubisysRemoteTemperature` from `hvacThermostat` 

https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/zcl/definition/cluster.ts#L1974

https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/zcl/definition/cluster.ts#L2085

https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/controller/model/endpoint.ts#L445-L447

The information would be available in the loop above in the `cluster.getAttribute()` result.

We should split this into N frames:

https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/controller/model/endpoint.ts#L449-L453 would need to be called one time for each manufacturerCode (null also being one if an attribute has none)

We would also need to collect the responses:
https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/controller/model/endpoint.ts#L462-L465

I think before we have not hit this as we seem to always provide the manufacturerCode ourselves, but there should be no reason to do so as we have the information.

A alternative and simpler solution would be to lookup the manufacturerCode here:
https://github.com/Koenkk/zigbee-herdsman/blob/5411ab6ce77e218b076628337e7001c59914f3d5/src/controller/model/endpoint.ts#L445-L447

But only update `options.manufacturerCode` if it is currently undefined, if it is already defined we check if `options.manufacturerCode` matches `attribute.manufacturerCode` and throw an exception if it doesn't.

It would then be up to the called to split the `endpoint.read()` so instead of reading both `localTemp` + `ubisysRemoteTemperature` , one would read `localTemp` and then `ubisysRemoteTemperature`.

[...]

PR #848, opened 2023-12-28
#847 Use cluster definition manufacturerCode when available

sjorge:
@Koenkk there is currently a failure of `Read from endpoint with custom attribute`, I think because mock device 129 does not have any data for ubisysDefaultOccupiedHeatingSetpoint to return ? Not sure how to fixed it, I tried simply adding `ubisysDefaultOccupiedHeatingSetpoint: 1800` for ep1 attribute data for device 129 but that didn't work.

We probably need a new mockDevice though, then we can also drop the `device._manufacturerID = 0x10f2;` we have in a few tests.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -362,6 +362,26 @@ class Endpoint extends Entity {
     ): Promise<void> {
         const cluster = Zcl.Utils.getCluster(clusterKey);
         options = this.getOptionsWithDefaults(options, true, Zcl.Direction.CLIENT_TO_SERVER, cluster.manufacturerCode);
+
+        const manufacturerCodes = new Set(Object.keys(attributes).map((nameOrID): string|number => {
+            // we fall back to caller|cluster provided manufacturerCode
+            if (cluster.hasAttribute(nameOrID)) {
+                const attribute = cluster.getAttribute(nameOrID);
+                return (attribute.manufacturerCode === undefined) ?
+		    options.manufacturerCode :
+		    attribute.manufacturerCode;
+            } else {
+                // unknown attribute, we should not fail on this here
+                return options.manufacturerCode;
+            }
+        }));
+        if (manufacturerCodes.size == 1) {
+            options.manufacturerCode = manufacturerCodes.values().next().value;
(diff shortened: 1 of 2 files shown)


src/controller/model/endpoint.ts
```diff
@@ -362,6 +362,26 @@ class Endpoint extends Entity {
     ): Promise<void> {
         const cluster = Zcl.Utils.getCluster(clusterKey);
         options = this.getOptionsWithDefaults(options, true, Zcl.Direction.CLIENT_TO_SERVER, cluster.manufacturerCode);
+
```
Koenkk, line 365: Can we move this and the 2 similar blocks below to a function to prevent some code duplication?

sjorge: I tried but only read + write one are the same, the data we have in configureReporting is different.

sjorge: Actually after some cleanup I did I think I see a way to merge them. Let me try.

sjorge: Bah the map line is also different on all 3 -_-

sjorge:
@Koenkk managed to move it to a private function

sjorge:
@Koenkk I had another go at fixing the broken test, but I couldn't. I think the mock device needs some extra data (or we need a new mock device)

src/controller/model/endpoint.ts
```diff
@@ -826,6 +835,42 @@ class Endpoint extends Entity {
         };
     }
 
+    private ensureManufacturerCodeIsUnique(
```
Koenkk, line 838:
```suggestion
    private ensureManufacturerCodeIsUniqueAndGet(
```

Koenkk:
Fixed the test and added a small last comment, let me know if this can be merged now.

sjorge:
I did the rename, so should be ready.

So for the test there was a check on the cluster to return the mocked frame and hvacThermostat wasn't listed. Makes sense.

sjorge:
Probably best to merge after the January release.

Koenkk:
Thanks!

PR #853, opened 2024-01-04
fix: catch more unhandled promise rejections

mikekap:
I think I have a somewhat bad adapter setup, but when upgrading zigbee2mqtt from 1.33.1 to 1.35.0 I get pretty consistent crashes due to unhandled promise rejections. This fixes the rejections that I encountered.

### DIFF
--- a/src/adapter/ezsp/driver/ezsp.ts
+++ b/src/adapter/ezsp/driver/ezsp.ts
@@ -263,6 +263,7 @@ export class Ezsp extends EventEmitter {
     }
 
     public async connect(path: string, options: Record<string, number|boolean>): Promise<void> {
+        let lastError = null;
         for (let i = 1; i < 5; i += 1) {
             try {
                 await this.serialDriver.connect(path, options);
@@ -271,10 +272,11 @@ export class Ezsp extends EventEmitter {
                 debug.error(`Connection attempt ${i} error: ${error.stack}`);
                 await Wait(5000);
                 debug.log(`Next attempt ${i+1}`);
+                lastError = error;
             }
         }
         if (!this.serialDriver.isInitialized()) {
-            throw new Error("Failure to connect");
+            throw new Error("Failure to connect", {cause: lastError});
         }
         if (WATCHDOG_WAKE_PERIOD) {
             this.watchdogTimer = setInterval(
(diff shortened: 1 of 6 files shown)


src/adapter/ezsp/driver/ezsp.ts
```diff
@@ -271,10 +272,11 @@ export class Ezsp extends EventEmitter {
                 debug.error(`Connection attempt ${i} error: ${error.stack}`);
                 await Wait(5000);
                 debug.log(`Next attempt ${i+1}`);
+                lastError = error;
             }
         }
         if (!this.serialDriver.isInitialized()) {
-            throw new Error("Failure to connect");
+            throw new Error("Failure to connect", {cause: lastError});
```
mikekap, line 279: Not strictly necessary since it's in the logs already, but let me know if you'd like me to remove.

src/adapter/ezsp/driver/uart.ts
```diff
@@ -83,29 +83,35 @@ export class SerialDriver extends EventEmitter {
...
-        });
+
+            debug('Serialport opened');
+            this.serialPort.once('close', this.onPortClose.bind(this));
+            this.serialPort.once('error', (error) => {
+                debug(`Serialport error: ${error}`);
+            });
+    
+            // reset
+            await this.reset();
```
mikekap, line 104: if resetting failed here, the serial port would be stuck open forever.

src/utils/queue.ts
```diff
@@ -15,29 +13,28 @@ class Queue {
         this.concurrent = concurrent;
     }
 
-    public execute<T>(func: () => Promise<T>, key: string | number = null): Promise<T> {
-        return new Promise((resolve, reject): void =>  {
-            this.jobs.push({key, func, running: false, resolve, reject});
+    public async execute<T>(func: () => Promise<T>, key: string | number = null): Promise<T> {
```
mikekap, line 16: the change here isn't strictly necessary, but it makes for nicer async stack traces since the call to `func` is in the same async stack chain as the caller.

src/adapter/ezsp/driver/ezsp.ts
Koenkk: It's fine

src/utils/queue.ts
Koenkk: Cool, thanks!

Koenkk:
Looks good, could you take a look at the failing build? Then this can be merged.

Koenkk:
Many tanks!

PR #858, opened 2024-01-07
Basic support for EZSP 13 (gecko 4.4.0, ncp 7.4.0.0)

kirovilya:


src/adapter/ezsp/adapter/ezspAdapter.ts
```diff
@@ -611,7 +611,7 @@ class EZSPAdapter extends Adapter {
     }
 
     public async supportsBackup(): Promise<boolean> {
-        return true;
+        return (this.driver && this.driver.ezsp && this.driver.ezsp.ezspV < 13);
```
xconverge, line 614: What change causes the need for backup to be disabled here?

Nerivec: Also `return this.driver?.ezsp?.ezspV < 13` can make it more concise.

kirovilya: > What change causes the need for backup to be disabled (for version 13) here, or is the version is 0 based?

I disable backup for ezsp 13... Receiving keys and storing them still needs to be worked out.

in any case, even if there is a backup, it is not restored yet - I have not yet completed the correct restoration

kirovilya:
can be merged. the current code may already work. without backup

Koenkk:
Awesome, many thanks for the super quick update!

src/adapter/ezsp/adapter/ezspAdapter.ts
Hedda: > What change causes the need for backup to be disabled (for version 13) here, or is the version is 0 based?

@kirovilya @xconverge Note! "This release of EmberZNet drops support for reading/writing network keys using our existing methods, breaking compatibility. The new commands thankfully maintain feature parity and include new features, allowing us to fully restore APS link keys." This might break backup and restore features. More info posted by puddly here -> https://github.com/zigpy/bellows/pull/603

xconverge:
Thanks a lot @kirovilya !

PR #859, opened 2024-01-08
Improve serialport code consistency across adapters.

Nerivec:
Proposing this to make debugging/readability easier across adapters. It's mostly cleanup and a switch to base class functions, but it needs a good look-over since it affects init code.

### DIFF
--- a/src/adapter/z-stack/znp/znp.ts
+++ b/src/adapter/z-stack/znp/znp.ts
@@ -281,6 +281,8 @@ class Znp extends events.EventEmitter {
                 this.socketPort.destroy();
             }
         }
+
+        this.emit('close');
     }
 
     public request(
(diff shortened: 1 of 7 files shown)


src/adapter/z-stack/znp/znp.ts
```diff
@@ -138,28 +139,28 @@ class Znp extends events.EventEmitter {
...
-        });
+        try {
+            await this.serialPort.asyncOpen();
+            debug.log('Serialport opened');
+
+            this.serialPort.once('close', this.onPortClose.bind(this));
+            this.serialPort.once('error', this.onPortError.bind(this));
+
+            this.initialized = true;
+            this.emit('connected');
```
Koenkk, line 152: I noticed this has been added but the event doesn't seem to be used anywhere (same for `ezsp/driver/uart.ts` where it was already present)

Nerivec: Yes it could be removed entirely indeed, if there is no intent to use it; probably leftover code, but I went conservative on removals at first.

Koenkk: Let's remove it

Koenkk:
Thanks!

PR #860, opened 2024-01-08
Improve Request Queue Behavior #6

slugzero:
This is the 6th and final PR in a series to implement https://github.com/Koenkk/zigbee2mqtt/issues/17177.

It removes the special behavior for `sendWhen=fastpoll` devices (https://github.com/Koenkk/zigbee-herdsman/commit/568abc3d2118531c1713180d06eaadefaca2f44b). I.e. the same queue behavior is now consistently used for all devices according to their `pendingRequestTimeout` and per-request `sendPolicy`.

The rest of the PR is cleanup: After the removal of the special behavior, `sendWhen` is not used anywhere any more and therefore I removed it from the request queue. To maintain backwards compatibility, I added conversion functions for the public interfaces. I also added tests for these functions and removed some other now-obsolete tests.


src/controller/model/device.ts
```diff
@@ -102,9 +101,19 @@ class Device extends Entity {
     set skipDefaultResponse(skipDefaultResponse: boolean) {this._skipDefaultResponse = skipDefaultResponse;}
     get skipTimeResponse(): boolean {return this._skipTimeResponse;}
     set skipTimeResponse(skipTimeResponse: boolean) {this._skipTimeResponse = skipTimeResponse;}
-    get defaultSendRequestWhen(): SendRequestWhen {return this._defaultSendRequestWhen;}
     set defaultSendRequestWhen(defaultSendRequestWhen: SendRequestWhen) {
```
Koenkk, line 104: There are only 3 references to this in zigbee-herdsman-converters, could we refactor those and remove this entirely?

slugzero: would be fine for me, too - I just thougt we could give it a little grace period, remove the 3 occcurrences (and maybe also all occurrences of `sendWhen`) and then remove the adapters. 
Should I better remove it right away and prepare a PR for zhc?

Koenkk: Let's do it right away, I will merge the PRs at the same time.

slugzero: Alright, corresponding zhc PR is here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6865

src/controller/model/endpoint.ts
```diff
@@ -277,21 +277,31 @@ class Endpoint extends Entity {
                 options.disableResponse, options.disableRecovery, options.srcEndpoint) as Promise<Type>;
         }): Promise<Type> {
         const logPrefix = `Request Queue (${this.deviceIeeeAddress}/${this.ID}): `;
-        const request = new Request(func, frame, this.getDevice().pendingRequestTimeout, options.sendWhen,
-            options.sendPolicy);
+
+        if(options.sendWhen) {
```
Koenkk, line 281: Could we also remove this part? (maybe in a next PR?)

Koenkk:
Thanks! I think the `option.sendWhen` can also be removed (https://github.com/Koenkk/zigbee-herdsman/pull/860#discussion_r1446480753)

src/controller/model/endpoint.ts
slugzero: Absolutely! And yes, that will be another PR:-)

Koenkk: Many thanks for your dedication 😄

PR #867, opened 2024-01-11
fix: Add setter for checkinInterval

slugzero:
See discussion here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/6865#issuecomment-1887342676

This is required because the pendingRequestTimeout is not persisted and therefore I'd like to change the quirk in zhc to `quirkCheckinInterval`.

Note: Checkin interval is a device property that should be fixed and that is persisted in the database. PendingRequestTimeout is intended to be changed dynamically if needed. It is reset to default on z2m restart.


Koenkk:
Thanks!

PR #871, opened 2024-01-13
EZSP: Implemented several attempts to send a request in case of problems

kirovilya:
The extended timeout setting has been removed (we need to analyze the need to set it).


Koenkk:
Thanks!

PR #879, opened 2024-01-19
fix: remove `options.sendWhen`

slugzero:
Follow-up PR for https://github.com/Koenkk/zigbee-herdsman/pull/860#issuecomment-1883606553


Koenkk:
Awesome 😄

PR #883, opened 2024-01-22
EZSP: Added base for `incomingNetworkStatusHandler`

Nerivec:
These "stack errors" are poorly documented in the SDK, but at least it gives a better idea for debugging for now.

PS: @Koenkk Can we bump lint max-len to 150 across all repos for consistency?


Koenkk:
> Can we bump lint max-len to 150 across all repos for consistency?

Fine with that!

PR #886, opened 2024-01-24
EZSP: Fix socket error handling & serial/socket reset/close logic.

Nerivec:
Align net.Socket logic with serialport's:
- On initial `error` event (when called by EZSP layer) => reject promise with error (EZSP layer handles it)
- On connected/ready
  - register `close` & `error` handlers

`close` event of serial/socket port will now signal UART>EZSP>Driver layers to reset on unexpected disconnect. This will result in a proper shutdown and then another round of reconnect tries. Same as the EZSP watchdog does. If it fails to reset, it will bubble up to Z2M and stop.

Added `disconnected` event-firing in the adapter layer to match global Z2M logic.

Removed the infinite retries loop at the Driver layer. While that might have allowed eventual automatic reconnection, there is the matter of a fail at the hardware level; since it was re-doing the full init with every try, if it failed further down the line than port-connect, this could have resulted in the loop endlessly crashing the hardware every 60 seconds (although somewhat unlikely to actually "crash" it, the mere possibility is enough to make me think it shouldn't be there...). And if it failed to connect to the port the first 4 times (the most likely scenario), there are very few reasons to think it would work any better with more time...
We keep the 4 tries of port-connect spaced by increments of 5 seconds. We can see about increasing this as needed.

@kirovilya your thoughts?

### DIFF
--- a/src/adapter/ezsp/adapter/ezspAdapter.ts
+++ b/src/adapter/ezsp/adapter/ezspAdapter.ts
@@ -40,6 +40,7 @@ class EZSPAdapter extends Adapter {
     private interpanLock: boolean;
     private backupMan: EZSPAdapterBackup;
     private queue: Queue;
+    private closing: boolean;
 
 
     public constructor(networkOptions: NetworkOptions,
@@ -50,12 +51,14 @@ class EZSPAdapter extends Adapter {
             this.waitressValidator, this.waitressTimeoutFormatter
         );
         this.interpanLock = false;
+        this.closing = false;
 
         const concurrent = adapterOptions && adapterOptions.concurrent ? adapterOptions.concurrent : 8;
         debug(`Adapter concurrent: ${concurrent}`);
         this.queue = new Queue(concurrent);
 
         this.driver = new Driver(this.serialPortOptions, this.networkOptions, this.greenPowerGroup);
+        this.driver.on('close', this.onDriverClose.bind(this));
         this.driver.on('deviceJoined', this.handleDeviceJoin.bind(this));
         this.driver.on('deviceLeft', this.handleDeviceLeft.bind(this));
         this.driver.on('incomingMessage', this.processMessage.bind(this));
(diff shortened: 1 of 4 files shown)


kirovilya:
It’s difficult to immediately understand the changed logic.
I would like the following problems to be solved (I tried to solve them before).
1. Initial connection failed. To solve the problem of late launch of multi-pan addon.
2. Communication failure that occurred after a successful connection. For example, they pulled out the stick and stuck it in again. Or the wifi/ethernet router crashes/reboots during a tcp connection.
3. Failure at the UART level of the data exchange protocol. Reconnect.

In any of these cases, adapter must attempt to connect while pausing.

Nerivec:
1. We should be able to adjust retries with `SERIAL_CONNECT_NEW_ATTEMPT_MIN_DELAY`, I left it at 5sec, but changed the logic to factor in attempt #, so it will try immediately, then retry 5sec after fail, then 10sec after new fail, then 15sec after new fail. I'd tend to put the issue in the same category as below "needs to be handled higher up the Z2M chain".
2. I don't believe any other adapter supports this right now (it all ends around [here](https://github.com/Koenkk/zigbee2mqtt/blob/3c9620428ea00d759bd1001322b0ead55eca2285/lib/controller.ts#L212) on any disconnect). That should be handled higher up than the ezsp adapter to support all at once if desired. (Actually, right now, EZSP is the only one that will retry at all with below, others will just fail if I'm not mistaken. @Koenkk can you confirm?)
3. I believe the reset at the UART level triggered by a disconnect (serial/socket) should take care of that. It will attempt a reset at the Driver level, which means it will then retry up to 4 times at the serial level. Same as the EZSP watchdog does.

kirovilya:
if so, then good

Koenkk:
@Nerivec 

> (Actually, right now, EZSP is the only one that will retry at all with below, others will just fail if I'm not mistaken. @Koenkk can you confirm?)

Indeed, zstack will just fail and stop z2m (supervisor can then restart z2m)

Nerivec:
@koenkk I'd suggest posting a discussion (your voice will have more impact than mine 😉) in Z2M repo to get some feedback from current testers on edge, since there were quite a few changes in EZSP recently; before this ever goes into the stable version. There are so many configuration possibilities, some might see different results...

Koenkk:
[Created] and pinned it. In case you need a more direct communication channel, feel free to contact me on telegram (@koenkk)

PR #888, opened 2024-01-27
Some more backup support for EZSP 13+

Nerivec:
Continuing on https://github.com/Koenkk/zigbee-herdsman/pull/858

This should properly populate the existing fields for `backup()`. 
All the new key-related commands seem to use SLStatus (darn thing is uint32_t!).

@kirovilya As always, if you can double-check...

Still quite a lot more to do to support full backup... I'm still going through the code...

### DIFF
--- a/src/adapter/ezsp/adapter/backup.ts
+++ b/src/adapter/ezsp/adapter/backup.ts
@@ -2,7 +2,13 @@
 import Debug from "debug";
 import {Driver} from '../driver';
 import * as Models from "../../../models";
-import {EmberKeyType, EmberKeyStruct, EmberNetworkParameters} from '../driver/types';
+import {
+    EmberKeyType,
+    EmberKeyStruct,
+    EmberNetworkParameters,
+    EmberSecurityManagerNetworkKeyInfo,
+    EmberKeyData
+} from '../driver/types';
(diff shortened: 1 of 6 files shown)


kirovilya:
This is great! Thank you for supporting the EZSP adapter (I’m terribly short of time).
In addition to backup, it is necessary to implement network restoration. I started doing it here.
https://github.com/kirovilya/zigbee-herdsman/tree/ezsp_restore

Koenkk:
Thanks!

Nerivec:
@kirovilya I noticed a few non-awaited promises while going through the EZSP codebase. Do you mind taking a look? You can probably identify faster than me what was not awaited on purpose and what should be awaited (they're all around frames handling). You can find them easily with the eslint rule `"@typescript-eslint/no-floating-promises": "error",`

@Koenkk we might want to add that rule as base, to force explicitly disabling lint for the lines that are not awaited on purpose? I didn't go through the list, but zstack seems to have a couple also. That's the kind of omission that can end up in weird-hard-to-trace errors...

kirovilya:
@Nerivec It seems that at the UART level one error handler does not wait for the reset.
The rest are not so critical (IMHO).

PR #890, opened 2024-01-28
cluster: vindstyrka does not not send manufacturerCode

sjorge:
When just dropping manufacturerCode there is a conflict. Docoding seems to work fine even if the DP does not match. Configure reporting in zhc will need a bit of extra help.


Koenkk:
Thanks!

PR #896, opened 2024-01-29
Enforce no-floating-promises (or explicit disabling).

Nerivec:
Prevents mistakenly calling a promise without awaiting for it (or explicitly disabling lint if not needed).

Of course this will fail CI (like I said, I saw a couple in z-stack at least). I don't want to blindly disable them all just to pass CI, in case one is actually a potential problem.
I don't know the other stacks to make the appropriate changes. I'll let you do that @Koenkk.


Nerivec:
Agreed. Looks like tests will need adjustments too.

Koenkk:
Thanks!

PR #897, opened 2024-01-30
EZSP: Enable software flow control only if RTS/CTS not enabled.

Nerivec:
Also covered an edge case where the host driver did not take care of XON/XOFF bytes. It will assert in the logs even if herdsman debug is off. This really shouldn't happen... but at least we'll know if some configs are causing trouble (the "hard-to-identify" bug kind).


Koenkk:
Thanks!

PR #905, opened 2024-02-03
docs: Add links to automaticaly generated ref docs

krzysztof-kwitt:
Hello,

I noticed that our API documentation was not updated since 4 years, so I propose to switch to typedocs.dev. It maintains that for us


Koenkk:
Thanks!

Issue #910, opened 2024-02-05
Z2M 1.35.2 stops with "Adapter disconnected, stopping" after few minutes of uptime (Sonoff-E / EZSP v12 / FW 7.3.2.0 build 212)


romarysonrier:
**1./ A first, i was running Z2M 1.35.1 withon a 2 devices zigbee network :**
- SONOFF	SNZB-02D
- LiXee	ZLinky_TIC

Radio Adapter is SONOFF -E / EFR32 MGM21 (upgraded to EZSP v12 / FW 7.3.2.0 build 212).
Radio link is strong and healthy

The  ZLinky_TIC device is using a lot of polling to report some measurement related to my Electric Meter (Linky).
The most simple way to monitor how well the polling is working is to track how often is updated a TIMEDATE value from the ZLinky_TIC device.

While running Z2M 1.35.1, i was experiencing some unexpected random freeze of any polling activities from Z2M, after some hours of uptime. Z2M never stopped or crashed. Only data from Reportable attributes were still updated. Network scan was stuck and not responding once the polling issue was observed.

When working, the polling did offers lower refresh rate for ZLinky_TICthan expected ( 60 sec requested),  TIMEDATE from Electric Meter was updated every 70sec to 12 min, average was near 2 min. And at least a day the whole netflow was suffering from lost of all polled attributes.

I struggled to identify a clear root cause at this point, could also be hardware since i am using DIY case.

**2./ But, after upgrading  Z2M 1.35.2, the 'abnormal' behaviour changed:**
- refresh rate for ZLinky_TIC is almost always 70sec (good progress)
- Z2M stops working randomly after some minutes of uptime ("Adapter disconnected, stopping" , no suspicious log) and restart automaticaly 

```
info  2024-02-05 14:15:36: MQTT publish: topic 'zigbee2mqtt-grenier/ENEDIS Linky Portail', payload '{...} ,"update_available":null}'
debug 2024-02-05 14:16:01: Received Zigbee message from 'ENEDIS Linky Portail', type 'attributeReport', cluster 'haElectricalMeasurement', data '{"apparentPower":217}' from endpoint 1 with groupID 0
info  2024-02-05 14:16:02: MQTT publish: topic 'zigbee2mqtt-grenier/ENEDIS Linky Portail', payload '{...} ,"update_available":null}'
error 2024-02-05 14:16:09: Adapter disconnected, stopping
[...]

PR #911, opened 2024-02-08
Ezsp on UART : implement NAK on unexpected sequence 

romarysonrier:
When Z2M get too slow to :
- read messages in the UART buffer 
- send an ACK message to EZSP FW
then EZSP FW restart transmission
- so a duplicate message will be received (with reTX flag), and should be ignored.
- if a message is really lost a NAK message must be send to EZSP FW, in order to trigger retransmission for first missed packet.

### DIFF
--- a/src/adapter/ezsp/driver/uart.ts
+++ b/src/adapter/ezsp/driver/uart.ts
@@ -44,6 +44,7 @@ export class SerialDriver extends EventEmitter {
     private sendSeq = 0; // next frame number to send
     private recvSeq = 0; // next frame number to receive
     private ackSeq = 0;  // next number after the last accepted frame
+    private recvReject = false;
     private waitress: Waitress<EZSPPacket, EZSPPacketMatcher>;
     private queue: Queue;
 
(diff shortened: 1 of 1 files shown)


Koenkk:
@Nerivec @kirovilya could you review this?

romarysonrier:
The patch is running fine since 4 days on my Raspbery pi zero , on Z2M dev branch, with lixee / Zlink_TIC device (polling fully disabled)  , which is having a huge number of reportable attribute (>15). debug_level=warn is mandatory to maintain cpu usage as low as possible.

Anyway i think we should implement also some counters to report NAK received and transmited on the web UI , so that user can diagnose how healthy is the EZSP communication. This could allow to restart automaticaly Z2M for that specific case, if desired by user.

romarysonrier:
PR related to https://github.com/Koenkk/zigbee-herdsman/issues/910 , but not solving the whole issue.

kirovilya:
not a very obvious patch. Several messages (up to 8) may arrive from the chip side, which will need to be confirmed (ACK). therefore, it is necessary to more correctly consider which frmNum is received and confirmed, and which is received but not yet confirmed.

I think here it is necessary to rework the general mechanism for processing incoming messages and immediately respond to them upon receipt, and not put them in a processing queue. those. ACC must be answered in the parser, and not at the UART level.
Although the protocol is asynchronous, we do not have time to process and our processing queue accumulates.

uart

kirovilya:
@romarysonrier I made some improvements, including ones similar to those you suggested. Thank you!
https://github.com/Koenkk/zigbee-herdsman/pull/916

Koenkk:
If I understand correctly, not needed more after https://github.com/Koenkk/zigbee-herdsman/pull/916

PR #915, opened 2024-02-12
added missing attributes for clipsalWiserSwitchConfigurationClusterServer

Nailik:
- added missing attribtues
- actually the same as for elko


Koenkk:
Thanks!

PR #924, opened 2024-02-20
[ember] Improve errors & checks + GreenPower support

Nerivec:
Cleaned up some error messages, and added extra check on frame parsing just in case... All for stuff that should never happen 😉


Nerivec:
Alright, might as well make this PR actually useful.
GreenPower should now have a proper implementation!
It was tested with:
- https://www.sunricher.com/zigbee-self-powered-remote-greenpower-sr-zgp2801ks.html
- https://nodon.fr/en/nodon/philips-hue-battery-less-switch/

Both are recognized as `PTM 215Z` by Z2M converters, and have been confirmed working.

Big thanks to @dduransseau for the remote testing! 😉

PR #926, opened 2024-02-23
[ember] Fix install code CRC validation.

Nerivec:
Validations tested locally again actual install code.
Remains to see if the code will now be added properly on the NCP side...


Koenkk:
thanks!

PR #932, opened 2024-02-25
[ember] Reworked errors in ASH protocol

Nerivec:
This should relax the erroring-out at lower levels a bit, and allow the EZSP layer to retry init when it fails to reach `CONNECTED` state on first attempt (will try 5 times).

_This should affect only "problem adapters/firmware". I'm probably at over 500 init sequences by now with all the testing, and have yet to see an error or a fail at this level (unless I forced one)..._


Koenkk:
Thanks!

PR #937, opened 2024-02-27
Update cluster.ts

DanielNagy:
Create new clipsalWiser Fan switch cluster for  https://github.com/Koenkk/zigbee2mqtt/issues/18620


src/zcl/definition/cluster.ts
```diff
@@ -5029,6 +5029,17 @@ const Cluster: {
         commands: {},
         commandsResponse: {},
     },
+    clipsalWiserFanSwitchConfigurationCluster: {
+        ID: 0xFC04,
+        manufacturerCode: ManufacturerCode.SCHNEIDER,
+        attributes: {
+            SwitchIndication: {ID: 0x0002, type: DataType.uint8},
+            SwitchOrientation: {ID: 0x0060, type: DataType.uint8},
```
Koenkk, line 5037:
```suggestion
            switchIndication: {ID: 0x0002, type: DataType.uint8},
            switchOrientation: {ID: 0x0060, type: DataType.uint8},
```

DanielNagy:
@Koenkk sorry, was fighting with github web ui to do your recommended change.. Please review now :)

src/zcl/definition/cluster.ts
```diff
@@ -5029,6 +5029,16 @@ const Cluster: {
         commands: {},
         commandsResponse: {},
     },
+    clipsalWiserFanSwitchConfigurationCluster: {
```
sjorge, line 5032: Can we give it the `manuSpecific` prefix, most newer manufacturer specific clusters start with that prefix (@Koenkk we should probably at some point rename all the others, but that seems like it will require a lot of PRs against zhc too)

Also it seems the manufcaturer is schneider so probaby something like `manuSpecificSchneiderFanSwitchConfiguration`

src/zcl/definition/cluster.ts
```diff
@@ -5029,6 +5029,16 @@ const Cluster: {
...
+        ID: 0xFC04,
+        manufacturerCode: ManufacturerCode.SCHNEIDER,
+        attributes: {
+            switchIndication: {ID: 0x0002, type: DataType.uint8},
```
sjorge, line 5036: We can probably also drop the `switch` prefix here as it already clear we're talking about a switch from the cluster name.

src/zcl/definition/cluster.ts
DanielNagy: I'm open to whatever naming convention you all wanna go with. 

Just need it to merged so we can update the device that requires it. 👍

src/zcl/definition/cluster.ts
DanielNagy: Hindsight, this probably should be named Led indication, rather than switch indication. Will adjust once naming of cluster is agreed upon.

src/zcl/definition/cluster.ts
lucasteligioridis: @sjorge not that I don't disagree with the statements about the convention here, but I wonder if its better to do these as a big sweep and stick to the existing convention or not? If you see the cluster above this one which this is based off you'll see how its named as well as the attribute naming.

Not sure what the best approach is 👍🏼

sjorge: Not adding new ones that do follow the convention would be a good start I think.

Otherwise we're just making more extra work when we rename the others later.

DanielNagy:
@sjorge @Koenkk Please confirm latest changes.

Koenkk:
Thanks!

PR #938, opened 2024-02-27
EZSP: Fixed connection of devices via routers

kirovilya:
Starting from version 1.35.2, connecting devices via a router did not work.
When connecting through a router, the device did not receive a key, although it entered the network and began the interview process.
Then, it left the network.
And then everything repeated again.


Koenkk:
Thanks!

PR #942, opened 2024-03-01
Refactor names to follow convention in SchneiderSwitch cluster

lucasteligioridis:
# What?

Just modifies the names according to what the convention should be in this cluster configuration.

cc @sjorge discussion on conventions from a previous PR, just cleaning it up while we're doing work here.


Koenkk:
Thanks!

PR #945, opened 2024-03-02
[ember][tempfix] Skip port re-init on reset retry

Nerivec:
Seems the RESET_WATCHDOG firmware issue is indeed plaguing SkyConnect/Yellow according to early 1.36 reports with `ember`...
Since it appears to be linked to the opening of the serial port, I believe this should offer a temp fix until the firmware trouble can be resolved.
_It prevents a true clean slate on restart, but I guess for now it will have to do... The impact should only be on serial/writer/parser, so I don't think too many troubles should arise from it..._

NOTE: While I tested the reset logic to work as I expected, I can't actually test on a SkyConnect/Yellow since I don't have one... i.e. attempting a fix blindfolded...

### DIFF
--- a/src/adapter/ember/uart/ash.ts
+++ b/src/adapter/ember/uart/ash.ts
@@ -318,7 +318,7 @@ export class UartAsh extends EventEmitter {
             return false;
         }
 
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #951, opened 2024-03-04
[ember] Drastically lower default requests & callbacks processing delay

Nerivec:
The delay can be configured with the [advanced config](https://www.zigbee2mqtt.io/guide/configuration/adapter-settings.html#advanced-configuration) `adapter_delay`.
Default is now 5 (if not provided/invalid), max is 60, min is 5.
For reference, the default was previously 60 (static).

_Fixed a bunch of comments._


Koenkk:
Thanks!

PR #965, opened 2024-03-09
fix: Don't emit deviceLeave event for devices that already left

Koenkk:
CC: @Nerivec

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -471,10 +471,13 @@ class Controller extends events.EventEmitter {
         debug.log(`Device leave '${payload.ieeeAddr}'`);
 
         const device = Device.byIeeeAddr(payload.ieeeAddr);
-        if (device) {
-            debug.log(`Removing device from database '${payload.ieeeAddr}'`);
-            device.removeFromDatabase();
+        if (!device) {
+            debug.log(`Device leave is from unknown device '${payload.ieeeAddr}'`);
+            return;
         }
+
+        debug.log(`Removing device from database '${payload.ieeeAddr}'`);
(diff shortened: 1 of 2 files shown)


test/controller.test.ts
```diff
@@ -936,8 +936,8 @@ describe('Controller', () => {
 
         // leaves another time when not in database
         await mockAdapterEvents['deviceLeave']({networkAddress: 129, ieeeAddr: '0x129'});
-        expect(events.deviceLeave.length).toBe(2);
-        expect(events.deviceLeave[1]).toStrictEqual({ieeeAddr: '0x129'});
+        expect(events.deviceLeave.length).toBe(1);
+        expect(events.deviceLeave[0]).toStrictEqual({ieeeAddr: '0x129'});
```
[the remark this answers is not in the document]
Koenkk: True, done!

src/controller/controller.ts
```diff
@@ -471,11 +471,14 @@ class Controller extends events.EventEmitter {
         debug.log(`Device leave '${payload.ieeeAddr}'`);
 
         const device = Device.byIeeeAddr(payload.ieeeAddr);
-        if (device) {
-            debug.log(`Removing device from database '${payload.ieeeAddr}'`);
-            device.removeFromDatabase();
+        if (!device) {
+            debug.log(`Device leave is from unknown device '${payload.ieeeAddr}'`);
```
Nerivec, line 475: `Device leave is from unknown or already deleted device`
I think would that be more precise for debug purposes.

Koenkk: done!

PR #968, opened 2024-03-10
New command and attribute for Tuya-based devices

kirovilya:


Koenkk:
Thanks!

9runner:
Hi, after upgrade to 1.36.1, unfortunately, all my tuya TS0726 Action expose is down, it didn't response any button press, is it relate to this new function? if it is, how can i adopt with this update? 

many thinks!

9runner:
I do have a converter running, but it is only related with a light controller. I will also double check it by removing this converter away, and lets see what will happen.

9runner:
after eliminated all the converter, upgraded to 1.36.1, all the Action stay quiet again, so i rolled back to 1.36.0again.  

looking forward for your good news , thanks!

9runner:
Thank you Kirovilya!

I have a lot of TS0726 4 or 3 or 2 gang Tuya switchs, but what the Z2M reconization is looks like this:

截屏2024-04-07 12 09 17

截屏2024-04-07 12 11 51

I double checked several times, once upgrade 1.36.1, the Action expose will keep quiet, once downgrade to 1.36.0, every thing back to normal.

Hope my proble is not related with your update. I will keep on trying to figure out what is wrong with my system.

BTW, even stay with 1.36.0, my switch Action keep on bouncing every time when the scenc buttom pressed, I did try to fixing this issue by extenal converter, but since I am a newbee on Z2M, I nevery get the converter woking. Right now what I do is using a several micro seconds delay to eleminating the debounce problem. And another problem is that I find my switchs would provide backlight setting in Tuya native App, but I can't get this expose on Z2M.  It will be very appriciated if you can give me any advice on those issues. Thanks a lot!

kirovilya:
ah, I found the problem... I'll fix it now

9runner:
thank you Kirovilya!

BTW, how can I enjoy this update? just upgrade to 1.36.1 or I need to use dev version?

kirovilya:
You can upgrade to the dev/edge version until May

9runner:
thanks a lot!

Ilya Kirov ***@***.***> 于 2024年4月8日周一 17:53写道：

PR #970, opened 2024-03-10
[ember] Increase default NCP config + edge case device leave support

Nerivec:
Add some more firmware current defaults for reference + links.
Increase default NCP configuration for max children and binding table. _The changes are pretty drastic, but the initial values were actually very low compared to all the firmware defaults out there._

Allow Z2M to know when an ID conflict forced the stack to remove devices (edge case scenario). _Allow lookup by network address in `Controller>onDeviceLeave` to make this work without extra NCP call._

Move default APS options to const.

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -470,7 +470,7 @@ class Controller extends events.EventEmitter {
     private onDeviceLeave(payload: AdapterEvents.DeviceLeavePayload): void {
         debug.log(`Device leave '${payload.ieeeAddr}'`);
 
-        const device = Device.byIeeeAddr(payload.ieeeAddr);
+        const device = payload.ieeeAddr ? Device.byIeeeAddr(payload.ieeeAddr) : Device.byNetworkAddress(payload.networkAddress);
         if (!device) {
             debug.log(`Device leave is from unknown or already deleted device '${payload.ieeeAddr}'`);
             return;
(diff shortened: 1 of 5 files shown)


Nerivec:
@Koenkk Please confirm controller test/logic as expected. PS: I have to fix that darn coverage bug on my machine :(

src/controller/controller.ts
```diff
@@ -470,16 +470,16 @@ class Controller extends events.EventEmitter {
     private onDeviceLeave(payload: AdapterEvents.DeviceLeavePayload): void {
         debug.log(`Device leave '${payload.ieeeAddr}'`);
 
-        const device = Device.byIeeeAddr(payload.ieeeAddr);
+        const device = payload.ieeeAddr ? Device.byIeeeAddr(payload.ieeeAddr) : Device.byNetworkAddress(payload.networkAddress);
```
Koenkk, line 473: - Does this only happen on deviceLeave?
- I propose to update the `DeviceLeavePayload` type to: `type DeviceLeavePayload = {networkAddress?: number, ieeeAddr: string} | {networkAddress: number, ieeeAddr?: string}
`

Nerivec: - Yes, for now that is the only place it is needed anyway.
- Indeed that would be better. Is it currently an `interface` though.

Koenkk:
Thanks!

Issue #933, opened 2024-02-25
Load additional manufacturer-specific clusters from device converter


rgiordano30:
Hi, with new zigbee chips like Espressif ESP32-C6, more and more personal/hobby projects using the Zigbee protocol are starting to appear.

Problem is that manufacturer-specific clusters (and attributes) are hardcoded in zigbee-herdsman and there are no ways to extend them without modifing the source code.

It's would be great to add a new "cluster" option to the device converter and load it in zigbee-herdsman when parsing messages.

sjorge:
Been wanting this for a while too, it would be great to move those non ZCL ones to the device definitions.

@koenkk I wonder if can somehow leverage some of the extend stuff to inject (and also share some between devices)

Although thinking about it some more that won't work as all the code to resolve these lives in zh. 🤔

Koenkk:
@sjorge it needs to be injected somehow into ZH during startup, can be done in `onEvent` but didn't had the time to look at this yet.

Suxsem:
@Koenkk  Yeah I'm really waiting for this one too, it would open a world of possibilities for makers!
If you are willing to provide me some insights in how you would implement this I'm happy to try making a PR!

Suxsem:
@Koenkk 
I see that in https://github.com/Koenkk/zigbee2mqtt/blob/2a53b8e59b457e0805318bdb38d2c74de651724d/lib/zigbee.ts#L61
ZH is initialized with an herdsmanSettings object, do you think this could be a good place to inject manufacturer-specific clusters?
Or this settings should only contain global (aka adapter) settings?

PR #971, opened 2024-03-11
feat: Load additional cluster definitions from options

Suxsem:
Related to https://github.com/Koenkk/zigbee-herdsman/issues/933

This PR add a new (optional) option to the controller constructor in order to handle additional (manufacturer-specific) cluster definitions.

An additional PR to Z2M will follow to use this new feature in order to load additional cluster definitions from external files.


Koenkk:
Instead of adding it to the global clusters, I propose to inject them on the device level, such that the cluster stays device specific.

Koenkk:
> still the clusters are technically manufacturer specific and not device specific

The problem is that different manufacturer define cluster using the same ID, therefore they need to be per device.

> but I don't think that adding the custom cluster definition to the database is a good idea for memory reasons, what do you think?

It's indeed not a good idea, so I was thinking they are injected when Z2M starts using the [definition onEvent start](https://github.com/Koenkk/zigbee-herdsman-converters/blob/45a4a41ef4b90eabe4a35fe37f5df769a3c12edf/src/devices/lixee.ts#L1035). 

So what needs to happen:
- Add a method to [`device.ts`](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts) which allows to add a custom cluster definition (it's saved to a property, not to the database)
- Add a `device` parameter to [`ZclUtils.getCluster`](https://github.com/Koenkk/zigbee-herdsman/blob/b0781d2bdecdbc9fa8c68e129675722cc12d001d/src/zcl/utils.ts#L34)

Suxsem:
@Koenkk Oh I didn't know about the onEvent definition option!
I'm going to re-implement the feature as you suggested, give me a couple of days. Thank you!

Suxsem:
@Koenkk I have implemented your suggestions about device.ts and ZclUtils.getCluster, but I'm having hard time changing calls to ZclUtils.getCluster.

For example, getCluster is called in zclFrame.ts and buffalocl.ts but I don't know how to retrieve a device object (or the device ieee address in order to call the static method Device.byIeeeAddr) in those classes. Any help?

Nerivec:
Just a note on changing ZCL lookup function signatures:

> how to retrieve a device object (or the device ieee address in order to call the static method Device.byIeeeAddr)

`zclFrame` & `buffaloZcl` being called at pretty low levels (stack drivers), it may or may not be IEEE that is available (could be network address).

Also, calling Device.by{x} at such low levels, means it may be called multiple times, since `ZclDataPayload`, passed by drivers, also contains `address` for [device lookup higher up the Z2M chain](https://github.com/Koenkk/zigbee-herdsman/blob/c7cbfcbab8502d8dd0eb3eba295843d1352afee9/src/controller/controller.ts#L629).

Koenkk:
I think the tricky ones is where the ZclFrame is created at the driver level, e.g. [this call](https://github.com/Koenkk/zigbee-herdsman/blob/c7cbfcbab8502d8dd0eb3eba295843d1352afee9/src/adapter/z-stack/adapter/zStackAdapter.ts#L818). I was thinking about pushing the ZclFrame creation up to the controller level, so that would be [here](https://github.com/Koenkk/zigbee-herdsman/blob/c7cbfcbab8502d8dd0eb3eba295843d1352afee9/src/controller/controller.ts#L610) (we already do the device lookup there). This also allows us to get rid of the nasty [try/catch](https://github.com/Koenkk/zigbee-herdsman/blob/c7cbfcbab8502d8dd0eb3eba295843d1352afee9/src/adapter/z-stack/adapter/zStackAdapter.ts#L816) logic. The adapter will then only emit what is currently the `RawDataPayload`

Koenkk:
>  Sure it's not better to add the specific clusters to the global clusters list instead of keeping them in the device object?

The problem is that there are conflicting cluster, e.g. manufacturer A defines cluster X with ID 20, manufacturer B defines cluster Y with ID 20, how would you handle this?

Suxsem:
Manufacturer is already matched when searching for correct cluster in zclutils.getCljster, isn't it?

Koenkk:
@Suxsem not all have the manufacturer code set

Suxsem:
Oh I see. So I think we should ensure we have at least the address when getCluster is called in order to find the correct device. As I said I don't fully know the project architecture so I'll live you discuss the best way to proceed, please tell me if there is something I can do to help

Nerivec:
Just to recap a few things here:

- Moving the `fromBuffer` logic up to the Controller would be good.
  - Would also likely require moving all (or at least some) waitress logic up to the Controller (matching uses `fromBuffer`-derived properties in many cases). Probably good too, but refactoring of all stacks involved to some degree 😉
- ZCL logic (basically everything under [zcl folder](https://github.com/Koenkk/zigbee-herdsman/tree/master/src/zcl)) should remain unaware of `Device` (or anything else really), to keep the spec separate from the Z2M logic.

Koenkk:
@Nerivec perfectly summarised, I will start on the fromBuffer refactor

sjorge:
> * Can't trust that the same attribute ID will use the same data type... (same as above really just from a different point of view)

Just for reference and practical examples. https://github.com/Koenkk/zigbee-herdsman/blob/85447fbc839fd5d0b42fc7708c2b61486368b3f8/src/zcl/definition/cluster.ts#L2681 the IKEA STARKVYND uses the same cluster and attribute ID's for their pm25 measurement but use SinglePerc DataType. So we do infact already encounter these. We currently have some hacks in place to 'make this work' https://github.com/Koenkk/zigbee-herdsman-converters/blob/523c1e7147a2178382c9cb883e047bffdd0c2a18/src/devices/ikea.ts#L1184-L1187

But this is far from ideal was you can't simple reconfigure them using the frontend, or without some knowledge via MQTT. We're just lucky that for the payload parsing logic all number variants are treated the same, uint8, uint16, int8, int16, singlePerc... all end up as a typescript number. If this was written in a more strict type language we would probably not be so lucky.

Suxsem:
Let's close this PR for now waiting for a better strategy

Suxsem:
Hi @Koenkk, sorry for asking, but did you managed to think about the best way to proceed with this one or already started?

Koenkk:
@Suxsem yes, its on my todo list

Suxsem:
Great glad to hear that, happy to help if there is something specific I could help with. Thank you

Koenkk:
Started working on this in https://github.com/Koenkk/zigbee-herdsman/pull/1011

Suxsem:
Awesome news, thank you!!

PR #976, opened 2024-03-15
Added non-CSA-assigned codes currently used.

Nerivec:
As far as I could find, these are not referenced in CSA website, nor in silabs manuf codes, so I'm assuming they are not assigned to them... _We can rename/move them if that changes._

re: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7223


Koenkk:
Thanks!

PR #977, opened 2024-03-16
Base support for channel switching without re-pairing

Nerivec:
Adds base for supporting channel switching without having to re-pair devices.
Whenever `channel` is changed in `configuration.yaml`, after restart, Z2M will trigger the switch after `Adapter.start()`, as long as no other network parameters was changed, of course, otherwise the previous logic will kick in.
Done via standard ZDO broadcast of `Mgmt_NWK_Update_req` (see 2.4.3.3.9 of Zigbee spec), so all stacks should be able to implement this eventually (for now a `supports` function takes care of ignoring the logic if not implemented).

Obviously, there will be cases where devices do not follow instructions at all (bad firmware...) and will have to be re-paired manually after the channel change; expected with brands like Aqara. In general, battery-powered devices are likely to cause more trouble than routers with this feature, but thankfully, they usually are easier to access than in-wall routers & the likes.
Still, whatever percentage of the network automatically changes is a bonus!

Observations with `ember` driver:
- Tested with mostly crap devices (so if they support it, I guess it's a good sign... 😉)
- Coordinator reports "changed" after approx. 9 seconds
- Router devices seem to follow without problem (they announce a new route, and seamlessly work, on first report/activation)
- End devices seem to follow on first report/activation (for example, a motion sensor when first detecting motion)
  - The first activation only forces the device to rejoin, the triggered state is not actually updated (behavior might vary depending on sensor, likely will be the same as when pairing)
  - Reconnect via `REJOIN`, no permit join necessary (might not be the case for all stacks?).
  - It can take up to a minute to rejoin after that first report/activation.
- Devices connected via router (not direct children of coordinator) seem to follow same as above.
  - Tested only with 1 level deep

TODOs:

- [ ] deconz
- [x] ember
- [ ] ezsp
- [ ] zstack
  - @Koenkk I did the base "on the fly", and flagged locations (with `TODO`) that should need updating if I'm not mistaken, please adjust/finish as necessary. Also needs testing to ensure it works there (no reason it wouldn't, but...). 😉
- [ ] zigate
- [x] Tests

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -114,7 +114,8 @@ import {
     SIMPLE_DESCRIPTOR_RESPONSE,
     BIND_RESPONSE,
     UNBIND_RESPONSE,
-    LEAVE_RESPONSE
+    LEAVE_RESPONSE,
+    NWK_UPDATE_REQUEST
 } from "../zdo";
 import {
     EMBER_BROADCAST_ADDRESS,
@@ -1093,8 +1094,7 @@ export class EmberAdapter extends Adapter {
 
             // XXX: should not force a form when it's only a channel change, just change the channel, wait a sec, then continue the logic
             if ((npStatus === EmberStatus.SUCCESS) && (nodeType === EmberNodeType.COORDINATOR) && (this.networkOptions.panID === netParams.panId)
-                && (equals(this.networkOptions.extendedPanID, netParams.extendedPanId))
-                && (this.networkOptions.channelList.includes(netParams.radioChannel))) {
+                && (equals(this.networkOptions.extendedPanID, netParams.extendedPanId))) {
                 // config matches adapter so far, no error, we can check the network key
                 const context = initSecurityManagerContext();
                 context.coreKeyType = SecManKeyType.NETWORK;
@@ -1157,8 +1157,6 @@ export class EmberAdapter extends Adapter {
                 console.log(`[INIT TC] No valid backup found.`);
                 action = NetworkInitAction.FORM_CONFIG;
             }
-        } else {
-            action = NetworkInitAction.DONE;// just to be clear
         }
 
         //---- from here on, we assume everything is in place for whatever decision was taken above
@@ -2591,6 +2589,51 @@ export class EmberAdapter extends Adapter {
         return this.sendZDORequestBuffer(target, PERMIT_JOINING_REQUEST, options);
     }
 
+    /**
+     * ZDO 
+     * 
+     * @see NWK_UPDATE_REQUEST
+     * 
+     * @param target 
+     * @param scanChannels uint8_t[]
+     * @param duration uint8_t 
+     * @param count uint8_t
+     * @param manager 
+     */
+    private async emberNetworkUpdateRequest(target: EmberNodeId, scanChannels: number[], duration: number, count: number | null,
+        manager: EmberNodeId | null, options: EmberApsOption): Promise<[EmberStatus, apsFrame: EmberApsFrame, messageTag: number]> {
+        this.zdoRequestBuffalo.setPosition(ZDO_MESSAGE_OVERHEAD);
+
+        this.zdoRequestBuffalo.writeUInt32(scanChannels.reduce((a, c) => a + (1 << c), 0));// to uint32_t
+        this.zdoRequestBuffalo.writeUInt8(duration);
+
+        if (count != null) {
+            this.zdoRequestBuffalo.writeUInt8(count);
+        }
+
+        if (manager != null) {
+            this.zdoRequestBuffalo.writeUInt16(manager);
+        }
+
+        debug(`~~~> [ZDO NWK_UPDATE_REQUEST target=${target} scanChannels=${scanChannels} duration=${duration} count=${count} manager=${manager}]`);
+        return this.sendZDORequestBuffer(target, NWK_UPDATE_REQUEST, options);
+    }
+
+    private async emberScanChannelsRequest(target: EmberNodeId, scanChannels: number[], duration: number, count: number, options: EmberApsOption):
+        Promise<[EmberStatus, apsFrame: EmberApsFrame, messageTag: number]> {
+        return this.emberNetworkUpdateRequest(target, scanChannels, duration, count, null, options);
+    }
+
+    private async emberChannelChangeRequest(target: EmberNodeId, channel: number, options: EmberApsOption):
+        Promise<[EmberStatus, apsFrame: EmberApsFrame, messageTag: number]> {
+        return this.emberNetworkUpdateRequest(target, [channel], 0xFE, null, null, options);
+    }
+
+    private async emberSetActiveChannelsAndNwkManagerIdRequest(target: EmberNodeId, scanChannels: number[], manager: EmberNodeId,
+        options: EmberApsOption): Promise<[EmberStatus, apsFrame: EmberApsFrame, messageTag: number]> {
+        return this.emberNetworkUpdateRequest(target, scanChannels, 0xFF, null, manager, options);
+    }
+
     //---- END Ember ZDO
 
     //-- START Adapter implementation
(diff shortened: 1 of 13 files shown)


src/controller/controller.ts
```diff
@@ -184,6 +184,18 @@ class Controller extends events.EventEmitter {
             databaseCoordinator.changeIeeeAddress(coordinator.ieeeAddr);
         }
 
+        if (startResult === 'resumed') {
+            const netParams = (await this.getNetworkParameters());
+
+            if (this.options.network.channelList[0] !== netParams.channel) {
+                debug.log(`Scheduling switch channel broadcast in 3min.`);
```
Koenkk, line 191: Why do we wait 3 minutes before doing this? (why not e.g. 10 secs)

src/controller/controller.ts
```diff
@@ -184,6 +184,18 @@ class Controller extends events.EventEmitter {
...
+                // NOTE: stacks that won't support channel switching will not report as 'resumed', they'll have 'reset'/'restored'
```
Koenkk, line 192: Is this true? E.g. the deconz adapter also returns `resumed`

https://github.com/Koenkk/zigbee-herdsman/blob/85447fbc839fd5d0b42fc7708c2b61486368b3f8/src/adapter/deconz/adapter/deconzAdapter.ts#L98

src/controller/controller.ts
Nerivec: I figured it was better to let things settle after init (to ensure the broadcast reached as many devices as possible). Since that worked exceptionally well on first try... I kept the 3 minutes after that, without looking back 😝
Probably should be adjusted after more tests. _Might even need a per-stack value if the post-init "ideal state" time isn't quite the same..._

src/controller/controller.ts
Nerivec: I might have assumed too much here. Since switching without re-pair was not supported, I assumed the combo `'resumed' + different channel` would not happen currently. I'll double check.

Nerivec: - deconz:
  - https://github.com/Nerivec/zigbee-herdsman/blob/channel-switching/src/controller/controller.ts#L188
  - https://github.com/Nerivec/zigbee-herdsman/blob/channel-switching/src/adapter/deconz/adapter/deconzAdapter.ts#L891
  - Not sure why all that logic is in `getNetworkParameters`, but seems that will short-circuit the channel mismatch logic in Controller. _Although, that `getNetworkParameters` has a variable scope/re-declaration problem?_
- zigate
  - https://github.com/Nerivec/zigbee-herdsman/blob/channel-switching/src/adapter/zigate/adapter/zigateAdapter.ts#L654
  - Should align channel before reaching the end of start, if I'm not mistaken?
- ezsp
  - https://github.com/Nerivec/zigbee-herdsman/blob/channel-switching/src/adapter/ezsp/driver/driver.ts#L294
  - Will result in 'reset'/'restored'.

src/controller/controller.ts
Nerivec: Lowered to 60sec:
- Triggering immediately on start (no setTimeout) causes collisions with permit join stuff, and post-init pings/broadcasts, creating unnecessary errors (although it still works fine, it's not great for user-feedback).
- It takes around 15 seconds for post-init stuff to be over on my very small test network => probably should allocate 30 minimum, and counting other stacks may need more time => 60... nice round number... 😅
- Tested half a dozen times with 60sec, consistent results every time.

Koenkk: 60 is fine then!

src/controller/controller.ts
Koenkk: I'm wondering if we can do something more simple here, I was thinking about adding a method to the adapter: `supportChangeChange()`

src/controller/controller.ts
```diff
@@ -414,6 +426,15 @@ class Controller extends events.EventEmitter {
         return Group.create(groupID);
     }
 
+    /**
+     * Broadcast a network-wide channel change.
+     */
+    public async switchChannel(newChannel: number): Promise<void> {
+        await this.adapter.switchChannel(newChannel);
+
+        this.networkParametersCached = null;// invalidate cache
```
Nerivec, line 435: Would be better if we could use an event triggered by the adapter to invalidate the cache at the proper time, but it depends on if all stacks (that will implement channel switching) are capable of reporting when the channel actually changed on the coordinator..?

src/controller/controller.ts
Nerivec: Added. Can probably be removed once all stacks confirm their logic in regards to this.

Nerivec:
I'm out of ideas to make that coverage happy, it doesn't seem to like the `spyOn`, but without it, `setTimeout` breaks the tests 😢

@Koenkk I set zStack & its test to `false` for support for now. Please adjust when support is done for zStack.

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -522,6 +523,12 @@ export class EmberAdapter extends Adapter {
             console.log(`[STACK STATUS] Network closed.`);
             break;
         }
+        case EmberStatus.CHANNEL_CHANGED: {
+            // invalidate cache
+            this.networkCache.parameters.radioChannel = INVALID_RADIO_CHANNEL;
+            console.log(`[STACK STATUS] Channel changed.`);
```
Koenkk, line 529:
```suggestion
            debug(`[STACK STATUS] Channel changed.`);
```

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -2812,6 +2861,37 @@ export class EmberAdapter extends Adapter {
...
+
+                    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+                    const [status, apsFrame, messageTag] = (await this.emberChannelChangeRequest(
+                        EMBER_SLEEPY_BROADCAST_ADDRESS,
+                        newChannel,
+                        DEFAULT_APS_OPTIONS,
+                    ));
+
+                    if (status !== EmberStatus.SUCCESS) {
+                        console.error(`[ZDO] Failed broadcast channel change to "${newChannel}" with status=${EmberStatus[status]}.`);
```
Koenkk, line 2883: Would it make sense to `reject` in this case?

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -918,6 +918,27 @@ class ZStackAdapter extends Adapter {
...
+    public async switchChannel(newChannel: number): Promise<void> {
+        return this.queue.execute<void>(async () => {
+            this.checkInterpanLock();
+
+            const payload = {
+                dstaddr: 0xFFFF,// broadcast with sleepy
+                dstaddrmode: AddressMode.ADDR_BROADCAST,
+                channelmask: [newChannel].reduce((a, c) => a + (1 << c), 0),
+                scanduration: 0xFE,/*switch channel*/
+                // scancount: null,// TODO: what's "not present" here?
```
Koenkk, line 934: Note to self; we can use the values from https://github.com/zigpy/zigpy-znp/blob/e3c9fcfdd249e089c578f99c4ed22159e9325c42/zigpy_znp/zigbee/application.py#L369

Nerivec: See 2.4.3.3.9.2 of [spec](https://csa-iot.org/wp-content/uploads/2023/04/05-3474-23-csg-zigbee-specification-compressed.pdf) for the flow on reception.

Nerivec:
Removed setTimeout (makes the tests/coverage happy too!). Ran immediately post NCP init, and blocking handled at adapter level as needed/available to stack.

`ember`: use stack status event to detect change over (async frames can start to come in within a few ms after that).

Basically, just needed to move the logic up in the scope; it's much simpler and just works 😉

Hedda:
> Adds base for supporting channel switching without having to re-pair devices.

FYI, [zigpy (for ZHA)](https://github.com/zigpy/zigpy) implemented a similar channel switching feature a while ago and a tip from that implemention is that they at least for z-stack/znp implemented resending the channel change broadcast command a few times with various delay before the Zigbee Coordinator actually switches channel, as that way you give devices that did not recieve the message the first time for whatever a chance a few chances recieve that message 

https://github.com/zigpy/zigpy/pull/1194

Also see related PR for ConBee 

https://github.com/zigpy/zigpy/pull/1194

Koenkk:
There are still some TODOs here which I will follow up on:
- Support for zstack
- Update Zigbee2MQTT documentation
- Add channel change events so we can log them in Z2M

PR #955, opened 2024-03-06
Add eWeLink Manufacturer Specific Cluster

jamesonuk:
There is a Sonoff manufacturer specific cluster which currently appears as 64529 when viewed in the UI
image

I have essentially gone through https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts looking for everything under cluster 0xfc11 and also incorporated the two new values from https://github.com/Koenkk/zigbee-herdsman-converters/pull/7130

<details>
  <summary>Converter Definitions</summary>
```
binary({
  name: 'tamper',
  cluster: 0xFC11,
  attribute: {ID: 0x2000, type: 0x20},
  description: 'Tamper-proof status',
  valueOn: [true, 0x01],
  valueOff: [false, 0x00],
  zigbeeCommandOptions: {manufacturerCode: 0x1286},
  access: 'STATE_GET',
}),
enumLookup({
  name: 'illumination',
  lookup: {'dim': 0, 'bright': 1},
  cluster: 0xFC11,
  attribute: {ID: 0x2001, type: 0x20},
  zigbeeCommandOptions: {manufacturerCode: 0x1286},
  description: 'Only updated when occupancy is detected',
  access: 'STATE',
}),
binary({
  name: 'child_lock',
  cluster: 0xFC11,
  attribute: {ID: 0x0000, type: 0x10},
  description: 'Enables/disables physical input on the device',
  valueOn: ['LOCK', 0x01],
  valueOff: ['UNLOCK', 0x00],
}),
binary({
  name: 'open_window',
  cluster: 0xFC11,
  attribute: {ID: 0x6000, type: 0x10},
  description: 'Automatically turns off the radiator when local temperature drops by more than 1.5°C in 4.5 minutes.',
  valueOn: ['ON', 0x01],
  valueOff: ['OFF', 0x00],
}),
numeric({
  name: 'frost_protection_temperature',
  cluster: 0xFC11,
  attribute: {ID: 0x6002, type: 0x29},
  description: 'Minimum temperature at which to automatically turn on the radiator, ' +
'if system mode is off, to prevent pipes freezing.',
  valueMin: 4.0,
  valueMax: 35.0,
  valueStep: 0.5,
  unit: '°C',
  scale: 100,
}),
numeric({
  name: 'idle_steps',
  cluster: 0xFC11,
  attribute: {ID: 0x6003, type: 0x21},
  description: 'Number of steps used for calibration (no-load steps)',
  access: 'STATE_GET',
}),
[...]

PR #983, opened 2024-03-19
Added Sonoff manufacturer's Cluster attribute command definition.

liubotongs:
Added a definition for CommandID:0x01 in the Sonoff manufacturer's Cluster, which is used to set the point function of ZBMicro devices. We will be using zigbee-herdsman-converters.


Nerivec:
Re https://github.com/Koenkk/zigbee-herdsman/pull/955

liubotongs:
So do we need to change "manuSpecificSonoff" to "manuSpecificEwelink"?

Nerivec:
The same cluster ID is used by multiple manufacturers. This needs the cluster injection feature before it can be added; which is still a work in progress.

liubotongs:
Thank you for your reply. When can we finish the work?

Koenkk:
Implementing it in https://github.com/Koenkk/zigbee-herdsman/pull/1019

Koenkk:
Fixed in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7432

PR #986, opened 2024-03-19
[ember] Removed `zigbeed` detection for stack config until firmware supported.

Nerivec:
Removed `zigbeed` detection (for customized stack config) based on Socket path detection. Multiprotocol addon being unsupported with 4.4.x/7.4.x, this currently has no impact. This will allow use of Socket without forcing `zigbeed` stack config in the meantime.

Cleaned up ZDO status logging.
Added parsing for ZDO `PARENT_ANNOUNCE_RESPONSE` (not supported upstream, but useful in logs).

CC: @MnM001 (_please open a new issue if you encounter trouble with that new adapter, or find me in Z2M Discord_)


Koenkk:
Thanks!

PR #989, opened 2024-03-21
feat: Improve logging

Koenkk:
Credits to @Nerivec


Nerivec:
For base classes, tried to follow as below:
- Anything that has the potential to spam a bit => `debug`
- Error/failure => `error`
- Possible error or lack of expected response => `warning`
- Unsupported/unhandled => `warning`
- Anything that will alter or has altered the network in some way (few but meaningful occurrences) => `info`
- Rest => `debug`

src/utils/logger.ts
```diff
@@ -0,0 +1,24 @@
+export interface Logger {
+    debug: (message: string, service?: string) => void;
+    info: (message: string, service?: string) => void;
+    warning: (message: string, service?: string) => void;
+    error: (message: string | Error, service?: string) => void;
+    child: (options?: object) => Omit<Logger, 'child'>;
+}
+
+export let logger: Logger = {
+    debug: (message) => console.debug(message),
```
Koenkk, line 10: @Nerivec what do you think about keeping DEBUG here when no logger is set?

src/controller/controller.ts
```diff
@@ -444,19 +436,19 @@ class Controller extends events.EventEmitter {
     }
 
     private onNetworkAddress(payload: AdapterEvents.NetworkAddressPayload): void {
-        debug.log(`Network address '${payload.ieeeAddr}'`);
+        logger.debug(`Network address '${payload.ieeeAddr}'`, NS);
         const device = Device.byIeeeAddr(payload.ieeeAddr);
 
         if (!device) {
-            debug.log(`Network address is from unknown device '${payload.ieeeAddr}'`);
+            logger.debug(`Network address is from unknown device '${payload.ieeeAddr}'`, NS);
```
deviantintegral, line 443: Could we change this to `logger.warn`? I can't see how this would happen in a well-functioning system and network. Wouldn't it mean that somehow a device has the proper network keys, but isn't recognized?

I was thinking of doing this over at https://github.com/Koenkk/zigbee-herdsman/pull/992 but then saw this much more comprehensive PR.

deviantintegral (commented):
This is great. In particular, consistently injecting the logger will make it much easier to alter logging for specific messages (such as bumping one from debug to warn) to simplify output.

One comment and idea, but I haven't actually reviewed the rest of the code.

src/controller/controller.ts
Nerivec: If I'm not mistaken, colliding networks (configs) can trigger a boatload of these "device is unknown to Z2M, but we saw a message from it" events. That's a potential source of spam, and even though this particular event is unlikely to be triggered very often (for now, until associated ZDO request supported globally), in my opinion, it wouldn't make sense to have just this one as `warning`, and the other similar events as `debug`.

PS: If we really need a `warning` for device lookup, it would probably be better to put it directly in the lookup functions; it cleans up the code (several places check for unknown/deleted and log it), and allows to handle the `includeDeleted` behavior, in regards to logging, internally.

PR #978, opened 2024-03-16
Description of additional discover commands.

kirovilya:
Allow general commands to be sent through the endpoint.

Later add general device converters and send these commands to the device from the developer console or via MQTT.
This is for developers of new converters so that they can find out the attributes of the clusters.
Unfortunately, devices do not report all of their attributes and commands. Some don't report anything.

For example:
```
zigbee2mqtt/0xa4c138b8c414612e/1/set {"discover":{"cluster": 6, "options":{}}}

Received 'zcl' data '{"frame":{"Header":{"frameControl":{"frameType":0,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":3,"manufacturerCode":null,"commandIdentifier":13},"Payload":{"discComplete":1,"attrInfos":[{"attrId":65533,"dataType":33},{"attrId":16385,"dataType":33},{"attrId":16386,"dataType":33},{"attrId":32769,"dataType":48},{"attrId":32770,"dataType":48},{"attrId":20480,"dataType":48}]},"Command":{"ID":13,"name":"discoverRsp","parameters":[{"name":"attrId","type":33},{"name":"dataType","type":32}]}},"address":41775,"endpoint":1,"linkquality":255,"groupID":0,"wasBroadcast":false,"destinationEndpoint":1}'


zigbee2mqtt/0xa4c138b8c414612e/2/set {"discoverExt":{"cluster": "manuSpecificTuya_3", "options":{}}}

Received 'zcl' data '{"frame":{"Header":{"frameControl":{"frameType":0,"manufacturerSpecific":false,"direction":1,"disableDefaultResponse":true,"reservedBits":0},"transactionSequenceNumber":17,"manufacturerCode":null,"commandIdentifier":22},"Payload":{"discComplete":1,"attrInfos":[{"attrId":53264,"dataType":48,"access":7},{"attrId":53280,"dataType":48,"access":7}]},"Command":{"ID":22,"name":"discoverExtRsp","parameters":[{"name":"attrId","type":33},{"name":"dataType","type":32},{"name":"access","type":32}]}},"address":41775,"endpoint":2,"linkquality":232,"groupID":0,"wasBroadcast":false,"destinationEndpoint":1}'
```

Issue #686, opened 2023-03-28
Unable To Send Commands To Clusters Which Share Same Cluster ID


danieledwardgeorgehitchcock:
There is an existing cluster in Herdsman `manuSpecificUbisysDeviceSetup` which has a cluster id  of `64512`

I created a new cluster `manuSpecificAssaDoorLock` which shares the same cluster id of `64512` or Hex `0xFC00`

When trying to send a command to `manuSpecificAssaDoorLock` using the cluster id (as the cluster name cannot be used as far as I am aware), I get an error response stating that the cluster `manuSpecificUbisysDeviceSetup` has no command `X`.

This presents an interesting problem, as these manufacturer specific clusters can have overlapping cluster and command id's that have completely different functions so, cannot be merged together.

Is there a way to control which cluster (i.e. by name) Herdsman (and Z2M) uses?

Error message:

```
Zigbee2MQTT:debug 2023-03-28 07:33:45: Received MQTT message on 'zigbee2mqtt/0x000d6f0011141333/1/set' with data '{"command":{"cluster":64512,"command":18,"payload":{}}}'
Zigbee2MQTT:debug 2023-03-28 07:33:45: Publishing 'set' 'command' to '0x000d6f0011141333'
Zigbee2MQTT:error 2023-03-28 07:33:45: Publish 'set' 'command' to '0x000d6f0011141333' failed: 'Error: Cluster 'manuSpecificUbisysDeviceSetup' has no command '18''
Zigbee2MQTT:debug 2023-03-28 07:33:46: Error: Cluster 'manuSpecificUbisysDeviceSetup' has no command '18'
    at Object.getCommand (/app/node_modules/zigbee-herdsman/src/zcl/utils.ts:114:19)
    at Endpoint.command (/app/node_modules/zigbee-herdsman/src/controller/model/endpoint.ts:781:33)
    at Object.convertSet (/app/node_modules/zigbee-herdsman-converters/converters/toZigbee.js:54:26)
    at Publish.onMQTTMessage (/app/lib/extension/publish.ts:247:52)
    at EventEmitter.emit (node:events:525:35)
    at EventBus.emitMQTTMessage (/app/lib/eventBus.ts:109:22)
    at MQTT.onMessage (/app/lib/mqtt.ts:140:27)
    at WebSocket.<anonymous> (/app/lib/extension/frontend.ts:126:27)
    at WebSocket.emit (node:events:513:28)
[...]

PR #999, opened 2024-04-01
[Fix for #978] Cluster defined with a key are not found anymore

LaurentChardin:
After #978, the `Zcl.ZclFrame.create` is using `cluster.name` instead of `cluster.ID`. 

However, the cluster key is actually expected from the method signature.

This has as the following consequence: consecutive calls to `getCluster` will change the key from a number to a string, and failing on line 61 of utils.ts, while the key was still valid.

This is needed when testing clusters that are not defined (yet) in _cluster.ts_.

Fix: use cluster.ID instead of cluster.name in order to support custom clusters.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -749,7 +749,7 @@ class Endpoint extends Entity {
         const frame = Zcl.ZclFrame.create(
             frameType, options.direction, options.disableDefaultResponse,
             options.manufacturerCode, options.transactionSequenceNumber ?? ZclTransactionSequenceNumber.next(),
-            command.name, cluster.name, payload, options.reservedBits
+            command.name, cluster.ID, payload, options.reservedBits
         );
 
         const log = `ZCL command ${this.deviceIeeeAddress}/${this.ID} ` +


LaurentChardin:
So i see the test failing here.

To put some context, i was trying this change because of the extend in my device definition:
```
            binary({
                name: 'eOnTimer',
                cluster: 64518,
                attribute: {ID: 0x0003, type: 0x10},
                description: 'Enable (0x01) / Disable (0x00) use of onTimer.',
                valueOn: ['ON', 0x01],
                valueOff: ['OFF', 0x00],
                entityCategory: 'config'
            }),
```
Cluster 64518 is not defined in cluster.ts but zclCommand was failing because 64518 becomes '64518'.
By using cluster.id instead of cluster.name in the Zcl.ZclFrame.create call, the call is executed properly.

But it seems the use of cluster.name was intentional in order to detect Cluster ID collision by forcing a lookup in the main referential of cluster.ts, assuming cluster.name is unique and cluster.ID is not.

LaurentChardin:
Ok found it back to https://github.com/Koenkk/zigbee-herdsman/issues/686 

I understand the context, but then it kinda breaks the *on-the-fly* cluster definition. 

I will just go back and adjust cluster.ts for the moment then. And i suspect other examples i have seen in ZHC will have to be changed as well.

PR #1004, opened 2024-04-03
Update last seen on network address events

deviantintegral:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/22043.

This wasn't directly caused by https://github.com/Koenkk/zigbee-herdsman/pull/992, but exposed by it because it caused the network address event to be emitted far more often. While the current code is emitting the event, it's not actually updating the last seen timestamp.

First a test, then I will push the fix to prove it actually is!

The whitespace errors could be pulled out to a separate PR - they get cleaned up automatically every time I save in my IDE, and also throw warnings when encountered during git checkout / stash operations.

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -444,6 +444,7 @@ class Controller extends events.EventEmitter {
             return;
         }
 
+        device.updateLastSeen();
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #1006, opened 2024-04-03
Ember: Fixed logging levels.

Nerivec:
- Cleanup unused logs.
- Changed a few levels to better match.
- Always wrap `enqueue` with `Promise` and `reject` instead of just log on error.


Koenkk:
Thanks!

PR #1008, opened 2024-04-04
Add `manuSpecificBosch11` cluster

burmistrzak:
- Defines cluster No. `64684`, used by some Bosch devices, as `manuSpecificBosch11`
- Includes a bit of housekeeping for other Bosch clusters


Koenkk:
Thanks!

PR #1010, opened 2024-04-06
Ember: Code optimizations

Nerivec:
- Use recursive `setTimeout` instead of `setInterval`
- Remove unnecessary functions
- Always stop queue dispatching on `EmberAdapter.stop()`
- Clear all timers on EZSP init
- Add `info` log for counter rollover to match counter log.


Koenkk:
Thanks!

PR #1011, opened 2024-04-06
fix: Move `ZclFrame.fromBuffer()` out of adapter code

Koenkk:
Move the `ZclFrame.fromBuffer()` out of adapter code to prepare for custom cluster definitions as discussed in https://github.com/Koenkk/zigbee-herdsman/pull/971


src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -681,9 +676,13 @@ export class EmberAdapter extends Adapter {
             gpdHeader.writeUInt8(gpdCommandId, 13);// commandID
             gpdHeader.writeUInt8(gpdCommandPayload.length, 14);// payloadSize
 
-            const gpFrame = ZclFrame.fromBuffer(Cluster.greenPower.ID, Buffer.concat([gpdHeader, gpdCommandPayload]));
+            const data = Buffer.concat([gpdHeader, gpdCommandPayload]);
+            const header = ZclHeader.fromBuffer(data);
+            const gpFrame = ZclFrame.fromBuffer(Cluster.greenPower.ID, header, data);
```
Koenkk, line 681: @Nerivec this is the only place where I couldn't remove `ZclFrame.fromBuffer`, it's not really a problem since this cluster is always greenPower here (very unlikely that we will require custom cluster definitions for greenPower), but still. Do you see a way to do ` wasBroadcast: (gpFrame.Payload.gppNwkAddr != null) ? false : true,` without parsing the frame?

Nerivec: According to [this](https://github.com/Koenkk/zigbee-herdsman/blob/cd6906cfc37d74a00e7f7bd7c1745f566441dda8/src/controller/greenPower.ts#L76), it should be fine to set `wasBroadcast: true`.

src/adapter/events.ts
```diff
@@ -34,18 +33,10 @@ type DeviceLeavePayload = {
...
-    linkquality: number;
-    groupID: number;
-    wasBroadcast: boolean;
-    destinationEndpoint: number;
-}
-
-interface RawDataPayload {
     clusterID: number;
     address: number | string;
+    zclFrameHeader: ZclHeader | undefined;
```
Nerivec, line 38: Can we just name this `header`? Makes for shorter variables everywhere, and quick one-word assignment to objects.
I'd even say we can rename `ZclDataPayload` to just `ZclPayload` while we're at it.

```typescript
const header: ZclHeader;

try {
    header = ...;
} ...

const payload: ZclPayload = {
    header,
    ...
};
```

src/zcl/zclHeader.ts
```diff
@@ -1,10 +1,51 @@
...
+        this.transactionSequenceNumber = transactionSequenceNumber;
+        this.commandIdentifier = commandIdentifier;
+    }
+
+    public getLength(): number {
+        // Returns the amount of bytes of this header
+        return 3 + (this.manufacturerCode === null ? 0 : 2);
+    }
+
+    public static fromBuffer(buffer: Buffer): ZclHeader {
```
Nerivec, line 24: Can this return `ZclHeader | undefined` instead of throwing? So in adapters code:

```typescript
const payload: ZclDataPayload: {
    header: ZclHeader.fromBuffer(messageContents),
    ...
};
```

Controller can log if `undefined`.

src/adapter/events.ts
Koenkk: Definitely, will do it once all tests are fixed.

src/zcl/zclHeader.ts
Koenkk: Good idea, will also do this once all tests are fixed.

Nerivec:
A few more suggestions...

- Moving [this function](https://github.com/Koenkk/zigbee-herdsman/blob/443bee1d5ecca21b0312ae3a864a872fa6158259/src/zcl/zclFrame.ts#L80) to `ZclHeader` as `write()` (or something like that) for consistency. _Also, `frameControl` doesn't need a variable in there, it can just be passed directly to `buffalo.writeUInt8`._

- Have something like this in `ZclHeader`:

```typescript
    /** Returns the amount of bytes used by this header */
    get length(): number {
        return 3 + (this.manufacturerCode === null ? 0 : 2);
    }

    get isGlobal(): boolean {
        return this.frameControl.frameType === FrameType.GLOBAL;
    }

    get isSpecific(): boolean {
        return this.frameControl.frameType === FrameType.SPECIFIC;
    }
```

- In `ZclFrame`, `this.Header.commandIdentifier` is a [duplicate](https://github.com/Koenkk/zigbee-herdsman/blob/443bee1d5ecca21b0312ae3a864a872fa6158259/src/zcl/zclFrame.ts#L59) of `this.Command.ID`. Also, `frame.getCommand()` seems to be called often enough just to get the ID (`frame.getCommand().ID`). Probably some optimizations needed around this.

src/adapter/events.ts
Koenkk: Done!

src/zcl/zclHeader.ts
Koenkk: Checking this now, just `throw new Error("ZclHeader length is lower than minimal length");`  -> `return undefined` is not enough. It can also go wrong at other lines in this function. Which mean we would have to do:

```js
try {
 .. all function code
} catch (error) {
  return undefined
}
```

which doesn't feel right to me.

Koenkk:
All done! Let me know if this is OK now.

Nerivec (approved):
Looks good!

PR #1019, opened 2024-04-14
Support custom clusters

Koenkk:
Requires https://github.com/Koenkk/zigbee-herdsman/pull/1011, implements https://github.com/Koenkk/zigbee-herdsman/pull/971

PR #1021, opened 2024-04-19
Add Airzone Manufacturer ID

AlexisPolegato:
Add Airzone Manufacturer ID
Add manufacturer specific attributes from NodOn in Window Covering Cluster


sjorge:
Those custom attribute should probably wait until after https://github.com/Koenkk/zigbee-herdsman/pull/1019 so we don't have to move them again later.

Nerivec:
Is this an officially attributed (by CSA) manufacturer code? If not, it should be at the bottom, with `CUSTOM_` prefix to follow existing pattern.

Koenkk:
Custom clusters can now be defined in zigbee-herdsman-converters, example: https://github.com/Koenkk/zigbee-herdsman-converters/commit/99613a7a522f916cc3a0bfc2b52ac26363c07b3a , please make a PR for that instead.

PR #1028, opened 2024-04-23
Improve broadcasting via `sendZclFrameToAll`

burmistrzak:
Just adds an optional parameter to the `sendZclFrameToAll` method so [sleepy end devices](https://community.silabs.com/s/article/what-are-the-different-functions-of-the-3-broadcast-addresses-in-emberznet3-1-x?language=en_US) (i.e. `0xFFFF`) can be included in broadcast messages. 😊

Seems to be required for [this](https://github.com/Koenkk/zigbee-herdsman-converters/pull/7427) to work properly.


Nerivec:
I'd say no boolean, just a variable for the address:
```typescript
/** Broadcast to all routers. */
export const BROADCAST_ADDRESS = 0xFFFC;
/** Broadcast to all non-sleepy devices. */
export const BROADCAST_RX_ON_WHEN_IDLE_ADDRESS = 0xFFFD;
/** Broadcast to all devices, including sleepy end devices. */
export const BROADCAST_SLEEPY_ADDRESS = 0xFFFF;
```

These should be with Zigbee spec types, usable everywhere.

And the adapter function becomes
```typescript
public abstract sendZclFrameToAll(endpoint: number, zclFrame: ZclFrame, sourceEndpoint: number, address: number = BROADCAST_ADDRESS): Promise<void>;
```


Note:

> ZigBee specifies three different broadcast addresses that reach different collections of nodes.  Broadcasts are normally sent only  to routers.  Broadcasts can also be forwarded to end devices, either  all of them or only those that do not sleep. Broadcasting to end devices is both significantly more resource-intensive and significantly less reliable than broadcasting to routers.

Nerivec:
Only 6 use cases of `sendZclFrameToAll` currently, all are for GP, so not complicated to refactor. Better to use the proper default broadcast (FFFC) and change as needed to FFFD or FFFF by specifying it.

Nerivec:
@Koenkk Not sure what convention we want to use for naming, but since we're gonna introduce the zdo spec too, I figure we can put global stuff at the root and zcl/zdo as subfolders. Feel free to rename! (I didn't move the content of the zcl folder here, since that's a bigger refactor. I also didn't add the million or so consts that can be introduced to remove magic numbers throughout the code 😁)

I added a test that mirrors the requirements of the Bosch in https://github.com/Koenkk/zigbee-herdsman-converters/pull/7427. I think it makes more sense that way?
ZHC can just do
```typescript
 await meta.device.triggerBroadcast(255, 1, BroadcastAddress.SLEEPY, Zcl.Direction.CLIENT_TO_SERVER, 'boschSmokeDetectorSiren', Zcl.Clusters.ssIasZone.ID, {data: index});
```

Nerivec:
Waiting for Koenkk's input on naming & co, so we can export this properly. The spec folder structure will affect a lot of stuff, better to start with a solid base.

src/controller/model/device.ts
```diff
@@ -828,6 +829,23 @@ class Device extends Entity {
         }
         this._customClusters[name] = cluster;
     }
+
+    public async triggerBroadcast(endpoint: number, sourceEndpoint: number, destination: BroadcastAddress, direction: Zcl.Direction,
```
Koenkk, line 833:
```suggestion
    public async broadcastCommand(endpoint: number, sourceEndpoint: number, destination: BroadcastAddress, direction: Zcl.Direction,
```

To align with the naming in `endpoint` (e.g. `endpoint.command`)

Nerivec:
I figure, might as well keep entirely in line with the rest, so `zclCommandToAll` seems the best option?

@Koenkk Can you confirm if `this.customClusters` should be passed in this scenario (to `getCluster` and to `ZclFrame.create`)? Same question for `this.manufacturerID` to `getCluster` (that's what endpoint is using)? Seems a bit conflicting between using `device` (`this`) and using `options`.

Nerivec:
The move will be in a separate PR. It was just the right time to create the base for it since we needed a couple of "global zigbee" consts for this.

Koenkk:
@burmistrzak let's do this in a separate PR such that we can also merge https://github.com/Koenkk/zigbee-herdsman-converters/pull/7427 before the new release.

PR #1030, opened 2024-04-24
Parsing of the genOnOff.tuyaAction data parameter as a buffer

kirovilya:
image
image
The same proprietary command in a standard cluster has a different payload size...
Therefore, let's add one more parameter to the command tuyaAction and process it as a buffer.


Koenkk:
Thanks!

PR #1031, opened 2024-04-25
fix(ZCL): update msOccupancySensing definition

mrskycriper:
Updates `msOccupancySensing` cluster definition according to the chapter 4.8 of ZCL R8 spec:

- Add missing attributes defined by the ZCL
  - `occupancySensorTypeBitmap`
  - `contactOToUDelay`
  - `contactUToODelay`
  - `contactUToOThreshold`
- Use hex attribute ids to avoid confusion


Koenkk:
Thanks!

PR #1055, opened 2024-05-09
Add the ClusterID of the eWelink Manufacturer.

liubotongs:
Currently, devices in sonoff.ts with clusterId=0xFC11 private properties or command functions are not available, so we choose to add eWeLinkManufacturer to cluster.ts.


Koenkk:
Please use the new custom cluster API, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/21e1635c087add5ba25214278e3a0f3b8526725e/src/devices/sonoff.ts#L809

liubotongs:
Thank you very much. As you mentioned, deviceAddCustomCluster can help us. However, when I test the attribute with ManufacturerCode, Z2M can read it. However, the error message "failed (Write for 'BUFFER' not available)" is displayed during the write.

liubotongs:
It is also observed that in zigbeeherdsmanconverters sonoff.ts, childLock: {ID: 0x0000, type: Zcl.DataType.BOOLEAN}, but childLock: {ID: 0x0000, type: zigbee_herdsman_1.Zcl.DataType.boolean} in zigbee2mqtt sonoff.js. BOOLEAN and boolean case is different.

Koenkk:
But I guess with the change in this PR it's the same issue?

Issue #1059, opened 2024-05-13
[Cluster Type] Align cluster definition for 'name' attribute


LaurentChardin:
Assuming that the ClusterDefinition interface introduce with the CustomClusters PR could be use in a different context (which is the case for example when using the type definition in zigbee2mqtt-frontend), the name attribute is missing, while it is being used in the Cluster interface.

Proposing to add the name attribute as optional.

LaurentChardin:
Closing: see https://github.com/Koenkk/zigbee-herdsman/pull/1060

PR #1060, opened 2024-05-13
Adding name attribute to ClusteDefinition interface #1059

LaurentChardin:
Aligning ClusterDefinition interface with Cluster interface : enables reusing the type definition is another project (where name is present in the payload but not in the interface definition. And Cluster interface is too wide for just a datatype definition)

### DIFF
--- a/src/zspec/zcl/definition/tstype.ts
+++ b/src/zspec/zcl/definition/tstype.ts
@@ -95,6 +95,7 @@ export interface Cluster {
 
 export interface ClusterDefinition {
     ID: number;
+    name?: string;
     manufacturerCode?: number;
     attributes: Readonly<Record<string, Readonly<AttributeDefinition>>>;
     commands: Readonly<Record<string, Readonly<CommandDefinition>>>;


LaurentChardin:
By the way, the test are failing on the master branch too.

src/zspec/zcl/definition/tstype.ts
```diff
@@ -95,6 +95,7 @@ export interface Cluster {
 
 export interface ClusterDefinition {
     ID: number;
+    name?: string;
```
Koenkk, line 98:
```suggestion
    name: string;
```

I propose to make this mandatory

Nerivec: Making that mandatory would enforce duplicating the name in code for all ZCL clusters (uses the object key to add the name at runtime at the moment). That's why `Cluster` has `name` but not `ClusterDefinition`.

LaurentChardin: The reason i didn't put it mandatory is because we will need to update the CustomCluster that are currently rolling out. In In the current implementation, the actual name of the cluster is used as a key of the object, rather than an attribute of it. 

Which i guess is used to enforced some form of primary key.

Example of current definition model:
```
const myClustersDefinition: {
    [s: string]: ClusterDefinition
} = {
    **manuSpecificCluster**: {
        ID: 0xFC01,
        manufacturerCode: Zcl.ManufacturerCode.MANUF,
        attributes: {},
        commands: {},
        commandsResponse: {},
},
etc ... 
```

Update : @Nerivec  answered at the same time :)

LaurentChardin: @Nerivec @Koenkk The reason of this change is because i was looking for an data-only representation of a Cluster object to use in the frontend UI.

- `Cluster` interface enforce implementing some methods
- `ClusterDefinition` interface is lacking the `name` attribute

Name optional was an ok-inbetween.

Nerivec: That also means frontend would remain dependent on the herdsman package since it would need the types import?

LaurentChardin: Yes.. now that @Koenkk has clarified what we do with https://github.com/Koenkk/zigbee2mqtt/pull/22583, i will make the necessary changes in order to remove the dependency completely.

LaurentChardin:
Closing this PR : as per https://github.com/nurikk/zigbee2mqtt-frontend/pull/2019, the type definition has been copied directly into the codebase.

PR #1070, opened 2024-05-20
fix: improve path handling on windows

tobiasdiez:
I got the following error on Windows.
```
Uncaught Error Error: EINVAL: invalid argument, lstat '\\.\COM4\'
    at realpathSync (fs:2659:25)
    at realpathSync (d:\Programming\zigbee2mqtt\node_modules\zigbee-herdsman\dist\utils\realpathSync.js:10:25)
    at eval (repl:1:26)
```

This is fixed by just resolving the path to the basename (COM4 in this case).

(Perhaps it's now a good idea to rename the method to `resolveDevicePath(path: string)` to clarify that it's mainly used for this purpose and not a generic "realPathSync".)

### DIFF
--- a/src/utils/realpathSync.ts
+++ b/src/utils/realpathSync.ts
@@ -1,9 +1,17 @@
 /* istanbul ignore file */
 import fs from 'fs';
+import { basename } from 'path';
 
-/* Only used for mocking purposes */
 function realpathSync(path: string): string {
-    return fs.realpathSync(path);
+    try {
+        return fs.realpathSync(path);
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


Koenkk:
Did you get this with autodetect or did you put `\\.\COM4\ in the `configuration.yaml` manually?

tobiasdiez:
I put it in following https://www.zigbee2mqtt.io/guide/installation/05_windows.html#configuring.

Koenkk:
Let's update the docs instead to use `COM4` instead of `\\.\COM4`

PR #1081, opened 2024-06-06
Ezsp: log failed message delivery

Nerivec:
`ezsp` has no implementation for  failed deliveries at all, they are just ignored.
This should at least make it easier for users going back and forth to identify network problems, vs driver problems in cases like https://github.com/Koenkk/zigbee2mqtt/issues/22453


Koenkk:
Thanks!

PR #1086, opened 2024-06-09
HIDEs the network key in the logs

ghoz:
We do it when calling Zigbee-herdsman from zigbee2mqtt, but not when receiving the config.

This has security implications and might require a communication as any user who shared their logs
also shared their network secret

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,6 @@
     "fast-deep-equal": "^3.1.3",
     "mixin-deep": "^2.0.1",
     "mz": "^2.7.0",
-    "object-assign-deep": "^0.4.0",
     "slip": "^1.0.2"
   },
   "deprecated": false,
@@ -51,7 +50,6 @@
     "@types/mz": "^2.7.8",
     "@types/nedb": "*",
     "@types/node": "*",
-    "@types/object-assign-deep": "^0.4.3",
     "@typescript-eslint/eslint-plugin": "*",
     "@typescript-eslint/parser": "*",
     "babel-jest": "*",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,7 @@
     "fast-deep-equal": "^3.1.3",
     "mixin-deep": "^2.0.1",
     "mz": "^2.7.0",
+    "object-assign-deep": "^0.4.0",
     "slip": "^1.0.2"
   },
   "deprecated": false,
@@ -50,6 +51,7 @@
     "@types/mz": "^2.7.8",
     "@types/nedb": "*",
     "@types/node": "*",
+    "@types/object-assign-deep": "^0.4.3",
     "@typescript-eslint/eslint-plugin": "*",
     "@typescript-eslint/parser": "*",
     "babel-jest": "*",
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -12,7 +12,6 @@ import Touchlink from './touchlink';
 import GreenPower from './greenPower';
 import {BackupUtils} from "../utils";
 import assert from 'assert';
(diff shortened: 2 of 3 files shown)


src/controller/controller.ts
```diff
@@ -123,7 +124,11 @@ class Controller extends events.EventEmitter {
 
         // Adapter (create and inject)
         this.adapter = await Adapter.create(this.options.network, this.options.serialPort, this.options.backupPath, this.options.adapter);
-        logger.debug(`Starting with options '${JSON.stringify(this.options)}'`, NS);
+
+        let optionsLog = objectAssignDeep({}, this.options, {network: {networkKey: 'HIDDEN'}});
+        logger.debug(`Starting with options '${JSON.stringify(optionsLog)}'`, NS);
```
Koenkk, line 129: To prevent adding a new dependency, can we just do this here?

```suggestion
        logger.debug(`Starting with options '${JSON.stringify(this.options).replace(this.options.network.networkKey, 'HIDDEN')}'`, NS);
```

Nerivec: Probably best to do that [here](https://github.com/Koenkk/zigbee2mqtt/blob/f18473017ebfcc65e9db1de3002b7eb21079805a/lib/zigbee.ts#L57) too, instead of messing with objects.

Nerivec: Probably this would be better as a feature of https://github.com/nurikk/zigbee2mqtt-frontend/issues/2011 "Download sanitized". It could go through security (network key, IEEE...) and privacy (friendly names...) checks, on an existing file, specifically for sharing. That way local logs could retain all values for local debugging.

Koenkk: But that doesn't apply to the log files in the log directory, I think it's good to keep it hidden in the logs as well.

Nerivec: It all comes down to the security of the machine in question, and the handling of the logs/config/backup (kept local vs shared/uploaded), since everything is kept in plain in the same folder.

If frontend provides a button to download any log (current or archived) with the sanitization logic applied when downloaded, it can provide any log "ready for sharing". With the option of manually accessing the files (directly), to see the plain versions.

ghoz: Logs are usually in a less safe place than the config, we definitively need to hide the secrets there...
Of course it does not preclude from having a download a sanitized version with anonymization

ghoz: > Probably best to do that [here](https://github.com/Koenkk/zigbee2mqtt/blob/f18473017ebfcc65e9db1de3002b7eb21079805a/lib/zigbee.ts#L57) too, instead of messing with objects.

That's the line I used, for consistency indeed ...

ghoz: > To prevent adding a new dependency, can we just do this here?

I was worried with  ``networkKey`` as an array, but it seems to work. I'm JSON.stringify() it rather than toString() just in case and array.toString and JSON stringify don't behave the same for some reason.
One other thing I used replaceAll so that it wouldn't  miss if you set your networkKey to an existing key  for some weird reason...  and instead make for a fun "guess the HIDDEN key" game.
So yeah it is technically more fragile and could still leak those type of weird keys

package.json
```diff
@@ -34,6 +34,7 @@
     "fast-deep-equal": "^3.1.3",
     "mixin-deep": "^2.0.1",
     "mz": "^2.7.0",
+    "object-assign-deep": "^0.4.0",
```
Koenkk, line 37: Could you remove the added dependencies here? After that this can be merged.

ghoz: Forgot to git add ...
Done

Koenkk:
Thanks!

PR #1088, opened 2024-06-13
Expose all commands

Koenkk:
This PR makes zigbee-herdsman expose all commands. Currently it is filtered through `CommandsLookup` which became messy already. With the introduction of custom clusters this filtering has become a problem, since custom commands never emit events. This PR removes the `CommandsLookup` entirely.

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -734,12 +734,8 @@ class Controller extends events.EventEmitter {
             } else {
                 /* istanbul ignore else */
                 if (frame.header.isSpecific) {
-                    if (Events.CommandsLookup[command.name]) {
-                        type = Events.CommandsLookup[command.name];
-                        data = frame.payload;
-                    } else {
-                        logger.debug(`Skipping command '${command.name}' because it is missing from the lookup`, NS);
-                    }
+                    type = `command${command.name.charAt(0).toUpperCase()}${command.name.slice(1)}`;
+                    data = frame.payload;
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/controller/controller.ts
```diff
@@ -734,12 +734,8 @@ class Controller extends events.EventEmitter {
             } else {
                 /* istanbul ignore else */
                 if (frame.header.isSpecific) {
-                    if (Events.CommandsLookup[command.name]) {
-                        type = Events.CommandsLookup[command.name];
-                        data = frame.payload;
-                    } else {
-                        logger.debug(`Skipping command '${command.name}' because it is missing from the lookup`, NS);
-                    }
+                    type = `command${command.name.charAt(0).toUpperCase()}${command.name.slice(1)}`;
```
Nerivec, line 737: What was the initial drive behind using `command${}` instead of just using `command.name`?
`MessagePayloadType` also probably needs an overhaul.

src/controller/model/endpoint.ts
```diff
@@ -285,8 +285,9 @@ class Endpoint extends Entity {
 
         // send without queueing if sendPolicy is 'immediate' or if the device has no timeout set
         if (request.sendPolicy === 'immediate' || !device.pendingRequestTimeout) {
+            logger.debug(logPrefix + `send ${frame.command.name} request immediately (sendPolicy=${options.sendPolicy})`, NS);
+
             if (device.pendingRequestTimeout > 0) {
```
Nerivec, line 290: Empty `if` statement.

src/controller/controller.ts
Koenkk: Because these are commands, in this way they never can conflict with e.g. `read` and `message`. For the `MessagePayloadType` we maybe should put `command` there and put the command name on another property, but that's for later.

src/controller/model/endpoint.ts
Koenkk: Whoops, I shouldn't have committed this.

src/controller/controller.ts
Nerivec: I guess we can do this until then, to atleast narrow the typing a bit:
```typescript
type MessagePayloadType = 'attributeReport' | 'readResponse' | 'raw' | 'read' | 'write' | `command${string}`;
```

Koenkk: Didnt know this was possible, cool!

PR #1089, opened 2024-06-14
Ember: workaround: auto-register unknown multicasts in coordinator

Nerivec:
Automatically register unknown multicast group IDs in coordinator multicast table to allow devices using this scheme for state updates to work without needing manual code change. Should take care of https://github.com/Koenkk/zigbee2mqtt/issues/22928
@LamerTex I don't have a device that does this, can you confirm this works once this is available in dev branch?

_Also a bit of housekeeping, use ZSpec instead of adapter-specific vars._


LamerTex:
Yes, on Monday I will be able to test it if it is available in dev branch. 
Thanks

Koenkk:
thanks!

FloJ82:
Thanks! Looking forward, having the same issue...

LamerTex:
@Nerivec I just tested it and I can confirm that in the dev branch, with this fix, the EGLO 99099 now works also with the ember driver, thanks! 

image

Nerivec:
Great. Thanks for the feedback!

PR #1090, opened 2024-06-14
feat: Add tuyaSetMinimumBrightness command

mlichvar:
This command is used by MiBoxer TR-C1ZR dimmer. It was observed with a Tuya gateway.


Koenkk:
Thanks!

PR #1096, opened 2024-06-25
Inovelli Adding P26 For VZM36

InovelliUSA:


src/zspec/zcl/definition/cluster.ts
```diff
@@ -5033,14 +5033,14 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
...
             activePowerReports: {ID: 0x0012, type: DataType.UINT8},
-
             periodicPowerAndEnergyReports: {ID: 0x0013, type: DataType.UINT16},
             activeEnergyReports: {ID: 0x0014, type: DataType.UINT16},
             powerType: {ID: 0x0015, type: DataType.BOOLEAN},
             switchType: {ID: 0x0016, type: DataType.UINT8},
             quickStartTime: {ID: 0x0017, type: DataType.UINT8},
             quickStartLevel: {ID: 0x0018, type: DataType.UINT8},
             higherOutputInNonNeutral: {ID: 0x0019, type: DataType.BOOLEAN},
+            leadingTrailingEdge: {ID: 0x001a, type: DataType.UINT8},
```
Koenkk, line 5043: Since custom device specific clusters are now supported, could you add the cluster there? Then we can remove the cluster definition here. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/3e56ca454b942b4780f08976e81de93982f89606/src/devices/sonoff.ts#L42

InovelliUSA: I will put it on my list to move them over to the device converter file. Thanks.

Koenkk: Great, then I will merge this one for now.

InovelliUSA: @Koenkk I am working on adding the custom cluster to the driver but I am not sure I am doing it correctly. Does this look right: https://github.com/InovelliUSA/zigbee-herdsman-converters/commit/589cc64e1003ae8fa48a4376bc40e65ddd3b1fb0

I have tested the js in my z2m docker container. I also deleted the Inovelli specific stuff from clusters.js and tested it. Everything still seems to work, but I am not sure if there is another way to test.

Koenkk: Looks good!

InovelliUSA: That is great. If I submit a PR for it to go to dev, is the next production version on 8/7? Since it is a change I haven't tried before I would like a couple weeks for people to be able to test it.

Koenkk: Next release is on 1 August

InovelliUSA: Great, I purposely missed that deadline so the change can be in dev for longer. I submitted the change and you approved it (thanks!). Do we need to remove the Inovelli code from this repo too? I'm assuming that if we make the change in this repo that both changes need to make it to production at the same time. It is especially important that this change will make it after this one: https://github.com/Koenkk/zigbee-herdsman-converters/pull/7822

Koenkk: Yes please, could you also submit a PR to remove the cluster from zigbee-herdsman?

InovelliUSA: Yep, I will do it.

PR #1098, opened 2024-06-29
Add calibration attributes for Niko 552-72301

Sennevds:
This adds the custom calibration attributes for Niko 552-72301 see https://github.com/Koenkk/zigbee2mqtt/issues/12755#issuecomment-2198035302


Koenkk:
Thanks!

PR #1110, opened 2024-07-11
feat: Support channel change for z-stack

Koenkk:
Support channel change for z-stack

TODO:
- [x] Zigbee2MQTT (https://github.com/Koenkk/zigbee2mqtt/pull/23333)
- [x] Test
- [x] Update docs (https://github.com/Koenkk/zigbee2mqtt.io/pull/2893)


src/controller/controller.ts
```diff
@@ -131,11 +131,20 @@ class Controller extends events.EventEmitter {
...
-        if (startResult === 'resumed' && (await this.adapter.supportsChangeChannel())) {
+        if (startResult === 'resumed') {
             const netParams = await this.getNetworkParameters();
+            const configuredChannel = this.options.network.channelList[0];
+            const adapterChannel = netParams.channel;
 
-            if (this.options.network.channelList[0] !== netParams.channel) {
-                await this.changeChannel();
+            if (configuredChannel != adapterChannel) {
+                if (this.adapter.supportsChangeChannel()) {
```
Nerivec, line 140: Maybe we can just use a throw in `changeChannel` if not supported instead of 2 functions?
This codepath is expected to not be reached if adapter does not support channel changing anyway (will have done the logic to change by whatever means before this is reached).
Old refs: https://github.com/Koenkk/zigbee-herdsman/pull/977#discussion_r1527181194

src/controller/controller.ts
```diff
@@ -438,9 +447,10 @@ class Controller extends events.EventEmitter {
     /**
      * Broadcast a network-wide channel change.
      */
-    private async changeChannel(): Promise<void> {
-        logger.info(`Broadcasting change channel to '${this.options.network.channelList[0]}'.`, NS);
-        await this.adapter.changeChannel(this.options.network.channelList[0]);
+    private async changeChannel(oldChannel: number, newChannel: number): Promise<void> {
+        logger.warning(`Changing channel from '${oldChannel}' to '${newChannel}'`, NS);
+        await this.adapter.changeChannel(newChannel);
+        logger.warning(`Channel changed to '${newChannel}'`, NS);
```
Nerivec, line 453: Not sure the post-change feedback needs `warning`?

src/controller/controller.ts
Koenkk: Agree, fixed!

Koenkk:
@Nerivec thanks for the feedback, the docs PR is also ready. Let me know if this can be merged

src/controller/controller.ts
Nerivec: Oh, I meant in `changeChannel` of adapter, to be able to remove `supportsChangeChannel` completely.
_Sidenote: Probably should review these (there's one for backup too), the ones that are absolutely necessary and only return true/false should probably be moved to a var in adapter class or something._

PR #1111, opened 2024-07-12
This adds samsung door lock manufacturer specific unlock command

baeksj:
samsung door uses unlock command (ID 31) with for existing cluster 'closuresDoorLock'

I referred to the code below.

https://github.com/SmartThingsCommunity/SmartThingsEdgeDrivers/blob/main/drivers/SmartThings/zigbee-lock/src/samsungsds/init.lua


Koenkk:
Please use the new custom cluster functionality  instead ([example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/fc6fd20c8730b32ea1c382539d024eb9cf04a679/src/devices/sonoff.ts#L52))

baeksj:
Thank you

PR #1114, opened 2024-07-15
add IEEE -> NWK Addr translation on RX for conbee 3

schrluka:
Hi

Some zigbee devices report their IEEE address (64 bit) in the NWK header of all frames they send. In this case ConBee 3 reports only the IEEE Address to the driver, even if the 16 bit NWK address is also present in the frame. This is causes an issue in the driver for conbee 3, as it uses the NWK address for various purposes. 
Typically the interview for such devices fails.

To fix these issue I added a simple address lookup.


Koenkk:
Thanks!

schrluka:
Thanks for merging so fast!

PR #1116, opened 2024-07-15
Update cluster.ts

turboproc:
Adding support for PM1 and PM10 clusters, and, changing attribute types to single.

More detail:
- Zigbee cluster definition specifies all attributes that belong to PM2 cluster as single (not INT16). See 4.13.2.1
- Zigbee cluster definition does not specify the PM1 and PM10 clusters. However considering that Zigbee cluster definition and Matter cluster definition follow each other 1:1, it seems fairly safe to use the Matter cluster code points for PM1 and PM10.


turboproc:
Updated spacing to fix tab alignment issue

turboproc:
Updated tstype to reflect added clusters

Koenkk:
Thanks!

PR #1119, opened 2024-07-17
fix: also restore checkinInterval from database if no inputCluster.genPollCtrl

sjorge:
When modernExtend.quirkCheckinInterval() got added, this used to work. But now the restoring of the value from database was behinda check on inputClusters.genPollCtrl. This modernExtend was added for devices with a fixed checkin interval (e.g. lumi devices and there near frequent genBasic messages) but that do not expose genPollCtrl.


Koenkk:
Thanks!

PR #1125, opened 2024-07-21
fix: ZStack: throw error when bind/unbind fails

Koenkk:
The bind/unbind response status was not checked with ZStack which made it look like bind/unbind succeeded even though it failed. This PR fixes that.

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/22492


Nerivec (approved):
👍

PR #1126, opened 2024-07-21
Ember: Add tests for adapter layer.

Nerivec:
Should provide 100% coverage for `EmberAdapter` class.
Most of the changes are code and strings cleanup.

Fixes:
- Don't hardcode touchlink waitress matching
- Ensure valid endpoint info in APSFrame for ZCL requests
- Add missing throw when ZCL to endpoint request hard fails
- Improve InterPAN lock check

_Note: This adds a lot more tests, so CI is expected to take a bit more time now._


Koenkk:
Thanks!

PR #1128, opened 2024-07-23
fix: ZStack: throw errors when ZDO calls fail

Koenkk:
Continuation of https://github.com/Koenkk/zigbee-herdsman/pull/1125

Fixes https://github.com/Koenkk/zigbee2mqtt/issues/22492

### DIFF
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -6,6 +6,7 @@ import {Queue, Waitress, Wait} from '../../../utils';
 import {logger} from '../../../utils/logger';
 import {BroadcastAddress} from '../../../zspec/enums';
 import * as Zcl from '../../../zspec/zcl';
+import {Status} from '../../../zspec/zdo';
 import Adapter from '../../adapter';
 import * as Events from '../../events';
 import {
@@ -28,6 +29,7 @@ import {
 import * as Constants from '../constants';
 import {Constants as UnpiConstants} from '../unpi';
 import {Znp, ZpiObject} from '../znp';
+import {ZpiObjectPayload} from '../znp/tstype';
 import {ZnpAdapterManager} from './manager';
 import {ZnpVersion} from './tstype';
 
@@ -174,19 +176,17 @@ class ZStackAdapter extends Adapter {
     public async getCoordinator(): Promise<Coordinator> {
         return this.queue.execute<Coordinator>(async () => {
             this.checkInterpanLock();
-            const activeEpRsp = this.znp.waitFor(UnpiConstants.Type.AREQ, Subsystem.ZDO, 'activeEpRsp');
+            const activeEpRsp = this.waitForAreqZdo('activeEpRsp');
             await this.znp.request(Subsystem.ZDO, 'activeEpReq', {dstaddr: 0, nwkaddrofinterest: 0}, activeEpRsp.ID);
-            const activeEp = await activeEpRsp.start().promise;
+            const activeEp = await activeEpRsp.start();
 
             const deviceInfo = await this.znp.request(Subsystem.UTIL, 'getDeviceInfo', {});
 
             const endpoints = [];
             for (const endpoint of activeEp.payload.activeeplist) {
-                const simpleDescRsp = this.znp.waitFor(UnpiConstants.Type.AREQ, Subsystem.ZDO, 'simpleDescRsp', {endpoint});
-
+                const simpleDescRsp = this.waitForAreqZdo('simpleDescRsp', {endpoint});
                 await this.znp.request(Subsystem.ZDO, 'simpleDescReq', {dstaddr: 0, nwkaddrofinterest: 0, endpoint}, simpleDescRsp.ID);
-
-                const simpleDesc = await simpleDescRsp.start().promise;
+                const simpleDesc = await simpleDescRsp.start();
 
                 endpoints.push({
                     ID: simpleDesc.payload.endpoint,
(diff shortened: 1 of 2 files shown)


src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -6,6 +6,7 @@ import {Queue, Waitress, Wait} from '../../../utils';
 import {logger} from '../../../utils/logger';
 import {BroadcastAddress} from '../../../zspec/enums';
 import * as Zcl from '../../../zspec/zcl';
+import {Status} from '../../../zspec/zdo';
```
Nerivec, line 9: Might want to import Zdo namespace to use `Zdo.Status`
`import * as Zdo from '../../../zspec/zdo';`
or rename
`import {Status as ZdoStatus} from '../../../zspec/zdo';`
to avoid confusion in the code.

(can probably replace several hardcoded values with consts/enums from `Zdo`, so the namespace will probably end up being better)

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -1117,6 +1096,27 @@ class ZStackAdapter extends Adapter {
         });
     }
 
+    private waitForAreqZdo(command: string, payload?: ZpiObjectPayload): {start: () => Promise<ZpiObject>; ID: number} {
```
Nerivec, line 1099: The status checking is implemented directly in the Zdo spec with `BuffaloZdo.readResponse(clusterId, buffer);` [link](https://github.com/Koenkk/zigbee-herdsman/blob/557ec54b89c1ec6f6a209726473267163bd44ee3/src/zspec/zdo/buffaloZdo.ts#L1420), so you could remove some of the logic here, but I guess that would mean bypassing the ZpiObject parsing for AREQ/ZDO and using the spec instead?
(it would automatically add support for r23 too, since the spec **should be** r23 compliant)

test/adapter/z-stack/adapter.test.ts
```diff
@@ -3497,8 +3513,10 @@ describe('zstack-adapter', () => {
         basicMocks();
         await adapter.start();
         mockZnpRequest.mockClear();
-        bindStatusResponse = 1;
-        await expect(adapter.bind(301, '0x129', 1, 1, 4, 'endpoint', 9)).rejects.toThrow(`Failed to bind '1' from '4/9' to '0x129/1' (1)`);
+        bindStatusResponse = 0x8e;
+        await expect(adapter.bind(301, '0x129', 1, 1, 4, 'endpoint', 9)).rejects.toThrow(
+            `ZDO error: bind failed with status 'DEVICE_BINDING_TABLE_FULL' (142)`,
+        );
     });
```
Nerivec, line 3520: I don't see changes to test throwing for functions that now do? Coverage missed it?

src/adapter/z-stack/adapter/zStackAdapter.ts
Koenkk: Fixed!

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -1117,6 +1096,27 @@ class ZStackAdapter extends Adapter {
...
+        const start = (): Promise<ZpiObject> => {
+            const startResult = result.start();
+            return new Promise<ZpiObject>((resolve, reject) => {
+                startResult.promise
+                    .then((response) => {
+                        // Even though according to the Z-Stack docs the status is `0` or `1`, the actual code
+                        // shows it sets the `zstack_ZdpStatus` which contains the ZDO status.
+                        const code = response.payload.status;
+                        if (code !== Status.SUCCESS) {
+                            reject(new Error(`ZDO error: ${command.replace('Rsp', '')} failed with status '${Status[code]}' (${code})`));
```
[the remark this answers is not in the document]
Koenkk: Moved the `resolve` into the `else`

src/adapter/z-stack/adapter/zStackAdapter.ts
Koenkk: > but I guess that would mean bypassing the ZpiObject parsing for AREQ/ZDO and using the spec instead?

The difference with ember is that there is one ZDO response callback where you get the full buffer, for zstack there is a callback per ZDO type, so we do not have the full buffer. So I don't think we can use this for zstack (?)

test/adapter/z-stack/adapter.test.ts
Koenkk: There were not tests for this yet, but given they all use `waitForAreqZdo` now I don't think it makes sense to add a bunch of tests for this?

src/adapter/z-stack/adapter/zStackAdapter.ts
Nerivec: Seems like zstack is modding it (like not passing the transaction seq num?), so that wouldn't work as-is. 😢
If it's just the seq num, I guess we could make a param in the spec to not skip 1 byte when reading?

Koenkk: So it means we need to reconstruct the buffer again from the parsed ZDO message?

Nerivec: Glanced at the code... I think something like this should allow a smooth-ish transition.

in ZpiObject:

```typescript
const ZDO_RSP_COMMAND_ID_TO_CLUSTER_ID: Map<number, Zdo.ClusterId> = new Map([
    [128, Zdo.ClusterId.NETWORK_ADDRESS_RESPONSE],
    // etc.
]);
```

```typescript
    public static fromUnpiFrame(frame: UnpiFrame): ZpiObject {
        if (frame.type === Type.AREQ && frame.subsystem === Subsystem.ZDO) {
            const clusterId = ZDO_RSP_COMMAND_ID_TO_CLUSTER_ID.get(frame.commandID);

            return new ZpiObject(
                frame.type,
                frame.subsystem,
                Zdo.ClusterId[clusterId],
                frame.commandID,// or clusterId?
                BuffaloZdo.readResponse(clusterId, frame.data, false),
                null,// will this be ok? (avoids lookup)
            );
        }

       // ...
```

in BuffaloZdo:
```typescript
    public static readResponse(clusterId: ZdoClusterId, buffer: Buffer, hasOverhead: boolean = true): unknown {
        const buffalo = new BuffaloZdo(buffer, hasOverhead ? ZDO_MESSAGE_OVERHEAD : 0); // set pos to skip `transaction sequence number`
       // ...
```

This would require matching the Zdo spec payloads throughout the code (object keys, [all typed though](https://github.com/Koenkk/zigbee-herdsman/blob/557ec54b89c1ec6f6a209726473267163bd44ee3/src/zspec/zdo/definition/tstypes.ts#L225)) and likely some tweaks to response waiters (cluster name instead of command name?). And handling the ZDO status throwing where `fromUnpiFrame` is called.

PR #1129, opened 2024-07-24
Move ESLint config and plugins to devDependencies

matt-oakes:
Hello. Thanks for all your work on Herdsman!

This PR moves the ESLint plugin and config dependencies into `devDependencies`. This ensures they are no included by other packages which depend on `zigbee-herdsman`, reducing the overall bundle size, and ensuring there are no clashes with versions of packages used by other projects.


package.json
```diff
@@ -31,8 +31,6 @@
     "bonjour-service": "^1.2.1",
     "debounce": "^2.1.0",
     "debug": "^4.3.5",
```
Koenkk, line 33: I think this one can also be removed (not used anymore)

package.json
```diff
@@ -31,8 +31,6 @@
...
-    "eslint-config-prettier": "^9.1.0",
-    "eslint-plugin-perfectionist": "^2.11.0",
     "fast-deep-equal": "^3.1.3",
     "mixin-deep": "^2.0.1",
     "mz": "^2.7.0",
```
Koenkk, line 36: `prettier` can also be moved to `devDependencies`

matt-oakes:
Thanks for the review. I've made those changes now.

Koenkk:
Thank you too!

matt-oakes:
Just out of interest, I checked the installed size of 0.55.1 to 0.55.2 (with this change) and it's bigger than expected:

- [`0.55.1`](https://pkg-size.dev/zigbee-herdsman@0.55.1) - 52MB with 157 packages
- [`0.55.2`](https://pkg-size.dev/zigbee-herdsman@0.55.1) - 8MB with 25 packages

Thanks again for merging this so quickly!

Koenkk:
Great, many thanks!

PR #1130, opened 2024-07-24
Improve loops performance

Nerivec:
- Use iterators to avoid nested looping whenever possible.
- Extract cached deleted devices (runtime only) to separate object to avoid lots of unnecessary loop checks.
- Avoid using `Object.values` with looping on top.

```logs
On a database with 100 records (with matching in mid for "find" tests):

┌─────────┬───────────────────────────┬─────────────┬────────────────────┬──────────┬──────────┐
│ (index) │ Task Name                 │ ops/sec     │ Average Time (ns)  │ Margin   │ Samples  │
├─────────┼───────────────────────────┼─────────────┼────────────────────┼──────────┼──────────┤
│ 0       │ 'Old Find by nwk'         │ '46,133'    │ 21676.031102459012 │ '±0.22%' │ 230670   │
│ 1       │ 'New Find by nwk'         │ '149,131'   │ 6705.480378715173  │ '±0.19%' │ 745659   │
│ 2       │ 'Old Find deleted by nwk' │ '42,740'    │ 23396.819942899692 │ '±0.31%' │ 213705   │
│ 3       │ 'New Find deleted by nwk' │ '7,754,479' │ 128.95771808723978 │ '±0.18%' │ 38772399 │
└─────────┴───────────────────────────┴─────────────┴────────────────────┴──────────┴──────────┘
┌─────────┬──────────────────────────────────────┬──────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                            │ ops/sec  │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼──────────────────────────────────────┼──────────┼────────────────────┼──────────┼─────────┤
│ 0       │ 'Old Device.all looped over'         │ '42,299' │ 23640.73759176489  │ '±0.23%' │ 211500  │
│ 1       │ 'New Device.all looped over'         │ '47,153' │ 21207.48536824284  │ '±0.23%' │ 235766  │
│ 2       │ 'New Device.allIterator looped over' │ '94,555' │ 10575.785707822934 │ '±0.25%' │ 472779  │
└─────────┴──────────────────────────────────────┴──────────┴────────────────────┴──────────┴─────────┘
┌─────────┬────────────────────────────────────────────────┬──────────┬────────────────────┬──────────┬─────────┐
│ (index) │ Task Name                                      │ ops/sec  │ Average Time (ns)  │ Margin   │ Samples │
├─────────┼────────────────────────────────────────────────┼──────────┼────────────────────┼──────────┼─────────┤
[...]

PR #1131, opened 2024-07-24
fix: Clear the loaded device database entries when the controller is stopped

matt-oakes:
Hello! thanks again for all the work on this great library.

I have hit an edge case with the way we were using the library which was causing some strange issues for us.

### Steps to reproduce
The issue stems from the fact that the `Devices` model is storing a cache of the loaded device database entries in the `devices` static member. This was causing an issue when we are:

1. Stopping the zigbee-herdsman controller with `controller.stop()`
2. Manually deleting a device from the database.
3. Then later reconnecting the controller with `controller.start`

We're manually removing it because we need to allow users to remove devices even when the adapter is not connected. In situations where the adapter is connected we allow `zigbee-herdsman` to handle it.

All this is happening in an Electron program which is not closed between these steps.

### Current behaviour
At the point of reconnecting the controller, we ask it for the list of devices. At that point `Device.devices` still contains the cached version so we are being returned devices that are no longer in the database.

For example, if the steps are:

1. Start controller
2. Add "device 1"
3. Add "device 2"
4. Stop the controller
5. Delete "device 2" from the database manually
6. Start the controller

At this point we are asking for the devices and getting back both "Device 1" and "Device 2".

### Expected behaviour
I would expect that once the controller stops and restarts it essentially starts from a clean slate. This would allow us to restart the controller and know that it's going to respect the current state of the database.

In the example above, this would mean responding with just "Device 1".

### What the fix does
I have resolved this in this PR by adding an additional step to clear the `Device.devices` cache when the controller stops. This measn the next time `Device.loadFromDatabaseIfNecessary` is called it will now load from the database again and get the correct new values.

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -334,6 +334,8 @@ class Controller extends events.EventEmitter {
 
             this.adapterDisconnected = true;
         }
+
+        Device.clearLoadedDatabaseDevices();
     }
 
     private databaseSave(): void {
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -593,6 +593,10 @@ class Device extends Entity {
         }
     }
 
+    public static clearLoadedDatabaseDevices(): void {
+        Device.devices = null;
+    }
+
     public static find(ieeeOrNwkAddress: string | number, includeDeleted: boolean = false): Device {
         return typeof ieeeOrNwkAddress === 'string'
             ? Device.byIeeeAddr(ieeeOrNwkAddress, includeDeleted)


Nerivec:
Ok to incorporate this in https://github.com/Koenkk/zigbee-herdsman/pull/1130?
Would need to do the same for `Device.deletedDevices` and `Groups.groups` for consistent behavior across the board.

matt-oakes:
Sure! Are you going to incorporate it or are you asking me to do that?

### Related but much bigger potential change

In an ideal world for me the entire database concept would moved from zigbee-herdsman and into zigbee2mqtt.

I would replace it with an interface where herdsman can ask for the list of stored devices and pass through updates that should be saved.

This would make this library much easier to integrate with other systems as it is currently a little odd to have a library writing to the disk itself.

I understand why you’ve built it this way though, and know that’s a much bigger change, so I understand if your preference is to stick with the current design.

Nerivec:
I'll do it 👍

matt-oakes:
Ok. Thank you!

And thanks again for all the work on the library

PR #1134, opened 2024-07-27
fix: call fsync before rename

LaurentvdBos:
When writing `database.db` to disk, we should call fsync on the file descriptor used for writing. See

https://pubs.opengroup.org/onlinepubs/9699919799/functions/fsync.html

Which explicitly states that "The fsync() function shall request that all data for the **open file descriptor** ..." (emphasis mine)

I also enforced that database.db.tmp is not created if it already exists, in which case an Exception is thrown. That should be an indication that something went wrong in the past, but maybe this may be annoying. Should this behavior be properly handled in a try/catch and a nice logging message?

### DIFF
--- a/src/controller/database.ts
+++ b/src/controller/database.ts
@@ -101,7 +101,18 @@ class Database {
 
         const tmpPath = this.path + '.tmp';
 
-        const fd = fs.openSync(tmpPath, 'wx');
+        try {
+            // If there already exsits a database.db.tmp, rename it to database.db.tmp.<now>
+            const dateTmpPath = tmpPath + '.' + new Date().toISOString().replaceAll(':', '-');
+            fs.renameSync(tmpPath, dateTmpPath);
+
+            // If we got this far, we succeeded! Warn the user about this
+            logger.warning(`Found '${tmpPath}' when writing database, indicating past write failure; renamed it to '${dateTmpPath}'`, NS);
+        } catch {
+            // Nothing to catch; if the renameSync fails, we ignore that exception
+        }
+
+        const fd = fs.openSync(tmpPath, 'w');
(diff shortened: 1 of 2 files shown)


LaurentvdBos:
NB, this should address issue koenk/zigbee2mqtt#11759, or well, at least improve the situation.

Koenkk:
If I understand correctly, the current code doesn't do anything because a fsync is called on a file descriptor where nothing was written to?

> Should this behavior be properly handled in a try/catch and a nice logging message?

Yes, otherwise we cannot get out of the situation without manual intervention. I propose to, before opening the fd, do a check whether the tmp db file exists, if it does, postfix it with the unix timestamp and log a warning message.

LaurentvdBos:
> If I understand correctly, the current code doesn't do anything because a fsync is called on a file descriptor where nothing was written to?

Indeed, that can happen _in theory_. File systems add many checks to make sure the data does hit the disk in time, but especially when running on SD cards stuff may get cached aggressively before it appears on the disk (Raspberry PI users, for example).

> Yes, otherwise we cannot get out of the situation without manual intervention. I propose to, before opening the fd, do a check whether the tmp db file exists, if it does, postfix it with the unix timestamp and log a warning message.

Good point, so there is nothing else that catches the existence of a database.db.tmp. I added a slightly adapted behavior: if database.tmp already exists, rename it to database.tmp.<NOW> and warn. In that case, only one warning is emitted.

I broke CI since this adds an untested branch. Let me know whether you like the new behavior, because then I will add a test for it.

Koenkk:
> Let me know whether you like the new behavior, because then I will add a test for it.

Looks good!

Koenkk:
💯

PR #1139, opened 2024-08-03
fix: Allow to set adapter transmit power on startup

Koenkk:
@Nerivec probably the ember part is wrong, feel free to update.


Nerivec (approved):
👍

Ember part should be:
- set tx power if mismatching from adapter
- set tx power during form (saved across reboots, so should be a noop on future restarts as long as not changed)

Koenkk:
Thanks!

PR #1144, opened 2024-08-08
Improve bind/unbind logic.

Nerivec:
- Check if bind exists before triggering unbind request.
- Ensure group ID exists when unbinding with number (always created when addBinding/bind, so should always exist).
- Add internal `addBinding` function to avoid cluster lookup when available in caller.
- Optimize getters.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -453,6 +453,10 @@ class Endpoint extends Entity {
         );
     }
 
+    public hasBind(clusterId: number, target: Endpoint | Group): boolean {
+        return this.getBindIndex(clusterId, target) !== -1;
+    }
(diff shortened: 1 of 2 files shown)


src/controller/model/endpoint.ts
```diff
@@ -469,13 +453,22 @@ class Endpoint extends Entity {
         );
     }
 
+    public getBindIndex(clusterId: number, target: Endpoint | Group): number {
```
Koenkk, line 456: Maybe we should an additional `hasBind()`? That will be more clear for clients.

Nerivec: Done!

Koenkk:
Thanks!

PR #1146, opened 2024-08-10
Enforce TS `strict` type checking

Nerivec:
TODO:
- [x] adapter
- [x] adapter/deconz
- [x] adapter/ember
- [x] adapter/ezsp
- [x] adapter/z-stack
- [x] adapter/zigate
- [x] buffalo
- [x] controller
- [x] models
- [x] utils
- [x] zspec
- [ ] https://github.com/Koenkk/zigbee2mqtt/pull/23601
- [ ] Match changes in ZHC

Also:
- Use of `undefined` instead of `null` allows to reduce MQTT payload sizes by skipping the property entirely (_often multiplied by number of devices, so, should make a decent difference_)
- Fix related types
- Move small utils into single file
- Cleanup backup conversion utils
- Add `asyncSet` function to `SerialPort`
- Ember: use constructor/new to init state for Ezsp/Ash layers and use `disconnected` event to restart Z2M when adapter needs resetting instead of handling internally (aligns behavior with other stacks, ensures clean slate).

PR #1148, opened 2024-08-12
Recognize broadcasted GP frames

LaurentvdBos:
Change suggested by @Koenkk to fix issue [#23575](https://github.com/Koenkk/zigbee2mqtt/issues/23575) (@Nerivec may also want to look).

Confirmed that it solves the issue (at least for me)


Nerivec:
Thanks. I wrote the fix already (with an extra check to avoid too much noise) but will push it after https://github.com/Koenkk/zigbee-herdsman/pull/1146 is merged (lots of changes, don't want to mess up the flow with conflicts).
I'll come back and close this when I create the PR. 😉

PR #1149, opened 2024-08-13
Add manuSpecificLumi curtain attributes

marthubner:
Based on https://github.com/Koenkk/zigbee-herdsman-converters/pull/7848


Koenkk:
Could you use the custom cluster functionality in zigbee herdsmand converters instead? Look at lumi.ts for an example

marthubner:
There is no custom cluster in lumi.ts but I found an example in ikea.ts at STARKVIND namely "export function addCustomClusterManuSpecificIkeaAirPurifier()", is that what you mean?
Is it preferable to do this kind of thing with all the manuSpecificLumi attributes introduced with Aqara C3 curtain motor or is it fine the way it is now?
Also why not add it here?

Koenkk:
Because we want to get rid of the custom clusters in zigbee-herdsman, you can find another example [here](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts#L52)

PR #1150, opened 2024-08-15
feat: Add install code format with pipe delimiter

yoo:
Add a new install code in the format off `<address>|<key>`. This format can be found on Bosch devices.


Koenkk:
Thanks!

PR #1153, opened 2024-08-17
Ember: fix CCA issues in busy environments (broadcast errors).

Nerivec:
Targeting https://github.com/Koenkk/zigbee2mqtt/issues/22453
This issue appears to only occur when the 2.4GHz range is crowded.
Has been tested in very busy environments affected by the bug. Z2M features depending on broadcasts (like permit join) should no longer break.

This PR also provides a new stack config, `CCA_MODE`, to allow customization of the CCA behavior for more specific cases.

_Note: Awaiting an explanation from Silabs on this firmware behavior._


Koenkk:
Thanks!

PR #1154, opened 2024-08-17
Migrate to eslint 9

Koenkk:
In this PR:
- Migrate to eslint 9
- Swap perfections import sorting for prettier import sorting (via `@ianvs/prettier-plugin-sort-imports`, `trivago/prettier-plugin-sort-imports` throws several `SyntaxError: Unexpected token` errors)

Once merged I will make a similar PR for z2m and zhc


.prettierrc
```diff
@@ -5,5 +5,7 @@
     "printWidth": 150,
     "bracketSpacing": false,
     "endOfLine": "lf",
-    "tabWidth": 4
+    "tabWidth": 4,
+    "importOrder": ["", "^node:", "", "<THIRD_PARTY_MODULES>", "", "^zigbee", "", "^[./]"],
```
Nerivec, line 9: Good call on the alt package.
I'd say let's take advantage of it:
```js
    importOrder: [
        "<TYPES>^(node:)",
        "",
        "<TYPES>",
        "",
        "<TYPES>^[.]",
        "",
        "<BUILTIN_MODULES>",
        "",
        "<THIRD_PARTY_MODULES>",
        "",
        "^zigbee",
        "",
        "^[.]",
    ],
```

Most will never be used, but at least it's nice and tidy.

src/utils/backup.ts
```diff
@@ -1,7 +1,7 @@
-/* eslint-disable max-len */
-import {fs} from 'mz';
 import * as path from 'path';
 
+import {fs} from 'mz';
```
Nerivec, line 3: Should remove this outdated dependency in favor in built-in node equivalent.

.prettierrc
Koenkk: Done!

src/utils/backup.ts
Koenkk: Done

PR #1161, opened 2024-08-21
Inovelli - Removing manuSpecific Inovelli because it was moved to the…

InovelliUSA:
… converter

@Koenkk hopefully this is correct. If you remember, I moved this to the converter a couple of weeks ago.


Koenkk:
Thanks!

PR #1166, opened 2024-08-28
Update cluster.ts | standard ZCL cluster 042b

smartboxchannel:
Because this is a standard ZCL cluster, not some specific heiman cluster. Due to the current implementation, when configuring reports of any device working with this standard cluster, z2m sends the manufacturer's code in a z2m packet, and the devices reject such a setting. This needs to be fixed. If heiman has such an implementation, then please implement it in the converter of a specific device from heiman.


Koenkk:
Thanks!

Issue #1169, opened 2024-08-30
ZiGate reverse IEEE address on adding devices (probaby in other messages)


devbis:
Potentially after https://github.com/Koenkk/zigbee-herdsman/commit/73162478f5e5c3d9058bfada27dfff52b9d2c612 ieee addresses are parsed in the reverse order. 
Need to figure out the root cause.

E.g.
Telink as `0x7f4b22587b38c1a4` while real MAC prefix is `0xa4c138`
NXP as `0x1bb5da00008d1500`, read prefix is `0x00158d00`

PR #1170, opened 2024-08-30
fix: zigate: parse all values as big endian

devbis:
NXP chips are big-endian, so all serial communication is going in big endian multi-byte values.
I changed all conversions to Big Endian which simplifies the code and also fixes reversed IEEE addresses.

Fixes: https://github.com/Koenkk/zigbee-herdsman/issues/1169


devbis:
@Koenkk it would be nice to see this fix merged in the next release as the long as addresses are broken in 1.39.1 for Zigate

Koenkk:
Thanks, it will be part of the next release!

PR #1175, opened 2024-09-05
fix: Fix Deconz Green power implementation

Koenkk:
To fix https://github.com/Koenkk/zigbee2mqtt/issues/23814, waiting for confirmation


src/adapter/deconz/driver/frameParser.ts
```diff
@@ -382,7 +382,7 @@ function parseGreenPowerDataIndication(view: DataView): object | null {
             ind.commandId = view.getUint8(17);
             ind.commandFrameSize = view.byteLength - 18 - 6; // cut 18 from begin and 4 (sec mic) and 2 from end (cfc)
 
-            const payload = [];
+            const payload = Buffer.alloc(ind.commandFrameSize);
             let i = 0;
             for (let u = 18; u < ind.commandFrameSize + 18; u++) {
```
Nerivec, line 387: Can't just slice the underlaying buffer to avoid looping here (and below)?

Koenkk: Yes, much cleaner, thanks!

PR #1176, opened 2024-09-06
Use `Map` for runtime Device/Group lookups.

Nerivec:
Improved test to cover all cases where lookups should be synced.
Fixed group getter return type.

_Also marked functions changed for iterator versions in previous PR as deprecated. Pretty sure they can just be removed, but need to double-check and update tests (used in `expect` in a few places). Also might want to instead rename `xyzIterator` to `xyz` for cleaner names. Can be dealt with in another PR._

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -414,6 +414,7 @@ class Controller extends events.EventEmitter<ControllerEventMap> {
 
     /**
      * Get all devices
+     * @deprecated use getDevicesIterator()
      */
     public getDevices(): Device[] {
         return Device.all();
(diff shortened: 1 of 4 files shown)


Nerivec:
We could also optimize Device lookup by nwk addr with a `Map<number, string>`:
```typescript
    public static byNetworkAddress(networkAddress: number, includeDeleted: boolean = false): Device | undefined {
        Device.loadFromDatabaseIfNecessary();

        const ieeeAddr = Device.nwkToIeeeCache.get(networkAddress);

        return ieeeAddr ? Device.byIeeeAddr(ieeeAddr, includeDeleted) : undefined;
    }
```
With cache updated in `create`, `loadFromDatabaseIfNecessary` and `changeIeeeAddress` functions.

Koenkk (approved):

Koenkk:
> We could also optimize Device lookup by nwk addr with a Map<number, string>:

Sounds good!

Koenkk:
Thanks!

PR #1178, opened 2024-09-07
Improve performance when logging is disabled

tomer-w:
As described in: https://github.com/Koenkk/zigbee2mqtt/pull/23893, this PR change the expensive usage of ```JSON.stringify``` usage in tracing to use the new lambda functionality.


src/adapter/z-stack/znp/znp.ts
```diff
@@ -69,26 +69,22 @@ class Znp extends events.EventEmitter {
...
-            if (type === Type.SREQ) {
-                logger.debug(`SREQ: ${message}`, NS);
-            } else {
-                throw new Error(`Unknown type '${type}'`);
-            }
+    private static stringifyObject(prefix: string, object: ZpiObject): string {
+        return `${Type[object.type]}: ${prefix} ${Subsystem[object.subsystem]} - ${object.command} - ${JSON.stringify(object.payload)}`;
+    }
+
+    private log(prefix: string, object: ZpiObject): void {
```
Nerivec, line 76: Should remove `stringifyObject` and `log` functions and declare `toString` in `ZpiObject`

```typescript
    public toString(direction: '<--' | '-->'): string {
        return `${Type[this.type]}: ${direction} ${Subsystem[this.subsystem]} - ${this.command} - ${JSON.stringify(this.payload)}`;
    }
```
then just call `logger.debug(() => object.toString('<--'), NS);` directly in the 2 places where it's used.

src/controller/model/endpoint.ts
```diff
@@ -714,10 +714,10 @@ class Endpoint extends Entity {
             optionsWithDefaults.reservedBits,
         );
 
-        const log =
+        const messageBuilder = (): string =>
             `CommandResponse ${this.deviceIeeeAddress}/${this.ID} ` +
             `${cluster.name}.${command.name}(${JSON.stringify(payload)}, ${JSON.stringify(optionsWithDefaults)})`;
-        logger.debug(log, NS);
+        logger.debug(() => messageBuilder(), NS);
```
Nerivec, line 720: No wrapping anon function necessary here, can just pass `messageBuilder` as param. Same for all other occurrences in the PR.

src/controller/model/endpoint.ts
```diff
@@ -739,7 +739,7 @@ class Endpoint extends Entity {
             });
         } catch (error) {
             const err = error as Error;
-            err.message = `${log} failed (${err.message})`;
+            err.message = `${messageBuilder()} failed (${err.message})`;
             logger.debug((err as Error).stack!, NS);
```
Nerivec, line 743: Should cleanup these (even though not strictly related to this PR, looks like there are a few occurrences).
`(err as Error)` => `err` (already declared as such with `const err = error as Error;`)

src/adapter/z-stack/znp/znp.ts
tomer-w: I made the change, but since I don't think direction belongs to the object, I kept it outside of the toString().

src/controller/model/endpoint.ts
tomer-w: Done.

src/controller/model/endpoint.ts
tomer-w: Fixed.

tomer-w:
I submitted all the changes after the CR. Thanks!

src/controller/model/endpoint.ts
Nerivec: @Koenkk Your opinion on the name for these functions? `messageBuilder` doesn't follow the usual "action/verb" pattern of a function. `getLogString`?

src/adapter/z-stack/znp/znp.ts
```diff
@@ -69,26 +69,10 @@ class Znp extends events.EventEmitter {
...
-            }
-        }
-    }
-
     private onUnpiParsed(frame: UnpiFrame): void {
         try {
             const object = ZpiObject.fromUnpiFrame(frame);
-            const message = `<-- ${Subsystem[object.subsystem]} - ${object.command} - ${JSON.stringify(object.payload)}`;
-            this.log(object.type, message);
+            logger.debug(() => `<-- ${object}`, NS);
```
Koenkk, line 75: Why was this changed?

Koenkk: Never mind, just noticed the `toString()`

src/controller/model/endpoint.ts
Koenkk: What about `createLogMessage`?

Koenkk:
Thanks @tomer-w !

PR #1181, opened 2024-09-09
fix: Add missing response ID

canardos:
The ZCL view group command response is defined correctly in `Zcl.Clusters.genGroups.commandResponse.viewRsp` (id=1), but the ID is missing from `Zcl.Clusters.genGroups.commands.view`.

This leads to the a timeout when the the view groups command is sent, because the lib is waiting on a response with incorrect `frameType` and `commandIdentifier`, which never arrives.

This PR adds the missing response ID to the command.


Koenkk:
Thanks!

PR #1186, opened 2024-09-14
ZDO spec: improve build/read logic and typing.

Nerivec:
In preparation for ZDO removal from adapters:
- Fixup consts usage from spec and some typos
- No longer throw when reading response status !== 0, status is returned as tuple `[status, payload | undefined]`
- Add `buildRequest` with auto param typing based on `clusterId` to match `readResponse` usage pattern
- Add typing for `readResponse` based on `clusterId`
- Add `checkStatus` static util to type result of `readResponse` based on `status`
- Improve tests

_NOTE: Some adjacent stuff also needs improving, but better to wait for next PR, since some of it is going to change/disappear completely._

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -21,7 +21,7 @@ interface AdapterEventMap {
 }
 
 abstract class Adapter extends events.EventEmitter<AdapterEventMap> {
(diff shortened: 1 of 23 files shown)


Koenkk:
Thanks!

PR #1187, opened 2024-09-14
Prepare to move specific ZDO requests out of Adapter

Nerivec:
Final goal is to remove the following functions from `Adapter` in favor of a generic `sendZdo` with requests built and sent directly from `Controller`/`Device`/`Endpoint`:
  - `changeChannel`
  - `permitJoin`
  - `lqi`
  - `routingTable`
  - `nodeDescriptor`
  - `activeEndpoints`
  - `simpleDescriptor`
  - `bind`
  - `unbind`
  - `removeDevice`

This should make the API far more robust against changes (by not requiring updates in every adapter) and offer full payloads to the upstream code (instead of the current limited set of properties), which could lead to a few new features.
Due to the sizeable overhaul required in some stacks, this PR will focus first on introducing the `sendZdo` function while keeping above functions in place until they can safely be removed (all adapters are required to have support before that can happen).

_A working example (albeit using an older implementation) of the full update for `ember` (as well as some tentative partial implementations in other stacks) can be found [in this branch](https://github.com/Koenkk/zigbee-herdsman/compare/master...Nerivec:zigbee-herdsman:zdo-tmp)._

TODO:
- [x] Determine & implement per-stack request payloads variations (see [*-zdo.refactor.test.ts](https://github.com/Nerivec/zigbee-herdsman/tree/zdo-tmp/test))
- [x] Refactor above functions to use `sendZdo`
  - [x] `ember`
  - [x] `zstack`
  - [x] `deconz`
  - [x] `zigate` _(need testing)_
  - [x] `zboss`
  - [x] `ezsp`
- [x] Refactor payload parsing to use `Zdo.Buffalo.readResponse` for ZDO frames
  - [x] `ember`
  - [x] `zstack`
  - [x] `deconz`
  - [x] `zigate` _(need testing)_
  - [x] `zboss`
  - [x] `ezsp`

TODO separate PR:
- [ ] Move ZDO requests to upstream code (_done but not yet added to the PR for compat._)
- [ ] Remove `deviceAnnounce` and `networkAddress` event from `Adapter`
- [ ] Remove above functions from `Adapter` & impl.
  - [ ] Deal with `permitJoin` special-case
- [ ] Fix tests
  - [ ] `Controller`
  - [ ] `ember`
  - [ ] `zstack`

CC: @kirovilya @ChrisHae @G1K @devbis contributions/ideas on specific stacks welcome!


src/controller/controller.ts
```diff
@@ -164,6 +166,7 @@ class Controller extends events.EventEmitter<ControllerEventMap> {
         // Register adapter events
         this.adapter.on('deviceJoined', this.onDeviceJoined.bind(this));
         this.adapter.on('zclPayload', this.onZclPayload.bind(this));
+        this.adapter.on('zdoResponse', this.onZdoResponse.bind(this));
         this.adapter.on('disconnected', this.onAdapterDisconnected.bind(this));
         this.adapter.on('deviceAnnounce', this.onDeviceAnnounce.bind(this));
```
Koenkk, line 171: I think these can be removed now?

Nerivec: Will do upstream removals in other PR, but yea, with a bit of clever work (zigate...), should be able to remove `deviceAnnounce` and `networkAddress` completely in favor of `zdoResponse` that triggers the same logic at the end.

Nerivec:
@kirovilya Could you test out zboss?
@devbis @fairecasoimeme Could you test out zigate?

Lots of moving parts, need to confirm nothing broke. If you can report back issues.
A round of `permit join` + `join device` (interview) + `network map` + `remove device` should pretty much trigger all involved ZDOs.
Change channel requires changing the channel in `configuration.yaml` then starting Z2M again. (Though depending on the start logic of the adapter, [it might not trigger](https://github.com/Koenkk/zigbee-herdsman/blob/a334e76744bf7b0297f61b41f6dbc4dacbcd6d31/src/controller/controller.ts#L152) at all.)

kirovilya:
it doesn't work for zboss :)
we need to come up with something else for mapping the structures of ZDO responses.
for example, a query for active endpoints ZDO_ACTIVE_EP_REQ:
was:
```
    {name: 'len', type: DataType.UINT8},
    {name: 'endpoints', type: BuffaloZclDataType.LIST_UINT8, options: (payload, options) => (options.length = payload.len)},
    {name: 'nwk', type: DataType.UINT16},
=========================================================
zh:zboss:uart: <-- FRAME: 0001060211000002f2010000
zh:zboss:driv: <== Frame: {"version":0,"type":1,"commandId":518,"tsn":17,"payload":{"category":0,"status":0,"len":2,"endpoints":[242,1],"nwk":0}}

```
became:
```
    /**
     * Response: [transaction sequence number: 1]
     *           [status:1] [node ID:2] [endpoint count:1] [endpoint:1]*
     */
    ACTIVE_ENDPOINTS_RESPONSE = 0x8005,
=========================================================
zh:zboss:uart: <-- FRAME: 0001060215000002f2010000
zh:zboss:driv: <== Frame: {"version":0,"type":1,"commandId":518,"tsn":21,"payload":{"category":0,"status":0,"zdoClusterId":32773,"zdo":[0,{"nwkAddress":61954,"endpointList":[0]}]}}
```
for some other commands as well.

src/adapter/zboss/frame.ts
```diff
@@ -103,20 +106,40 @@ export function readZBOSSFrame(buffer: Buffer): ZBOSSFrame {
...
+        const zdo = BuffaloZdo.readResponse(false, zdoResponseClusterId, buffer.subarray(6));
+
+        return {
+            version,
+            type,
+            commandId,
+            tsn,
+            payload: {
+                category,
+                zdoClusterId,
```
kirovilya, line 130: need status attribute for success controll:
`status: zdo[0],`

Nerivec: Success control will be done upstream (Controller/Device, etc), `Zdo.Buffalo` handles reading based on status and passes it back to caller (will allow workaround for some quirky devices in the future).

So, removing this chunk:
https://github.com/Nerivec/zigbee-herdsman/blob/6ba120509c2cee2c3415e7fdbf72964da56cfa5f/src/adapter/zboss/driver.ts#L357-L359
should fix it?

kirovilya: if the status is controlled by the caller, then Yes, there is no need for driver-level control for ZDO commands

Nerivec: Last two commits should take care of the two issues. Can you confirm?

src/adapter/zboss/adapter/zbossAdapter.ts
```diff
@@ -187,124 +210,202 @@ export class ZBOSSAdapter extends Adapter {
...
-            return await this.queue.execute<void>(async () => {
-                await this.driver.permitJoin(networkAddress, seconds);
-                if (!networkAddress) {
-                    // send broadcast permit
-                    await this.driver.permitJoin(0xfffc, seconds);
+            const clusterId = Zdo.ClusterId.PERMIT_JOINING_REQUEST;
+
+            if (networkAddress !== undefined) {
+                // TODO: works on networkAddress === 0 for 'coordinator-only'?
+                await this.driver.execCommand(CommandId.ZDO_PERMIT_JOINING_REQ, {nwk: networkAddress, duration: seconds, tcSignificance: 1});
```
kirovilya, line 240: remove it, since the command.ZDO_PERMIT_JOINING_REQ was removed from the description

src/adapter/zboss/adapter/zbossAdapter.ts
```diff
@@ -187,124 +210,202 @@ export class ZBOSSAdapter extends Adapter {
...
+                    // TODO: will disappear once moved upstream
+                    throw new Zdo.StatusError(result[0]);
                 }
-            });
+            } else {
+                // broadcast permit joining ZDO
+                // `authentication`: TC significance always 1 (zb specs)
+                const zdoPayload = Zdo.Buffalo.buildRequest(this.hasZdoMessageOverhead, clusterId, seconds, 1, []);
+
+                await this.sendZdo(ZSpec.BLANK_EUI64, ZSpec.BroadcastAddress.DEFAULT, clusterId, zdoPayload, true);
```
kirovilya, line 256: there is a special feature for the ZBOSS - you need to send 2 commands: to the coordinator and the broadcast.
`await this.sendZdo(ZSpec.BLANK_EUI64, networkAddress, clusterId, zdoPayload, true);`

src/adapter/zboss/frame.ts
kirovilya: it is better, but there is a problem - when connecting the device, a frame of a slightly different format comes:

```
debug:  zh:zboss:uart: <-- FRAME: 00020c020cda603602602bd5b3708e
debug:    zh:zboss:uart: <-- error RangeError: The value of "offset" is out of range. It must be >= 0 and <= 8. Received 10
    at new NodeError (node:internal/errors:406:5)
    at boundsError (node:internal/buffer:88:9)
    at Buffer.readUInt8 (node:internal/buffer:254:5)
    at BuffaloZdo.readUInt8 (/home/kirov/Projects/zigbee2mqtt/node_modules/zigbee-herdsman/src/buffalo/buffalo.ts:34:35)
    at BuffaloZdo.readEndDeviceAnnounce (/home/kirov/Projects/zigbee2mqtt/node_modules/zigbee-herdsman/src/zspec/zdo/buffaloZdo.ts:1815:35)
    at Function.readResponse (/home/kirov/Projects/zigbee2mqtt/node_modules/zigbee-herdsman/src/zspec/zdo/buffaloZdo.ts:1517:32)
    at readZBOSSFrame (/home/kirov/Projects/zigbee2mqtt/node_modules/zigbee-herdsman/src/adapter/zboss/frame.ts:152:32)
    at ZBOSSUart.onPackage (/home/kirov/Projects/zigbee2mqtt/node_modules/zigbee-herdsman/src/adapter/zboss/uart.ts:265:41)

```

kirovilya: I'm looking for a way to fix it

src/adapter/zboss/adapter/zbossAdapter.ts
Nerivec: So proper logic is like this?

```typescript
if (networkAddress) {
    // device-only when network > 0
    await this.sendZdo(ZSpec.BLANK_EUI64, networkAddress, clusterId, zdoPayload, true);
} else {
    // coordinator-only when networkAddress === 0 or undefined (all)
    await this.sendZdo(ZSpec.BLANK_EUI64, ZSpec.COORDINATOR_ADDRESS, clusterId, zdoPayload, true);

    if (networkAddress === undefined) {
        // all
        await this.sendZdo(ZSpec.BLANK_EUI64, ZSpec.BroadcastAddress.DEFAULT, clusterId, zdoPayload, true);
    }
}

Comment (kirovilya) [reply src/adapter/zboss/adapter/zbossAdapter.ts:255]: yes. or combine in one block with 2 commands sendZdo - always send a broadcast, even if we connect it to some router

File: src/adapter/zboss/frame.ts (hunk 2 above)
Comment (kirovilya) [reply src/adapter/zboss/frame.ts:161]:
```
   /**
     * Request: [transaction sequence number: 1]
     *          [node ID:2] [EUI64:8] [capabilities:1]
     */
    END_DEVICE_ANNOUNCE = 0x0013,
```
```     // Device announce indication
    // [CommandId.ZDO_DEV_ANNCE_IND]: {
    //     request: [],
    //     response: [],
    //     indication: [
    //         {name: 'nwk', type: DataType.UINT16},
    //         {name: 'ieee', type: DataType.IEEE_ADDR},
    //         {name: 'macCapabilities', type: DataType.UINT8},
    //     ],
    // },
```
The problem here is that a frame with the FrameType.INDICATION type arrives, which means there is no category in the frame, but data immediately goes.
Therefore, for this type of frame, you need to skip the line:
`const category = buf.readUInt8();`

kirovilya: and something else...

src/adapter/zboss/adapter/zbossAdapter.ts
Nerivec: Not separating it would break what Z2M expects (from the dropdown in frontend).
- `123` => device-only
- `0` => coordinator-only
- `undefined` => all (broadcast)

So I'll push like that, let me know if that doesn't work out in zboss.

kirovilya: to join to the coordinator, I have to send 2 commands: one with the coordinator's address (0), and the second with the broadcast. only in this case, the joining of devices begins.
to join to the router, I sent only one command (without broadcast), but the device did not connect to it (the concentrator did not give permission to connect) and I haven't been able to beat it yet. maybe we should send another broadcast team, but I haven't checked yet.

Nerivec: I'm seeing a `zb_nlme_permit_joining_request(duration)` in espressif zboss library. Could it be similar to emberznet, where there's a special function to open the network on the coordinator-only without going through a ZDO?

src/adapter/zboss/frame.ts
Nerivec: Darn, missed that one! Should be fixed.

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -221,14 +214,25 @@ class ZStackAdapter extends Adapter {
     }
 
     public async permitJoin(seconds: number, networkAddress?: number): Promise<void> {
-        const addrmode = networkAddress === undefined ? 0x0f : 0x02;
-        const dstaddr = networkAddress || 0xfffc;
+        const clusterId = Zdo.ClusterId.PERMIT_JOINING_REQUEST;
```
Koenkk, line 217: Note to self: check if this doesn't break https://github.com/Koenkk/Z-Stack-firmware/blob/7398d834eb3a790876c280293c4181da96cc7114/coordinator/Z-Stack_3.x.0/firmware.patch#L274

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -308,68 +321,176 @@ class ZStackAdapter extends Adapter {
...
+    ): Promise<ZdoTypes.RequestToResponseMap[K] | void> {
+        return await this.queue.execute(async () => {
+            this.checkInterpanLock();
+
+            // stack-specific requirements
+            switch (clusterId) {
+                case Zdo.ClusterId.PERMIT_JOINING_REQUEST: {
+                    const finalPayload = Buffer.alloc(payload.length + 3);
+                    finalPayload.writeUInt8(ZSpec.BroadcastAddress[networkAddress] ? AddressMode.ADDR_BROADCAST : AddressMode.ADDR_16BIT, 0);
+                    // TODO: confirm zstack uses AddressMode.ADDR_16BIT + ZSpec.BroadcastAddress.DEFAULT to signal "coordinator-only" (assumed from previous code)
```
Koenkk, line 417: Note to self; check this

src/adapter/z-stack/znp/znp.ts
```diff
@@ -308,6 +312,30 @@ class Znp extends events.EventEmitter {
         });
     }
 
+    public requestZdo(clusterId: ZdoClusterId, payload: Buffer, waiterID?: number): Promise<void> {
```
Koenkk, line 315: My expectation was that we don't use the ZNP specific ZDO commands but just `AF_DATA_REQUEST`, wouldn't it be easier to use that?

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -308,68 +321,176 @@ class ZStackAdapter extends Adapter {
...
 
     public async nodeDescriptor(networkAddress: number): Promise<NodeDescriptor> {
-        return await this.queue.execute<NodeDescriptor>(async () => {
-            this.checkInterpanLock();
-            try {
-                const result = await this.nodeDescriptorInternal(networkAddress);
-                return result;
-            } catch (error) {
-                logger.debug(`Node descriptor request for '${networkAddress}' failed (${error}), retry`, NS);
-                // Doing a route discovery after simple descriptor request fails makes it succeed sometimes.
```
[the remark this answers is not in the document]
Nerivec: The discovery bit? I made a note of that.

I'm thinking maybe adding a function to `Adapter` for failed requests, something like this: `onFailedRequest(ieeeAddress, networkAddress, profileId, clusterId)`, which might also allow to make the other various retry logic more generic in the future (upstream in Controller/Device instead of Adapter).
Thoughts?
I guess we could also have a `discoverRoute` function in Adapter (which may not do anything if not needed). This is quite stack-specific, emberznet does this automatically for any frame via APS option for example.
I guess we could temporarily add it back in `sendZdo`, I expect this would more or less do the same?
```typescript
            if (clusterId === Zdo.ClusterId.NODE_DESCRIPTOR_REQUEST) {
                await this.discoverRoute(networkAddress);
            }

            await this.znp.requestZdo(clusterId, payload, waiter?.ID);
```

src/adapter/z-stack/znp/znp.ts
Nerivec: Don't know if the ZDO rsp payloads will match the same?
Probably have to rework the req payloads (looks like the network address would no longer be prefixing at least?).
Does it work for `NETWORK_ADDRESS_REQUEST` too (just wondering since it has to use the ieee from the zdo payload and not dstaddr from data req)?
Something like this would work out-of-the-box?
```typescript
// await this.znp.requestZdo(clusterId, payload, waiter?.ID);
await this.dataRequest(networkAddress, Zdo.ZDO_ENDPOINT, Zdo.ZDO_ENDPOINT, clusterId, Constants.AF.DEFAULT_RADIUS, payload, 15000);
```

Responses would still come in as AREQ/ZDO?

src/adapter/zboss/adapter/zbossAdapter.ts
```diff
@@ -54,56 +58,45 @@ export class ZBOSSAdapter extends Adapter {
...
-        }
-        if (frame.type == FrameType.INDICATION && frame.commandId == CommandId.ZDO_DEV_ANNCE_IND) {
-            logger.debug(`Device join request received: ${frame.payload.nwk} ${frame.payload.ieee}`, NS);
-            const payload: DeviceJoinedPayload = {
-                networkAddress: frame.payload.nwk,
-                ieeeAddr: frame.payload.ieee,
-            };
-
-            this.emit('deviceJoined', payload);
-        }
```
kirovilya, line 87: the device join handler is forgotten. if we do not process ZDO_DEV_ANNCE_IND, then we will process ZDO_DEV_AUTHORIZED_IND
```
            } else if (frame.commandId == CommandId.ZDO_DEV_AUTHORIZED_IND) {
                const payload: DeviceJoinedPayload = {
                    networkAddress: frame.payload.nwk,
                    ieeeAddr: frame.payload.ieee,
                };

                this.emit('deviceJoined', payload);
```

Nerivec: I made a few tweaks according to:
https://github.com/espressif/esp-zboss-lib/blob/2abb29f3a1934ac48bf58b2245dc18906359c97d/include/zboss_api_zdo.h#L1237-L1371
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/zboss/3.11.2.1/structzb__zdo__signal__device__update__params__s.html
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/zboss/3.11.2.1/structzb__zdo__signal__device__authorized__params__s.html
https://developer.nordicsemi.com/nRF_Connect_SDK/doc/zboss/3.11.2.1/group__zb__comm__signals.html#zdo_authorization_type

Should fit better?
_Looks a lot like EmberZNet logic._

kirovilya: Sorry. it's bad. There is a different frame structure here
https://cloud.dsr-corporation.com/index.php/s/BAn4LtRWbJjFiAm
image

kirovilya:
```
zh:zboss:uart: <-- FRAME: 00021502656a8a19004b1202698601
zh:zboss: processMessage: {"version":0,"type":2,"commandId":533,"tsn":0,"payload":{"ieee":"0x02124b00198a6a65","nwk":38088,"status":1}}
zh:zboss: Device 0x02124b00198a6a65:38088 UNSECURED_JOIN undefined via undefined.
zh:zboss: Device 0x02124b00198a6a65:38088 was denied joining via undefined (undefined).
```

Nerivec: Yikes! So nordic has the full spec on their website, but it doesn't match what they have in their devices? 😢
Is it an outdated version maybe? Seems strange that espressif would implement it that way too.

I reverted the changes. I'll let you deal with completing the implementation in future PRs 😛

src/adapter/deconz/adapter/deconzAdapter.ts
```diff
@@ -844,6 +593,63 @@ class DeconzAdapter extends Adapter {
...
+            clusterId,
+            srcEndpoint: Zdo.ZDO_ENDPOINT,
+            asduLength: payload.length,
+            asduPayload: payload,
+            txOptions: 0,
+            radius: PARAM.PARAM.txRadius.DEFAULT_RADIUS,
+            timeout: 30,
+        };
+
+        logger.debug(`UNSUPPORTED sendZdo(${ieeeAddress}, ${networkAddress}, ${clusterId}, ${payload}, ${disableResponse})`, NS);
```
Koenkk, line 636: Not sure why there is `UNSUPPORTED` here.

src/adapter/z-stack/znp/znp.ts
Koenkk: As discussed, let's keep doing the native AREQ requests as other things might break once we start to use `DATA_REQUEST`

src/adapter/z-stack/adapter/zStackAdapter.ts
Koenkk: It doesn't, because the native requests are still being used

src/adapter/z-stack/adapter/zStackAdapter.ts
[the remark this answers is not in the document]
Koenkk: Yes let's add it to `sendZdo` for now.

src/adapter/z-stack/adapter/zStackAdapter.ts
Koenkk: Looks OK

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -308,68 +321,176 @@ class ZStackAdapter extends Adapter {
...
+                    const zeroes = 21 - payload.length;
+                    const finalPayload = Buffer.alloc(payload.length + 2 + zeroes);
+                    finalPayload.writeUInt16LE(networkAddress, 0);
+                    finalPayload.set(payload, 2);
+
+                    payload = finalPayload;
+                    break;
+                }
+
+                case Zdo.ClusterId.NETWORK_ADDRESS_REQUEST: {
```
Koenkk, line 450:
```suggestion
                case Zdo.ClusterId.IEEE_ADDRESS_REQUEST:
                case Zdo.ClusterId.NETWORK_ADDRESS_REQUEST: {
```

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -308,68 +321,176 @@ class ZStackAdapter extends Adapter {
...
+                    finalPayload.writeUInt16LE(networkAddress === 0 ? ZSpec.BroadcastAddress.DEFAULT : networkAddress, 1);
+                    finalPayload.set(payload, 3);
+
+                    payload = finalPayload;
+                    break;
+                }
+
+                case Zdo.ClusterId.NWK_UPDATE_REQUEST: {
+                    // extra zeroes for empty nwkManagerAddr if necessary
+                    const zeroes = 9 - payload.length - 1; /* TODO: zstack doesn't have nwkUpdateId? */
```
Koenkk, line 427:
```suggestion
                    const zeroes = 9 - payload.length - 1; /* zstack doesn't have nwkUpdateId? */
```

Nope:

Screenshot 2024-09-21 at 14 37 55

Koenkk:
@devbis @fairecasoimeme please test the ZiGate adapter with the latest dev branch to ensure it will not break with the next release.

> A round of permit join + join device (interview) + network map + remove device should pretty much trigger all involved ZDOs.

Changes will be available in the [dev branch](https://www.zigbee2mqtt.io/advanced/more/switch-to-dev-branch.html) in a few hours from now.

devbis:
@Nerivec @Koenkk  
For zigate this code is broken: 
https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/zigate/adapter/zigateAdapter.ts#L238C32-L240

`z2m: Request 'zigbee2mqtt/bridge/request/device/configure_reporting' failed with error: 'Bind 0xa4c1381d13758a9b/1 haElectricalMeasurement from '0x00158d0003976ca2/1' failed (offset is out of bounds)'`

```
payload.length = 21
zeroes = -6 (!)
```

the payload here (in hex) was "a4c1381d13758a9b010b040300158d0003976ca201"

I suppose that the problem is in the ADDRESS_WITH_TYPE_DEPENDENCY field. It can be 2 or 8 bytes long. 
Changing the buffer length to 21 works in this case.

devbis:
@Nerivec 
Sorry to deliver feedback by portions but I don't have an opportunity to test it thoroughly at the moment.
Removing devices also doesn't work for zigate. 

```js

[2024-09-24 10:55:17] info: 	z2m: Removing device '0xa4c1381d13758a9b' (block: false, force: false)
[2024-09-24 10:55:17] debug: 	zh:zigate:driver: ZDO LEAVE_REQUEST(cmd code: 76) 5696a4c1381d13758a9b0000
[2024-09-24 10:55:17] debug: 	zh:zigate:driver: ZDO {"msgCodeBytes":{"type":"Buffer","data":[0,76]},"msgLengthBytes":{"type":"Buffer","data":[0,12]},"checksumBytes":{"type":"Buffer","data":[183]},"msgPayloadBytes":{"type":"Buffer","data":[86,150,164,193,56,29,19,117,138,155,0,0]},"rssiBytes":{"type":"Buffer","data":[]},"msgLengthOffset":0}
[2024-09-24 10:55:17] debug: 	zh:zigate:driver: <-- ZDO send command 0102104c0210021cb75696a4c1381d13758a9b0210021003
[2024-09-24 10:55:17] debug: 	zh:zigate:frame: {"msgCodeBytes":{"type":"Buffer","data":[128,0]},"msgLengthBytes":{"type":"Buffer","data":[0,9]},"checksumBytes":{"type":"Buffer","data":[13]},"msgPayloadBytes":{"type":"Buffer","data":[200,0,0,76,0,0,0,0]},"rssiBytes":{"type":"Buffer","data":[0]},"msgLengthOffset":-1}
[2024-09-24 10:55:17] debug: 	zh:zigate:driver: --> parsed frame >>>> Status 0x8000 <<<<
[2024-09-24 10:55:17] debug: 	zh:zigate:object: Last bytes of data were not parsed c8 00 00 4c 00 00 00 00
[2024-09-24 10:55:17] debug: 	zh:zigate:driver: {"status":200,"sequence":0,"packetType":76}
[2024-09-24 10:55:27] error: 	z2m: Request 'zigbee2mqtt/bridge/request/device/remove' failed with error: 'Failed to remove device '0xa4c1381d13758a9b' (block: false, force: false) (Error: {"clusterId":32820,"target":22166} after 10000ms)'
[2024-09-24 10:55:27] debug: 	z2m: Error: Failed to remove device '0xa4c1381d13758a9b' (block: false, force: false) (Error: {"clusterId":32820,"target":22166} after 10000ms)
    at Bridge.removeEntity (/Users/ivan.belokobylskiy/projects/zigbee2mqtt/lib/extension/bridge.ts:718:19)
    at Bridge.deviceRemove (/Users/ivan.belokobylskiy/projects/zigbee2mqtt/lib/extension/bridge.ts:259:16)
    at Bridge.onMQTTMessage (/Users/ivan.belokobylskiy/projects/zigbee2mqtt/lib/extension/bridge.ts:200:34)
    at EventEmitter.wrappedCallback (/Users/ivan.belokobylskiy/projects/zigbee2mqtt/lib/eventBus.ts:206:17)
```
    
Additionally, I found out that clusters for the device are read in wrong endian:

(from database.json)

```json
      "inClusterList": [
        0,
        768,
        256,
        1284,
        516,
        10756,
        12292,
        12548
      ],
      "outClusterList": [
        6400
      ],
```

6400 = 0x1900 should be 0x0019 = OTA (and others are reversed as well). I suppose, reading from the endpoint is parsed incorrectly

PR #1190, opened 2024-09-15
fix!: Fix always use `number` for `reportableChange`

Koenkk:
`reportableChange` can be a `number` or `[number, number]` (to support larger ints). However since `int40`, `uint48` and `uint40` easily fit in a Node `Number.MAX_SAFE_INTEGER` there is no need to use `[number, number]`.

```
uint48 max       = 281474976710655
MAX_SAFE_INTEGER = 9007199254740991
```

(u)int56 and (u)int64 could be supported through a `BigInt`, but in practice this will never be used.

This fixes https://github.com/Koenkk/zigbee2mqtt/issues/23711

TODO:
- [x] Create PR for zhc (pay special attention to Xiaomi struct parsing) (https://github.com/Koenkk/zigbee-herdsman-converters/pull/7996)


Nerivec (approved):
Well, that fixes a leftover TODO I had marked when we moved/revamped the spec 👍

_Probably should do the BigInt too (in buffalo.ts directly), if only to standardize the various parsers (currently inconsistent, as marked by the other TODO in buffaloZcl.ts). The `n`-append notation (`123n`) makes it easy enough to keep mostly plain numbers in the code (instead of `BigInt(123)`).
Note if ever needed: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/BigInt#use_within_json_

Koenkk:
@Nerivec Done!

Nerivec:
Made for a very nice cleanup in ZHC!

test/buffalo.test.ts
```diff
@@ -158,6 +158,68 @@ describe('Buffalo', () => {
...
+        const buffalo = new Buffalo(Buffer.from([254, 255, 255, 255, 255, 255, 255]));
+        const value = buffalo.readUInt56();
+        expect(buffalo.getPosition()).toEqual(7);
+        expect(value).toStrictEqual(72057594037927934n);
+    });
+
+    it.each([
+        [36028797018963964n, [0, 0, 252, 255, 255, 255, 255, 255, 127]],
+        [-36028797018963964n, [0, 0, 4, 0, 0, 0, 0, 0, 128]],
+    ])('INT56 write positive', (value: bigint, expectedValues: number[]) => {
```
Nerivec, line 178: Remove `positive` in label?

src/buffalo/buffalo.ts
```diff
@@ -91,6 +91,30 @@ class Buffalo {
         return value;
     }
 
+    public writeUInt56(value: bigint): void {
+        this.buffer.writeUIntLE(Number(value & 0xffffffffffffn), this.position, 6);
+        this.buffer.writeUint8(Number(value >> 48n), this.position + 6);
```
Nerivec, line 96: Being picky: should use `writeUIntX` format for consistency across the file (couple of occurrences use `writeUintX`). 😛

test/buffalo.test.ts
Koenkk: Fixed!

src/buffalo/buffalo.ts
```diff
@@ -157,6 +181,38 @@ class Buffalo {
         return value;
     }
 
+    public writeInt56(value: bigint): void {
+        let unsignedValue = value;
```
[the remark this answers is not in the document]
Koenkk: Fixed!

Issue #940, opened 2024-02-27
[Proposal] Removal of "forever" Permit Join


Nerivec:
After observing behaviors while testing `ember` and seeing reports here and there about this, I think removing the possibility to open the network "forever" via internal workarounds would make sense.

1) It doesn't seem to work well in general from Discord reports (compared to triggering as needed with the Permit Join button).
2) Some stacks now enforce no "forever", which messes with the interval refresh triggered by Z2M, basically breaking the "assumed state".
3) It prevents state validation, because of the interval refresh that won't actually do anything if the network is already open (at least in EmberZNet)
4) It's bad for security...

The TODOs on this would appear to be:
- Remove the config 
- Cleanup the config validations
- Remove the interval refresh logic to only have:
  - Open for 254 seconds (_0xFF can be a special value_) on button click.
  - Close on button click if within 254 seconds of opening.
  - Optionally use stack-level function/prop to determine if network is currently opened instead of assuming or using an internal timer (not sure if all stacks support this)
- Remove/fix related tests

Related code (probably not all...):

https://github.com/Koenkk/zigbee-herdsman/blob/b004cf827da04eb78e7658776bd2726289797b16/src/controller/controller.ts#L237

https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/data/configuration.yaml#L5
https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/lib/types/types.d.ts#L122
https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/lib/util/settings.schema.json#L48
https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/lib/util/settings.ts#L39
https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/lib/controller.ts#L139
https://github.com/Koenkk/zigbee2mqtt/blob/f04ade6d764518bd8228eb32a15e96cd4e3fbde6/lib/extension/bridge.ts#L161

Koenkk:
[...]

PR #1194, opened 2024-09-21
feat!: Standardize ZDO.

Nerivec:
Move ZDO requests out of Adapter to standardize execution.
All sending of standard ZDO requests is done via `sendZdo`.
ZDO responses are emitted from Adapter to Controller, where common logic can be executed as needed. _Removes the need for separate events._

Also worth mentioning:
- Automatically introduces new features to drivers that previously weren't supporting them (like change channel). _Will need testing however._
- Make use of IEEE Address Response (similar to Network Address Response) in Controller.
- Add support for changes in network address that weren't advertised by the device or missed by Z2M. (supersedes and should fix https://github.com/Koenkk/zigbee-herdsman/pull/1183)
- Fix some previously hidden coverage.
- Fix eslint in `controller.test.ts` for easier future updates.

TODO separate PR:
- [x] Move ZDO requests to upstream code
- [x] Remove `deviceAnnounce` and `networkAddress` event from `Adapter`
- [x] Remove old functions from `Adapter` & impl.
- [x] Fix tests
  - [x] `Controller`
  - [x] `ember`
  - [x] `zstack`
- [ ] Deal with `permitJoin` special-case (later PR? in conjunction with https://github.com/Koenkk/zigbee-herdsman/issues/940)
- [ ] Verify compatibility with zigbee2mqtt tests


Nerivec:
Regarding https://github.com/Koenkk/zigbee-herdsman/pull/1183 (and accompanying issues)
Logic implemented here is as follow:
On reception of a ZCL payload, if the attached network address does not correspond to a device known to Z2M, send a request for the IEEE that goes with that network address. As a result, this will trigger a ZDO response that will update the network address of the device matching that IEEE, if Z2M knows it, otherwise, the ZCL payload is skipped (as before).
That should hopefully cover these cases (see included tests):
- device changed network address but didn't advertise it
- device changed network address but Z2M missed it (offline)

src/controller/controller.ts
```diff
@@ -511,44 +530,96 @@ class Controller extends events.EventEmitter<ControllerEventMap> {
...
-        if (device.networkAddress !== payload.networkAddress) {
-            logger.debug(`Device '${payload.ieeeAddr}' announced with new networkAddress '${payload.networkAddress}'`, NS);
-            device.networkAddress = payload.networkAddress;
+        if (device.networkAddress !== payload.nwkAddress) {
+            logger.debug(`Device '${payload.eui64}' announced with new networkAddress '${payload.nwkAddress}'`, NS);
+            device.networkAddress = payload.nwkAddress;
             device.save();
+
+            // TODO: this is not emitting `deviceNetworkAddressChanged`?
+            // this.selfAndDeviceEmit(device, 'deviceNetworkAddressChanged', {device});
```
Koenkk, line 622: Nice catch! It should emit an event indeed.

Koenkk:
> Verify compatibility with zigbee2mqtt tests

Zigbee herdsman is completely stubbed in z2m and since the interface didn't change this should be OK.

PR #1071, opened 2024-05-20
feat: add auto-detection of CC2652P

tobiasdiez:
Add detection of the CC2652P (from https://item.taobao.com/item.htm?_u=j20f1ocri1b58f&id=673052067615&spm=a1z09.2.0.0.6c5f2e8dlM2Hvh or https://nl.aliexpress.com/item/1005006155691159.html?gatewayAdapt=glo2nld).

On my system, the reported information was in capitals so I also changed the `equalsPartial` method to be case-insensitive.

PR #1197, opened 2024-09-22
feat!: Improved adapter discovery.

Nerivec:
TODO:
- [x] Update `Adapter.create` (_should remove most of the logic_)
- [x] Remove `adapter.autoDetectPath`, `adapter.isValidPath` and definitions in all drivers
- [ ] Complete fingerprints. Closes https://github.com/Koenkk/zigbee-herdsman/pull/1071


Nerivec:
@kirovilya Can you provide a fingerprint for the nordic adapter (from executing `SerialPort.list()`)? I'll clean up the matching/tests with it.

kirovilya:
> @kirovilya Can you provide a fingerprint for the nordic adapter? I'll clean up the matching/tests with it.

for this? https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/zboss/adapter/zbossAdapter.ts#L24

Nerivec:
Whatever comes out of the bit of code I mentioned above should have it all; mostly the path. This PR improves upon the discovery a bit, more refined using regex matching against the paths.

kirovilya:
> Whatever comes out of the bit of code I mentioned above should have it all; mostly the path. This PR improves upon the discovery a bit, more refined using regex matching against the paths.
> 
> See the `adapterDiscovery.ts` file in the PR, in case you can provide more missing fingerprints 😉

```
{
    manufacturer: 'ZEPHYR',
    serialNumber: '54ACCFAFA6DADC49',
    pnpId: 'usb-ZEPHYR_Zigbee_NCP_54ACCFAFA6DADC49-if00',
    locationId: undefined,
    vendorId: '2fe3',
    productId: '0100',
    path: '/dev/ttyACM0'
  },
```
  
  and in windows
  ```

    {
    path: 'COM22',
    manufacturer: 'Корпорация Майкрософт',
    serialNumber: '6&11E6BDCD&1&0000',
    pnpId: 'USB\\VID_2FE3&PID_0100&MI_00\\6&11E6BDCD&1&0000',
    locationId: '0000.0014.0000.009.000.000.000.000.000',
    friendlyName: 'Устройство с последовательным интерфейсом USB (COM22)',
    vendorId: '2FE3',
    productId: '0100'
  },
```

**Comment by kirovilya** [comment]
there is also a proposal to extract the serial port from the adapters and transfer the ready-made connection inside.
connection and reconnection control can also be performed outside the adapter

**Comment by Nerivec** [comment]
It would require adding more variables to `Adapter` that can be overridden in subclasses to have the proper settings for serial/socket ports (hw/sw, stop bits, parity, watermark, etc.) since these can vary. Parser/writer would have to stay in adapters too, since these are very different in most cases (or would require an even bigger refactoring).
Probably wouldn't be bad to have just the writer instance and a parser callback function passed to adapters, but at first glance, it's a pretty large refactor.

File: src/adapter/adapterDiscovery.ts [hunk 1]
```diff
@@ -0,0 +1,430 @@
...
+        if (path.startsWith('mdns://')) {
+            return await findmDNSAdapter(path);
+        } else if (path.startsWith('tcp://')) {
+            return await findTCPAdapter(path, adapter);
+        } else if (adapter && adapter !== 'auto') {
+            const matched = await matchUSBAdapter(adapter, path);
+
+            /* istanbul ignore else */
+            if (!matched) {
+                logger.error(`Unable to match USB adapter: ${adapter} | ${path}`, NS);
```
Koenkk, line 415: In case both adapter and path is specified, why do we need to do any matching and log an error when it doesn't? I expect that setting both adapter and path wouldn't do any auto detection at all.

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,430 @@
...
+            // /dev/serial/by-id/usb-dresden_elektronik_ingenieurtechnik_GmbH_ConBee_II_DE2132111-if00
+            pathRegex: '.*conbee.*',
+        },
+        {
+            // Conbee III
+            vendorId: '0403',
+            productId: '6015',
+            manufacturer: 'dresden elektronik ingenieurtechnik GmbH',
+            // /dev/serial/by-id/usb-dresden_elektronik_ConBee_III_DE03188111-if00-port0
+            pathRegex: '.*conbee.*',
```
Koenkk, line 40:
```suggestion
            pathRegex: '.*conbee_iii_.*',
```

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,430 @@
...
+ */
+const USB_FINGERPRINTS: Record<DiscoverableUSBAdapter, USBAdapterFingerprint[]> = {
+    deconz: [
+        {
+            // Conbee II
+            vendorId: '1cf1',
+            productId: '0030',
+            manufacturer: 'dresden elektronik ingenieurtechnik GmbH',
+            // /dev/serial/by-id/usb-dresden_elektronik_ingenieurtechnik_GmbH_ConBee_II_DE2132111-if00
+            pathRegex: '.*conbee.*',
```
Koenkk, line 32:
```suggestion
            pathRegex: '.*conbee.*_ii_',
```

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,430 @@
...
+            // /dev/serial/by-id/usb-Silicon_Labs_CP2102N_USB_to_UART_Bridge_Controller_a215650c853bec119a079e957a0af111-if00-port0
+            pathRegex: '.*slzb-07.*',
+        },
+        {
+            // Sonoff ZBDongle-E V2
+            vendorId: '1a86',
+            productId: '55d4',
+            manufacturer: 'ITEAD',
+            // /dev/serial/by-id/usb-ITEAD_SONOFF_Zigbee_3.0_USB_Dongle_Plus_V2_20240122184111-if00
+            pathRegex: '.*sonoff.*plus.*',
```
Koenkk, line 89:
```suggestion
            pathRegex: '.*sonoff.*plus.*_v2.*',
```

To prevent it from being the same as the Dongle-P

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,430 @@
...
+
+            return [adapter, path];
+        }
+    }
+
+    // default to matching USB
+    const match = await findUSBAdapter(path);
+
+    if (!match) {
+        throw new Error(`Unable to find a valid USB adapter.`);
```
Koenkk, line 426:
```suggestion
        throw new Error(`Unable to find a valid USB adapter. Specify the adapter and path in your configuration.`);
```

src/adapter/adapterDiscovery.ts
Nerivec: It acts as "pre-validation" of the config.
It used to do the same before, but only to pre-validate `path`.
https://github.com/Koenkk/zigbee-herdsman/blob/c661d2ea744ddd80b4d0b37c6b54dec8cb6324e3/src/adapter/adapter.ts#L170-L182

It's not "strictly" necessary.

src/adapter/adapterDiscovery.ts
Nerivec: That was on purpose, `V2` is not always present in Dongle-E neither. (Just haven't gathered all possible combinations for the comments yet.)
But it doesn't matter since it can't conflict due to different `vendorId`+`productId`.

src/adapter/adapterDiscovery.ts
Koenkk: Then I propose to log it as debug, if auto detect doesn't work (which it doesn't with the HA addon for example) users will always see this error which can be confusing.

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,439 @@
...
+        {
+            // CC26X2R1
+            vendorId: '0451',
+            productId: 'bef3',
+            manufacturer: 'Texas Instruments',
+            pathRegex: '.*CC26X2R1.*', // TODO
+        },
+        {
+            // SMLight slzb-07p7
+            vendorId: '',
```
Nerivec, line 157: @darkxst Can you help in completing some of these fingerprints if you have some time?

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,439 @@
...
+        // port info is missing essential information for proper matching, ignore it
+        return;
+    }
+
+    for (const entry of entries) {
+        if (
+            matchString(portInfo.vendorId, entry.vendorId) &&
+            matchString(portInfo.productId, entry.productId) &&
+            (!entry.manufacturer || !portInfo.manufacturer || matchString(portInfo.manufacturer, entry.manufacturer) || isWindows) &&
+            (!entry.pathRegex || matchRegex(entry.pathRegex, portInfo.path) || matchRegex(entry.pathRegex, portInfo.pnpId) || isWindows)
```
Koenkk, line 270: Currently both the vendorId, productId and path should match, however I would argue that a match on either vendorId + productId OR path should be enough:
- Vendor + productID might be unknown (e.g. insufficient right for udev), then we can still match on the path
- Vendor + productID is known, but the path is not a `serial-by-id`, 2 cases I see here:
  - User uses e.g. `/dev/ttyUSB0`
  - Customised device name with Docker when running with `--device /dev/serial/by-id/usb-Silicon_Labs_CP2102N_USB_to_UART_Bridge_Controller_de4374fdea93eb11b514a85b3d98b6d1-if00-port0:/dev/adapter` (which is what I do)
  -

Nerivec: It would, but it wouldn't work... Cases where path might be the same (except for id hex) would match (like some Dongle-E vs some Dongle-P, only way to differentiate is with vendorId/productId).

darkxst: > (like some Dongle-E vs some Dongle-P, only way to differentiate is with vendorId/productId).

Dongle-E switched to cp2102 due to high failure rates in ch9102. The initial batches with cp2102 were flashed with the Dongle-P usb config, so are indistinguishable from Dongle-P. I believe this has been corrected now with latest devices, but it leaves many devices that can only be discovered by probing firmware  :(

src/adapter/adapterDiscovery.ts
darkxst: This is standard cp2102 usb id: `10c4:ea60`
Manufacturer: `SMLIGHT`
Product: `SMLIGHT_SLZB-07p7`

Later when p10 is available it will just change product string
Product: `SMLIGHT_SLZB-07p10`

src/adapter/adapterDiscovery.ts
Nerivec: Yes USB discovery definitely has its limits. Between mistakes, lack of details, various versions... We really can only "try to match as many as possible", the rest requires manual configuration.

src/adapter/adapterDiscovery.ts
darkxst: Also Mg24 is out now.

product string: `SMLIGHT_07Mg24` all other details same as SLZB-07

src/adapter/adapterDiscovery.ts
```diff
@@ -0,0 +1,486 @@
...
+        {
+            // SMLight slzb-07p7
+            vendorId: '10c4',
+            productId: 'ea60',
+            manufacturer: 'SMLIGHT',
+            // /dev/serial/by-id/usb-SMLIGHT_SMLIGHT_SLZB-07p7_be9faa0786e1ea11bd68dc2d9a583111-if00-port0
+            pathRegex: '.*SLZB-07p7.*',
+        },
+        {
+            // TubesZB ?
```
Nerivec, line 172: @tube0013 Can you take a look at the TubesZB fingerprints and confirm the data? If you can also provide the fingerprints for your other devices, that'd be great.

tube0013: I'm going to be adopting the generic zeroconf/mdns discovery coming to HA/ZHA: https://github.com/home-assistant/core/pull/126294

Nerivec: These are for USB. You don't have any USB device? I found a couple of fingerprints in zigpy code, but maybe they are no longer relevant?

tube0013: my bad I should read, this info is still valid: https://github.com/home-assistant/core/blob/dcb6c9a13375d5db73211f250a52c5342d36307d/homeassistant/components/zha/manifest.json#L51 

through L51.

Nerivec: Did you set the `manufacturer` property as well? (that would add an extra bit of info to match against).

tube0013: I set product string to "TubesZB" and serial is set as 971207DO or 971229NY

At least on the ch340b I'm not aware how to write a manufacturer string.

Koenkk:
Could you target this against the `feat/3.0.0` branch? Then this can be merged.

Nerivec:
Please see
https://github.com/Koenkk/zigbee2mqtt/discussions/24198
https://github.com/Koenkk/zigbee2mqtt/discussions/24364

PR #1198, opened 2024-09-23
zigate: Fix ZDO BIND/UNBIND payload mismatch.

Nerivec:
Fixes https://github.com/Koenkk/zigbee-herdsman/pull/1187#issuecomment-2367768315
Tested against old payloads, should now match exactly.
_Note: It might even be possible to remove the entire alteration for BIND/UNBIND, as long as zigate doesn't try to read the unused endpoint at the end (it shouldn't). Would require testing with actual adapter though._

CC: @devbis


Koenkk:
Thanks!

PR #1201, opened 2024-09-23
fix: Cleanup network address change code.

Nerivec:
Avoid multiple triggering of same logic (with tests).
Fix not waiting for broadcast in `device.updateNetworkAddress`. _Might want to rename that one since it's a side-effect that actually updates it, not the function itself? `requestNetworkAddress`?_


Koenkk:
`requestNetworkAddress` makes more sense indeed

Koenkk:
Thanks!

PR #1206, opened 2024-09-27
fix: On NO_ENTRY error during unbind cleanup database

sjorge:
I'm been unable to trigger the NOT_FOUND error again today while looking at this again, I'm not sure what changed since when we talked on discord.

I've not touched the reporting code where it was triggered. I did rebase before testing and there seem to have been some other changes so something in those might have slightly changed the behavior?

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -621,7 +621,11 @@ class Endpoint extends Entity {
             const response = await Entity.adapter!.sendZdo(this.deviceIeeeAddress, this.deviceNetworkAddress, zdoClusterId, zdoPayload, false);
 
             if (!Zdo.Buffalo.checkStatus(response)) {
-                throw new Zdo.StatusError(response[0]);
+                if (response[0] === Zdo.Status.NO_ENTRY) {
+                    logger.debug(`${log} no entry on device, removing entry from database.`, NS);
+                } else {
+                    throw new Zdo.StatusError(response[0]);
+                }
             }
 
             this._binds.splice(index, 1);
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -7995,6 +7995,34 @@ describe('Controller', () => {
         expect(error).toStrictEqual(new Error(`Unbind 0x129/1 genOnOff invalid target '1' (no group with this ID exists).`));
     });
 
+    it('Unbind against unbound cluster', async () => {
+        await controller.start();
+        await mockAdapterEvents['deviceJoined']({networkAddress: 129, ieeeAddr: '0x129'});
+        await mockAdapterEvents['deviceJoined']({networkAddress: 170, ieeeAddr: '0x170'});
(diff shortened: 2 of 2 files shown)


src/controller/model/endpoint.ts
```diff
@@ -628,9 +628,18 @@ class Endpoint extends Entity {
             this.save();
         } catch (error) {
             const err = error as Error;
-            err.message = `${log} failed (${err.message})`;
-            logger.debug(err.stack!, NS);
-            throw error;
+            if ((error as Zdo.StatusError).code == 136) {
```
Koenkk, line 631:
```suggestion
            if ((error as Zdo.StatusError).code == Status.NO_ENTRY) {
```

(pick it from the `zdo/definition/status.ts`)

Koenkk:
Didnt notice this on my initial review, but the error is thrown on line 624, I think it would be easier to just not throw the error instead, ie extend the check on line 623 to ignore NO_ENTRY

sjorge:
> Didnt notice this on my initial review, but the error is thrown on line 624, I think it would be easier to just not throw the error instead, ie extend the check on line 623 to ignore NO_ENTRY

So check if `response[0] == Zdo.Status.NO_ENTRY` and skip if it is?

Nerivec:
Something like this I'd say?
```typescript
            if (!Zdo.Buffalo.checkStatus(response)) {
                if (response[0] === Zdo.Status.NO_ENTRY) {
                    logger.debug(`${log} no entry to remove. Removing bind from database.`, NS);
                } else {
                    throw new Zdo.StatusError(response[0]);
                }
            }
```
It'll execute the rest of the logic as if the bind was removed.

sjorge:
> Something like this I'd say?
> 
> ```ts
>             if (!Zdo.Buffalo.checkStatus(response)) {
>                 if (response[0] === Zdo.Status.NO_ENTRY) {
>                     logger.debug(`${log} no entry to remove. Removing from database.`, NS);
>                 } else {
>                     throw new Zdo.StatusError(response[0]);
>                 }
>             }
> ```
> 
> It'll execute the rest of the logic as if the bind was removed.

That was close to what I had, it just wasn't 100% clear it was  response[0] that held the response code.

Nerivec:
Are you using an IDE that supports Typescript (like vscode)? ZDO stuff should all be strongly typed (and automatically inferred from `clusterId` parameters).
https://github.com/Koenkk/zigbee-herdsman/blob/5d02efe44826401521b85259dde5e3ca4421e312/src/zspec/zdo/definition/tstypes.ts#L892-L1062

sjorge:
No, I just use a very barebones vim. Never really bothered with an IDE or even syntax highlighting.

Koenkk:
Thanks!

Nerivec:
> Never really bothered with an IDE or even syntax highlighting.

https://github.dev/Koenkk/zigbee-herdsman/blob/master/src/zspec/zdo/buffaloZdo.ts
The code is so much better in color 😛

PR #1214, opened 2024-10-07
feat!: Improve permit join.

Nerivec:
- Match `Controller` API more closely to `Adapter`: `(time, device?)`
- Clamp time to [0-254] (never forever)
- Remove "forever permit join" (security concerns + sometimes unstable behavior)
  - Fix https://github.com/Koenkk/zigbee-herdsman/issues/940
- Remove `reason` (unused)
- `getPermitJoinTimeout` returns [0-254] range, with 0 being "not permitting joining".
- Remove `getPermitJoin` => use `getPermitJoinTimeout` instead (boolean easily determined on other end of mqtt transaction -see below-)
- For Z2M: Less unnecessary mqtt data required/published in various places (`/bridge/info`, `/request/permit_join`)

TODO:
- [x] https://github.com/Koenkk/zigbee2mqtt/pull/24257
- [ ] Remove unnecessary mqtt publish of `/bridge/info` every time the permit join timer ticks (the other end of mqtt transaction should be responsible for the ticks, and z2m should only publish begin -possibly end for extra sync purposes-).

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -78,7 +78,6 @@ class Controller extends events.EventEmitter<ControllerEventMap> {
     // @ts-expect-error assigned and validated in start()
     private touchlink: Touchlink;
 
-    private permitJoinNetworkClosedTimer: NodeJS.Timeout | undefined;
     private permitJoinTimeoutTimer: NodeJS.Timeout | undefined;
     private permitJoinTimeout: number | undefined;
     private backupTimer: NodeJS.Timeout | undefined;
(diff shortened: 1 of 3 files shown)


src/controller/controller.ts
```diff
@@ -274,62 +274,52 @@ class Controller extends events.EventEmitter<ControllerEventMap> {
...
-                try {
-                    await this.adapter.permitJoin(254, device?.networkAddress);
-                    await this.greenPower.permitJoin(254, device?.networkAddress);
-                } catch (error) {
-                    logger.error(`Failed to keep permit join alive: ${error}`, NS);
+        this.permitJoinTimeout = 0;
+
+        if (time > 0) {
+            // never permit more than uint8, and never permit 255 that is often equal to "forever"
+            time = time > 254 ? 254 : time;
```
Koenkk, line 284: I think we should assert on this, throw an error when the value > 254

Koenkk:
Could you target this against `feat/2.0.0`?

PR #1217, opened 2024-10-17
fix: Log socket errors

Koenkk:
Log the whole error to get more info, also always log them as `error`

https://github.com/zigbee2mqtt/hassio-zigbee2mqtt/issues/644


Nerivec (approved):
👍

PR #1222, opened 2024-10-23
fix!: Fix some transmit power issues after #1139

Nerivec:
https://github.com/Koenkk/zigbee-herdsman/pull/1139 and https://github.com/Koenkk/zigbee2mqtt/pull/23566 changed the way transmit power is set on startup, removing the need for the ZH Controller API `setTransmitPower`. This API was never cleaned up, and some adapters ended up no longer actually setting the transmit power on startup (ezsp, zigate, zboss).

This is technically breaking since the Controller API changes, so, going into v3.

@Koenkk can you confirm zstack is ok with having this triggered directly during startup (instead of through the queue)?

@kirovilya I don't know if zboss has a persistent way to set transmit power during form to skip setting it on every startup (like EmberZNet). I'll let you check and make the changes later, if possible.

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -97,8 +97,6 @@ abstract class Adapter extends events.EventEmitter<AdapterEventMap> {
 
     public abstract getNetworkParameters(): Promise<TsType.NetworkParameters>;
 
-    public abstract setTransmitPower(value: number): Promise<void>;
-
     public abstract addInstallCode(ieeeAddress: string, key: Buffer): Promise<void>;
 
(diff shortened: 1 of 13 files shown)


Koenkk:
I'm not sure if we should remove the `setTransmitPower` as it works fine right? I see some uses cases for changing the transmit power on the fly (without having to restart zigbee-herdsman)

Nerivec:
Changing the transmit power after startup is not a good idea since it disrupts the surroundings of the adapter. It can make the network unstable (break nearby routes, etc.).

PR #380, opened 2021-06-06
Added ELKO attributes

emilhetty:
Added ELKO manufacture specific attributes for varies product

PR #1234, opened 2024-11-10
Add Schneider Electric occupancy cluster and attribute

peroo:
The definitions are duplicated from the Elko definitions added in https://github.com/Koenkk/zigbee-herdsman/pull/380
Strictly speaking I think all the definitions in that PR should be moved to Schneider since Elko is a subsidiary and nothing actually uses them as they are, but that's beyond the scope of this PR.


Koenkk:
Thanks for the PR, but please use [`deviceAddCustomCluster`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/3a1640f775ef12edc0d53ac51f081aa55fe5ecc0/src/devices/amina.ts#L142) instead.

PR #1241, opened 2024-11-18
Fix command responses in the scenes cluster

canardos:
**View scene command**

The `viewRsp` command for the `genScenes` cluster contains 6 properties. These match the ZCL specified fields, but error responses will only contain `status`, `groupid`, and `sceneid`. From the [ZCL spec](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf) (rev.8):

>If the status is SUCCESS, the Transition time, Scene Name and Extension field fields are copied from the corresponding fields in the table entry, otherwise they are omitted.

This results in a buffer overrun in the `parsePayloadCluster` function of the `ZclFrame` during parsing of the incoming response command when not all fields are present.

I am not overly familiar with the library, but looking at other clusters, it appears that the right solution is use of the `conditions` property of the `ParameterDefinition`:

```ts
viewRsp: {
    ID: 1,
    parameters: [
        {name: 'status', type: DataType.UINT8},
        {name: 'groupid', type: DataType.UINT16},
        {name: 'sceneid', type: DataType.UINT8},
        {name: 'transtime', type: DataType.UINT16, conditions: [{type: ParameterCondition.STATUS_EQUAL, value: Status.SUCCESS}]},
        {name: 'scenename', type: DataType.CHAR_STR, conditions: [{type: ParameterCondition.STATUS_EQUAL, value: Status.SUCCESS}]},
        {name: 'extensionfieldsets', type: BuffaloZclDataType.EXTENSION_FIELD_SETS, conditions: [{type: ParameterCondition.STATUS_EQUAL, value: Status.SUCCESS}]},
    ],
},
```

Devices not supporting scene names are required to return the null string so this doesn't need to be accounted for.

**Get scene membership**

As with the prior command, the ZCL (rev.8) states:

> **3.7.2.5.6.1 Get Scene Membership Response Command**
> ...
> If the status is not SUCCESS, then the Scene count and Scene list field are omitted

Adding parameter conditions to these fields:

```ts
getSceneMembershipRsp: {
    ID: 6,
    parameters: [
        {name: 'status', type: DataType.UINT8},
        {name: 'capacity', type: DataType.UINT8},
        {name: 'groupid', type: DataType.UINT16},
        {name: 'scenecount', type: DataType.UINT8, conditions: [{type: ParameterCondition.STATUS_EQUAL, value: Status.SUCCESS}]},
        {name: 'scenelist', type: BuffaloZclDataType.LIST_UINT8, conditions: [{type: ParameterCondition.STATUS_EQUAL, value: Status.SUCCESS}]},
    ],
},
```

**Enhanced Add Scene command**

> **3.7.2.5.7 Enhanced Add Scene Response Command**
> ...
> The payload of this command SHALL be formatted in the same way as the Add Scene Response command specified in the ZCL scenes cluster.

The current implementation has no parameters. Copying from add add scene response:

```ts
enhancedAddRsp: {
    ID: 64,
    parameters: [
        {name: 'status', type: DataType.UINT8},
        {name: 'groupId', type: DataType.UINT16},
        {name: 'sceneId', type: DataType.UINT8},
    ],
},
```

**Enhanced View Scene command**

> **3.7.2.5.8 Enhanced View Scene Response Command**
> ...
> The payload of this command SHALL be formatted in the same way as the View Scene Response command, with the following difference:
> The Transition Time field SHALL be measured in tenths of a second rather than in seconds.

The same parameter conditions as for the View Scene command above are required.


Nerivec (approved):
LGTM 👍

I noticed the attributes in `genBasic` are also a bit out of whack, will need to fix this too.

Koenkk:
Thanks!

PR #1252, opened 2024-12-04
Print source and target as hexadecimals in ember errors.

antst:
This resolves #25024

### DIFF
--- a/src/adapter/ember/ezsp/ezsp.ts
+++ b/src/adapter/ember/ezsp/ezsp.ts
@@ -5336,7 +5336,7 @@ export class Ezsp extends EventEmitter<EmberEzspEventMap> {
      *        for this discovery arrive, but the callback is made only once.
      */
     ezspIncomingManyToOneRouteRequestHandler(source: NodeId, longId: EUI64, cost: number): void {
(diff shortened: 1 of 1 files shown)


Nerivec:
All logging is done in decimal throughout the codebase. This would require a much bigger PR that changes every occurrence (thousands) to not be even more confusing. It adds a sizeable amount of code (those `toString` repeated thousands of times...). And breaks several setups that rely on some logging events to automate things.

Also, frontend was updated to show both hex/dec versions recently.

antst:
> All logging is done in decimal throughout the codebase. This would require a much bigger PR that changes every occurrence (thousands) to not be even more confusing. It adds a sizeable amount of code (those `toString` repeated thousands of times...). And breaks several setups that rely on some logging events to automate things.
> 
> Also, frontend was updated to show both hex/dec versions recently.

Then we need to add to frontend dec search capability.
Because now, every time I need to check what it is, I need to copy from logs, convert in calculator and only then place into search string in frontend. Damn annoying! :)

Nerivec:
True, frontend definitely needs some love. 😉

PR #1256, opened 2024-12-08
ZStack: fix request network address blocking all requests

Koenkk:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/24671


src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -317,7 +316,7 @@ class ZStackAdapter extends Adapter {
         networkAddress: number,
         clusterId: Zdo.ClusterId,
         payload: Buffer,
-        disableResponse: true,
+        disableResponse: boolean,
```
Nerivec, line 311: Switching to this "internal" function breaks proper typing, since that means it allows awaiting for response even though it shouldn't (doesn't have one for given cluster). The 2 overloads are not used anymore for the "internal" function.

PR #1257, opened 2024-12-11
feat: Generic start strategy

Nerivec:
Provide a generic startup strategy for all adapters
- universal strategy, no mismatching behaviors between adapter types
- will help with implementing migration from ZHA
- should remove a lot of duplicate tests (zstack & ember currently have to test this internally)
- `initNetwork` is generic but called internally to adapter, allowing call in proper location (depending on adapter init), or not calling at all for adapters with less features (e.g. deconz, zigate...)

TODO:
- [x] ember
  - [x] Implementation
  - [x] Tests
- [ ] zstack
  - [ ] Implementation
  - [ ] Tests
- [ ] deconz
- [ ] ezsp
- [ ] zboss
- [ ] zigate


Nerivec:
Let's close this. Not being able to thoroughly test it on some stacks makes this overly complicated.

PR #1259, opened 2024-12-11
fix: Proper extended PAN ID in `getNetworkParameters`

Nerivec:
Looks like this was never used for anything other than display, so never caused problems.

- Fix inconsistencies between adapters
- Fix typing
- Use string (like IEEE), never number (can be well beyond MAX_SAFE_INTEGER)


Koenkk:
Thanks!

PR #1268, opened 2024-12-17
fix(ignore): import using `node:` prefix for built-in

Nerivec:


Koenkk:
Does it make sense to enable the [eslint rule](https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/prefer-node-protocol.md)?

Note to self; after this merge https://github.com/Koenkk/zigbee2mqtt/pull/25248

Nerivec:
VSCode kept bugging on this merge conflict for some reason. Looks ok now though.
The rule would be good since it's the expected scheme for the future from what I gather... but it does require the dep (should be an eslint built-in... maybe wait until it is?).

PR #1270, opened 2024-12-19
feat: Use vitest for testing

Nerivec:
https://vitest.dev/guide/why.html

Reasoning:
- More modern approach (TS, ESM...).
  - Jest is annoying (at best) with ESM, clearly something that will be more and more of a problem
- API is almost compatible with Jest, required less refactoring (than going to Node built-in for example, where several things would need complete refactoring)
- Support for [snapshots](https://vitest.dev/guide/snapshot.html)
- Support for [bench](https://vitest.dev/api/#bench) (might be good to have something like this in CI with automatic reporting in PRs, at least for critical logic)
- Support for [types testing](https://vitest.dev/guide/testing-types.html#testing-types)
- Can use [`@vitest/ui`](https://vitest.dev/guide/ui.html) locally if needed
- Similar to above, can use [vscode extension](https://marketplace.visualstudio.com/items?itemName=vitest.explorer) (this one is definitely a great addition!)
- Faster, cleaner (lot less deps)

Related changes:
- Fix couple of ZNP tests that weren't behaving as expected (testing timeouts)
- Fix CONNECTED state not cleared in `closePort()` in ASH protocol for ember (only has an impact in tests due to various bypasses)
- Cleanup coverage ignore (many stale ones that were never cleared)
- Fix inconsistent mocking of date in controller (and cleanup some stale mocks).


package.json
```diff
@@ -81,28 +75,15 @@
...
         "eslint": "eslint --max-warnings=0",
         "pretty:write": "prettier --write .",
         "pretty:check": "prettier --check .",
         "typedoc": "typedoc --gitRevision $(git describe --tag --abbrev=0) --tsconfig tsconfig.json --excludePrivate --excludeProtected --excludeExternals --entryPointStrategy expand ./src --sourceLinkTemplate 'https://github.com/Koenkk/zigbee-herdsman/blob/{gitRevision}/{path}#L{line}' -out typedoc",
         "clean": "rimraf temp coverage dist",
         "prepack": "pnpm run clean && pnpm run build"
     },
-    "version": "3.0.4",
-    "jest": {
-        "coverageThreshold": {
```
Koenkk, line 96: How is this now defined?

Nerivec: In https://github.com/Nerivec/zigbee-herdsman/blob/vitest/test/vitest.config.mts (I used the shorthand that puts it all at 100 for coverage.)

PR #1274, opened 2024-12-27
chore: remove manuSpecificNodOnPilotWire that was moved to zigbee-herdsman-converters

vmeurisse:
Previous logic stopped on the first cluster with no manufacturerCode.

Fixes Koenkk/zigbee2mqtt#25333

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -5095,20 +5095,6 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
         },
         commandsResponse: {},
     },
-    manuSpecificNodOnPilotWire: {
-        ID: 0xfc00,
-        manufacturerCode: ManufacturerCode.NODON,
-        attributes: {
-            mode: {ID: 0x0000, type: DataType.UINT8},
-        },
-        commands: {
-            setMode: {
-                ID: 0x0000,
-                parameters: [{name: 'mode', type: DataType.UINT8}],
-            },
-        },
-        commandsResponse: {},
-    },
     manuSpecificProfalux1: {
         ID: 0xfc21, // Config cluster, 0xfc20 mostly for commands it seems
         manufacturerCode: ManufacturerCode.PROFALUX,
--- a/src/zspec/zcl/definition/tstype.ts
+++ b/src/zspec/zcl/definition/tstype.ts
@@ -233,6 +233,5 @@ export type ClusterName =
     | 'zosungIRControl'
     | 'manuSpecificAssaDoorLock'
     | 'manuSpecificDoorman'
-    | 'manuSpecificNodOnPilotWire'
     | 'manuSpecificProfalux1'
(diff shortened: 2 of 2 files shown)


Nerivec:
This looks like a very specific case of custom cluster ID clashing?
Can't we shuffle the Clusters object instead of de-optimizing the lookup (re-order `manuSpecificNodOnPilotWire` before `manuSpecificAssaDoorLock`)?
This new code will force lookup through the entire Clusters object every time, considering the amount of times this is called, de-optimizing it will have a rather large negative impact.

Nerivec:
Actually... @Koenkk can't we move these two (and others that may conflict) to custom clusters directly and remove the problem entirely?

vmeurisse:
On a Raspberry PI 4, the function takes 0.003ms for an empty list and 0.015ms to loop over the 132 clusters, measured on the average of 20 executions. I removed the break to make sure to loop over the complete list.

If perf is an issue, it might be better to preprocess the list on start to group if by id, as it is less error prone than relying on the cluster order in the source code.

<details>
<summary>Code used</summary>
```js
function findClusterNameByID(id, manufacturerCode, clusters) {                                                               
    console.log(Object.keys(clusters).length);                                                                               
    let time = process.hrtime();                                                                                             
    let matchingManufacturer;                                                                                                
    let noManufacturer;                                                        
    let anyManufacturer;                                      
    for (const clusterName in clusters) {                     
        const cluster = clusters[clusterName];                
        if (cluster.ID === id) {                              
            // When a manufacturerCode is requested, match the first cluster with matching manufacturerCode,
            if (manufacturerCode && cluster.manufacturerCode === manufacturerCode) {                        
                matchingManufacturer = clusterName;                                                         
                // Otherwise, match the first cluster with no manufacturerCode,                             
            }                                                                                               
            else if (!noManufacturer && !cluster.manufacturerCode) {                                        
                noManufacturer = clusterName;                                                               
                // Finally, match the first cluster with a matching ID.                                     
            }                                                                                               
            else if (!anyManufacturer) {                                                                    
                anyManufacturer = clusterName;                                      
            }                                                                  
        }                                                                      
    }                                                                  
    const exactMatch = (manufacturerCode && matchingManufacturer) || (!manufacturerCode && noManufacturer);
    const clusterName = matchingManufacturer || noManufacturer || anyManufacturer;                         
    time = process.hrtime(time);                                                                           
    console.log('benchmark took %d seconds and %d nanoseconds', time[0], time[1]);                         
    return [clusterName, !exactMatch];                                                                     
}
```
</details>

Koenkk:
Yes it would be better to completely remove the clusters here (at least one of them) and move them into zhc. Example on how to do that: https://github.com/Koenkk/zigbee-herdsman-converters/blob/20c0958f6e99343040b334898de494fb97f18aa8/src/devices/sonoff.ts#L1105

vmeurisse:
I moved the cluster definition in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8535

Koenkk:
Thanks!

PR #1280, opened 2024-12-30
fix(modify): Repair the issue of channel updating

CubeZ2mDeveloper:
When I update channels in zigbee2mqtt and restart, I find that some devices fail to be controlled. I discovered that when changing channels, the value of the nwkUpdateId field remains at 0.

I referred to the ZigBee specification and found the instructions regarding the modification of the nwkUpdateId field when changing channels:

"The network manager should broadcast a Mgmt_NWK_Update_req notifying devices of the new channel. The broadcast shall be to all devices with RxOnWhenIdle equal to TRUE. The network manager is responsible for incrementing the nwkUpdateId parameter from the NIB and including it in the Mgmt_NWK_Update_req. The network manager shall set a timer based on the value of apsChannelTimer upon issue of a Mgmt_NWK_Update_req that changes channels and shall not issue another such command until this timer expires. However, during this period, the network manager can complete the above analysis. However, instead of changing channels, the network manager would report to the local application using Mgmt_NWK_Update_notify and the application can force a channel change using the Mgmt_NWK_Update_req."

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -74,5 +74,5 @@ export interface NetworkParameters {
     panID: number;
     extendedPanID: string; // `0x${string}` same as IEEE address
     channel: number;
-    nwkUpdateID?: number; 
+    nwkUpdateID?: number;
 }
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -74,4 +74,5 @@ export interface NetworkParameters {
     panID: number;
     extendedPanID: string; // `0x${string}` same as IEEE address
     channel: number;
+    nwkUpdateID?: number; 
 }
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -74,5 +74,5 @@ export interface NetworkParameters {
     panID: number;
     extendedPanID: string; // `0x${string}` same as IEEE address
     channel: number;
-    nwkUpdateID?: number;
+    nwkUpdateID: number;
 }
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -144,13 +144,7 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
             const netParams = await this.getNetworkParameters();
             const configuredChannel = this.options.network.channelList[0];
             const adapterChannel = netParams.channel;
-            // According to the Zigbee specification:
-            // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req.
-            // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
-            let nwkUpdateID = netParams.nwkUpdateID ?? 0;
-            if (++nwkUpdateID > 0xff) {
-                nwkUpdateID = 0x00;
-            }
+            const nwkUpdateID = netParams.nwkUpdateID;
 
             if (configuredChannel != adapterChannel) {
                 logger.info(`Configured channel '${configuredChannel}' does not match adapter channel '${adapterChannel}', changing channel`, NS);
@@ -505,6 +499,13 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
      */
     private async changeChannel(oldChannel: number, newChannel: number, nwkUpdateID: number): Promise<void> {
         logger.warning(`Changing channel from '${oldChannel}' to '${newChannel}'`, NS);
+
+        // According to the Zigbee specification:
+        // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req.
+        // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
+        if (++nwkUpdateID > 0xff) {
+            nwkUpdateID = 0x00;
+        }
 
         const clusterId = Zdo.ClusterId.NWK_UPDATE_REQUEST;
         const zdoPayload = Zdo.Buffalo.buildRequest(
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -499,8 +499,21 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
     private async changeChannel(oldChannel: number, newChannel: number): Promise<void> {
         logger.warning(`Changing channel from '${oldChannel}' to '${newChannel}'`, NS);
 
+        // According to the Zigbee specification:
+        // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req. 
+        // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
+        let nwkUpdateId: number = 0x00;
+        const isSupportsBackup = await this.adapter.supportsBackup();
+        if (isSupportsBackup) {
+            const backup = await this.adapter.backup(this.getDeviceIeeeAddresses());
+            nwkUpdateId = backup.networkUpdateId ?? 0x00;
+        }
+        if (++nwkUpdateId > 0xFF) {
+            nwkUpdateId = 0x00;
+        }
+
         const clusterId = Zdo.ClusterId.NWK_UPDATE_REQUEST;
-        const zdoPayload = Zdo.Buffalo.buildRequest(this.adapter.hasZdoMessageOverhead, clusterId, [newChannel], 0xfe, undefined, 0, undefined);
+        const zdoPayload = Zdo.Buffalo.buildRequest(this.adapter.hasZdoMessageOverhead, clusterId, [newChannel], 0xfe, undefined, nwkUpdateId, undefined);
 
         await this.adapter.sendZdo(ZSpec.BLANK_EUI64, ZSpec.BroadcastAddress.SLEEPY, clusterId, zdoPayload, true);
         logger.info(`Channel changed to '${newChannel}'`, NS);
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -500,7 +500,7 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
         logger.warning(`Changing channel from '${oldChannel}' to '${newChannel}'`, NS);
 
         // According to the Zigbee specification:
-        // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req. 
+        // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req.
         // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
         let nwkUpdateId: number = 0x00;
         const isSupportsBackup = await this.adapter.supportsBackup();
@@ -508,12 +508,20 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
             const backup = await this.adapter.backup(this.getDeviceIeeeAddresses());
             nwkUpdateId = backup.networkUpdateId ?? 0x00;
         }
-        if (++nwkUpdateId > 0xFF) {
+        if (++nwkUpdateId > 0xff) {
             nwkUpdateId = 0x00;
         }
 
         const clusterId = Zdo.ClusterId.NWK_UPDATE_REQUEST;
-        const zdoPayload = Zdo.Buffalo.buildRequest(this.adapter.hasZdoMessageOverhead, clusterId, [newChannel], 0xfe, undefined, nwkUpdateId, undefined);
+        const zdoPayload = Zdo.Buffalo.buildRequest(
+            this.adapter.hasZdoMessageOverhead,
+            clusterId,
+            [newChannel],
+            0xfe,
+            undefined,
+            nwkUpdateId,
+            undefined,
+        );
 
         await this.adapter.sendZdo(ZSpec.BLANK_EUI64, ZSpec.BroadcastAddress.SLEEPY, clusterId, zdoPayload, true);
         logger.info(`Channel changed to '${newChannel}'`, NS);
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -144,10 +144,17 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
             const netParams = await this.getNetworkParameters();
             const configuredChannel = this.options.network.channelList[0];
             const adapterChannel = netParams.channel;
+            // According to the Zigbee specification:
+            // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req.
+            // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
+            let nwkUpdateID = netParams.nwkUpdateID ?? 0;
+            if (++nwkUpdateID > 0xFF) {
+                nwkUpdateID = 0x00;
+            }
 
             if (configuredChannel != adapterChannel) {
                 logger.info(`Configured channel '${configuredChannel}' does not match adapter channel '${adapterChannel}', changing channel`, NS);
-                await this.changeChannel(adapterChannel, configuredChannel);
+                await this.changeChannel(adapterChannel, configuredChannel, nwkUpdateID);
             }
         }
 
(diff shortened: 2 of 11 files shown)


src/controller/controller.ts
```diff
@@ -499,8 +499,29 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
     private async changeChannel(oldChannel: number, newChannel: number): Promise<void> {
         logger.warning(`Changing channel from '${oldChannel}' to '${newChannel}'`, NS);
 
+        // According to the Zigbee specification:
+        // When broadcasting a Mgmt_NWK_Update_req to notify devices of a new channel, the nwkUpdateId parameter should be incremented in the NIB and included in the Mgmt_NWK_Update_req.
+        // The valid range of nwkUpdateId is 0x00 to 0xFF, and it should wrap back to 0 if necessary.
+        let nwkUpdateId: number = 0x00;
+        const isSupportsBackup = await this.adapter.supportsBackup();
+        if (isSupportsBackup) {
+            const backup = await this.adapter.backup(this.getDeviceIeeeAddresses());
```
Koenkk, line 508: I don't think making a new backup is necessary here? Can't we get the value from the existing backup?

Nerivec:
Was this tested against the described problem?
Some devices are expected to not switch (bad design), some devices require activation to wake them up, then they switch after a few seconds (as described in docs).
I'd expect if this was _required_, all devices would fail to switch, which they do not (I've not found a device, even crappy Tuya, that doesn't switch with current implementation, on my end, several users also tested it and reported no issue).

From what I can tell, that quoted paragraph of the spec is for `Upon receipt of a Mgmt_NWK_Unsolicited_Enhanced_Update_notify message`, which this is not.

In the paragraph about the reception of nwk update req by a device, that value doesn't seem to matter.
```
3) If the ScanDuration parameter is equal to 0xfe, the message is a command to change channels. The device SHALL do the following.
  a) If the nwkNextChannelChange value in the NIB is non-zero, do the following. Compare the channel to change received over the air to the value in the NIB. If the values do not match, do the following:
    i) Follow the Error Response procedure setting the status to NOT_AUTHORIZED.
    ii) The request SHALL be dropped and no more processing SHALL take place.
  b) If there is more than 1 channel indicated in the ScanChannels bitmap (if the message is Mgmt_NWK_Update_req) or in the ScanChannelsListStructure (if the message is Mgmt_NWK_Enhanced_Update_req), then this is an invalid request. Do the following:
    i) Follow the Error Response procedure setting the status to INV_REQUESTTYPE.
    ii) The request SHALL be dropped and no more processing SHALL take place.
  c) The receiving device SHALL determine if the channel is one within the range of all supported channels.
    i) Examine the SupportedChannels element for each entry in the nwkMacInterfaceTable, and determine if there is a match within the received ScanChannels bitmap or ScanChannelsListStructure.
    ii) If no match is found, do the following:
      (1) Follow the Error Response procedure setting the status to INV_REQUESTTYPE.
      (2) The request SHALL be dropped and no more processing SHALL take place.
    iii) If a match is found, perform a channel change.
      (1) Execute a MLME-SET.request for the PIB value phyCurrentPage.
      (2) Execute a MLME-SET.request for the PIB value phyCurrentChannel.
      (3) No further processing SHALL be done.
```

As far as I can tell, the case where the nwkUpdateId would be used would be if a router decides to rejoin (for some reason), and it finds the same network on at least two different nodes, then it should pick the node with the higher nwkUpdateId (if any). But that scenario should not matter for a channel change, since routers receive the broadcast, and switch immediately. Leaving any offline router eventually powered back on, with only the new network available to rejoin.

The spec is a bit lacking on that nwkUpdateId, which I noted in the implementation a while back:
https://github.com/Koenkk/zigbee-herdsman/blob/1c8d8864da7f89c8f18b5290150277ec61f5d873/src/zspec/zdo/buffaloZdo.ts#L1199-L1204

Short version: I don't expect it would solve the problem at hand?
I'm also a little worried about changing this considering the various stacks involved, because it was tested rather extensively during implementation, and after by several users and no particular issue was found that couldn't be attributed to devices being "too cheap". _Hopefully, at worse, it does nothing... but I half expect that most devices just don't use the value at all..._

src/controller/controller.ts
Nerivec: Agreed.
Probably the property should be added to the adapter type NetworkParameters, since it's part of that, and then can be used throughout controller without issue (cached on start).
Remains to see if all adapters provide it though.

test/controller.test.ts
```diff
@@ -56,6 +94,7 @@ const mocksendZclFrameToGroup = vi.fn();
 const mocksendZclFrameToAll = vi.fn();
 const mockAddInstallCode = vi.fn();
 const mocksendZclFrameToEndpoint = vi.fn();
+const mockApaterBackup = vi.fn().mockReturnValue(mockDummyBackup);
```
Nerivec, line 97:
```ts
const mockAdapterBackup = vi.fn(() => Promise.resolve(mockDummyBackup));
```
should provide more accurate typing/testing (and fix typo).
_(A lot of others also need refactoring, but might as well add new ones properly.)_

src/controller/controller.ts
CubeZ2mDeveloper: Okay, I will try to change it to the implementation method you suggested.

CubeZ2mDeveloper:
I added the same device to Home Assistant, then changed the channel. I found that it can be controlled normally in Home Assistant, and I also discovered that Home Assistant will increase the value of nwkUpdateID.

Nerivec:
Was this tested in Z2M against the initial problem?
And, if it does fix it, we need to implement this for all adapters (state management for nwkUpdateId, make sure it's properly reported by adapters, etc.).

CubeZ2mDeveloper:
I have already tested the code on z2m, my device can be controlled normally after the fix.  
Currently, the NwkUpdateId state has been maintained on both the Ember and EZSP adapters, but I am not familiar with the protocol of other adapters, so it has not been implemented yet.

CubeZ2mDeveloper:
After pulling the latest zigbee2mqtt code from the master branch, I started z2m using a Dongle that supports the zstack protocol stack and modified the channel. After changing the channel, the device could be controlled normally on zstack, and the nwkUpdateId incremented by 1. However, when using an Ember Dongle, the device control failed after changing the channel, and the nwkUpdateId remained unchanged, still at 0.

Since our users are currently encountering related issues, this PR might help resolve the problem. Could you let me know about your plans for this PR? For example, are there any further changes needed, or do you have an estimated timeline for merging it? If there’s anything we can do to assist, such as providing devices for testing, please feel free to let us know. Looking forward to your reply, thank you!

src/controller/controller.ts
```diff
@@ -144,10 +144,17 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
             const netParams = await this.getNetworkParameters();
             const configuredChannel = this.options.network.channelList[0];
             const adapterChannel = netParams.channel;
+            // According to the Zigbee specification:
```
Nerivec, line 147: The increment should be done in `changeChannel` so it only happens when the channel is actually changed (re-get `getNetworkParameters` in the function instead of passing a param, it's cached anyway).
Also might need to deal with `sendZdo` potentially failing (I think that will stop Z2M startup, so should be fine, but needs to be verified).

src/adapter/tstype.ts
```diff
@@ -74,4 +74,5 @@ export interface NetworkParameters {
     panID: number;
     extendedPanID: string; // `0x${string}` same as IEEE address
     channel: number;
+    nwkUpdateID?: number;
```
Nerivec, line 77: This should not be optional, and support should added be in all drivers.

---

The change should also be tested at least in zstack and ember (the two _officially_ supporting change channel) with at least a few devices that worked before, to make sure they still work after this change. Since it doesn't appear to be required for a change channel (based on spec & previous testing with many devices), it might have negative impact on other aspects (like some devices not expecting a changed nwkUpdateID when receiving a change channel request, and ignoring the request entirely).
In that regard, can you confirm which devices (models) are having the issue in the first place?

src/controller/controller.ts
CubeZ2mDeveloper: I’ve updated the code according to your suggestions, moving the increment operation to changeChannel.

src/adapter/tstype.ts
CubeZ2mDeveloper: Currently, the main devices we have seen issues with are S31ZB, S40ZBTPB and S26R2ZB.
 
Regarding the necessity of changing channels, I’d like to share some real-world user experiences: in some households with complex network conditions (e.g., high Wi-Fi interference), changing channels has significantly improved the operational experience of devices. That said, I fully acknowledge that this might introduce some negative impacts, such as potential incompatibility with certain devices.

CubeZ2mDeveloper: Regarding the change of nwkUpdateID during channel changes, based on my understanding of the Zigbee specification (which I’ve uploaded as an image), incrementing the nwkUpdateID value is necessary. Additionally, when I performed channel changes using the same dongle on ZHA, devices that failed to work properly after a channel change in Z2M were able to function normally after a channel change in ZHA.

I fully understand your concerns about the potential impact of this PR on devices that are already working correctly. The brands you mentioned (Tuya, Hue, Inovelli, Ikea, Ledvance) indeed require comprehensive testing. However, due to limited device availability, I can only test with the devices I currently have. I will provide you with a detailed list of tested devices shortly, so you can see the coverage of my testing.

image
image
image

Nerivec: _I updated images to latest rev._
I find that part of the spec is not very clear:
- could be specific: `Upon receipt of a Mgmt_NWK_Unsolicited_Enhanced_Update_notify message`
- logic is optional/not strictly defined: `MAY do the following`

Which, at best explains why it's not needed in most cases, most stacks likely just ignore it, because it's easier and not _required_.
There are two scenario that concern me:
- a device stack could ignore the update request if the nwkUpdateID is different from current due to some weirdness in implementation (or some leftovers from older spec revisions)
- a coordinator stack could be overriding the nwkUpdateID logic internally, resulting in mismatch

Also, it would seem the devices you mentioned have a firmware quirk no matter what, because it shouldn't matter for routers that are online at the time of broadcast (they should just switch channel).

I don't have enough devices either for something like this. I had a few users test it out when it was first implemented to confirm with a bigger sample.
Koenkk will confirm zstack impl and run some tests too.
Once we confirm with a few (non-affected) brands, it should be fine to merge.
Need to fix the zeroes for other stacks though.

deconz: page 11 seems to indicate where the parameter is: [deCONZ-Serial-Protocol_en_1.22.pdf](https://github.com/user-attachments/files/18709156/deCONZ-Serial-Protocol_en_1.22.pdf)
zigate: no clue, doesn't seem mentioned in https://zigate.fr/documentation/commandes-zigate/ but change channel is not confirmed working either, so, zero should be fine for now...
zboss: @kirovilya can you check how to retrieve the current nwkUpdateID?

And also, this is likely to require some tests to be updated in Z2M repo since it changes the return type of https://github.com/Koenkk/zigbee2mqtt/blob/dev/lib/zigbee.ts#L216

Koenkk: > I'd like to confirm that other brands/models still work with this PR, at least Tuya, Hue, Inovelli, Ikea, Ledvance for a good sample, and including routers & end devices.

I tried changing channel with this change and the following devices 3 times and it worked every time, so looks OK to me (Elivco = Tuya)

Screenshot 2025-02-07 at 21 35 03

Koenkk: I just found out that zstack ignores the nwkUpdateId send from zh as it is not part of the API:

Screenshot 2025-02-08 at 14 26 53

Instead zstack automatically increments the nwkUpdateId (based on the one in the NIB, I check their source code). This explains why zstack worked even before this change:

> After changing the channel, the device could be controlled normally on zstack, and the nwkUpdateId incremented by 1. However, when using an Ember Dongle, the device control failed after changing the channel, and the nwkUpdateId remained unchanged, still at 0.

For adapters that do this, I would propose the return a dummy nwkUpdateID of 0, no need to do unnecessary requests to the adapter.

Nerivec: Wouldn't returning zero be a problem though? If you ever change the channel twice (or more) without resetting the network in-between but restarting Z2M, that would no longer be properly incremented (0, 1, 0, 1...).

Koenkk: > If you ever change the channel twice (or more) without resetting the network in-between but restarting Z2M, that would no longer be properly incremented (0, 1, 0, 1...).

What I tried is changing the channel multiple times, it was always properly incremented (1,2,3).

> ZDO request being broadcasted? Is the request being altered with the new NIB value too?

Yes, note that zstack doesn't even ignore the value, it isn't even passed to it since `nwkUpdateId` is not in https://github.com/Koenkk/zigbee-herdsman/blob/877e905e251d54aabd313901fc97db5e3f09a4fd/src/adapter/z-stack/znp/definition.ts#L1364

Nerivec: As expected, this param is a mess...
Have you tried sniffing the ZDO broadcast that the coordinator then sends over the network to see what value the nwkUpdateId has?

Koenkk: > Have you tried sniffing the ZDO broadcast that the coordinator then sends over the network to see what value the nwkUpdateId has?

Yes, it is incremented by 1 after every channel change.

Nerivec: For deconz, I think this should do, can you confirm with your adapter?

```ts
NWK_UPDATE_ID: 0x24,
```
```ts
const nwkUpdateId = await this.driver.readParameterRequest(PARAM.PARAM.Network.NWK_UPDATE_ID);
```

Although the network params changing is all screwed up in deconz, so, won't do much good for changing channel.
https://github.com/Koenkk/zigbee-herdsman/pull/1257 should solve that, but requires extensive testing...

Remains zboss.

Koenkk: It doesn't work (always returns `null`), also the [existing logic](https://github.com/Koenkk/zigbee-herdsman/blob/481a773e8252812c4f0f855f6d90259b2478fd10/src/adapter/deconz/adapter/deconzAdapter.ts#L93) doesn't work. Since we don't support changing channel for Deconz anyway (https://www.zigbee2mqtt.io/guide/adapters/deconz.html) I propose it to leave it `0`.

Koenkk: @kirovilya could you check the zboss part? Than we can merge this PR

Nerivec: About the `NWK_UPDATE_ID`, I guess it could be about firmware version? Lots of features appear to be missing in some deconz firmware versions.

About the existing logic, I'm not surprised since the read is for `CHANNEL` but the write is for `CHANNEL_MASK`
https://github.com/Koenkk/zigbee-herdsman/blob/481a773e8252812c4f0f855f6d90259b2478fd10/src/adapter/deconz/adapter/deconzAdapter.ts#L89
https://github.com/Koenkk/zigbee-herdsman/blob/481a773e8252812c4f0f855f6d90259b2478fd10/src/adapter/deconz/adapter/deconzAdapter.ts#L158
I'm thinking that `CHANNEL_MASK` is the mask used when scanning for available networks, not the channel the coordinator should use.

Koenkk: This was with the latest Conbee II firmware (`0x26780700`)

kirovilya: Hi. For **zboss** I did not find a direct request to read the current nwk update id . There is only scanning (`NWK_DISCOVERY`) of channels, but it gets the values ​​of networks that are on the specified channel.

kirovilya: Some different request format for zboss:
image

Rewrite zboss adapter code for NWK_UPDATE:
image
update id = 0.
the connected socket changed channel to 13 (saw it with sniffer on new channel).
but! at what point should the coordinator change the channel to 13?

And get timeout for update response:
image

Nerivec: Looks like same behavior as zstack, it overrides the update ID internally, or at least it should since missing from request params, but apparently isn't since effective request has 0.

Also looks like it's not answering the ZDO properly (or the ZH logic to validate the waiter is missing something?). That should be near immediate. The channel change is truly effective after a while, usually around 9sec (time for broadcast to propagate throughout the network), not sure if zboss emits a stack-specific event or not when that happens (ember does, zstack doesn't). ZH has a 12sec delay after changing to be sure.

Koenkk: Since zboss is experimental and also channel change is not listed as [supported](https://www.zigbee2mqtt.io/guide/adapters/zboss.html) I propose to ignore zboss for now, that means this PR should be ready for merge now.

Nerivec: I think one issue remains. Now that `networkUpdateId` in Backup is no longer always zero, we have to have support for it in restore codepaths, otherwise it will restore mismatching IDs between coordinator and devices.
Like here for ember:
https://github.com/Koenkk/zigbee-herdsman/blob/946f073eb01ad776c1d7a37179ca3c3575d5d4ee/src/adapter/ember/adapter/emberAdapter.ts#L1103

CubeZ2mDeveloper: Agree. However, I tried to start z2m by restoring from a backup and setting the nwkUpdateID to 1, but after forming the network, it seems that the nwkUpdateID was not set to 1.
The version of z2m is 2.1.1.
d70f84ff098e7cf51805dca53c3136b1

Nerivec: Did you make the changes so the value from the backup would be used? Have to pass one more param to the function I linked above, then replace that line with `nwkUpdateId,`
When called from restored path (`FORM_BACKUP`), you set it to whatever is in backup, when called from reset path (`FORM_CONFIG`), you set it to zero.
Have to do roughly the same for `ezsp`.
I'm not sure how that works out for `zstack`. @Koenkk can you check?
Others don't support backup anyway...
_Several tests will need fixing after this is properly added._

Koenkk: @CubeZ2mDeveloper for zstack you have to write it in the NV item which was previously read.

CubeZ2mDeveloper: During the restore process of ezsp and zstack, the nwkUpdateID has been set to the nwkUpdateId from the backup.

ezsp
https://github.com/Koenkk/zigbee-herdsman/blob/76d3de1e45e4bb61de3652ac016d5e7e75ab172e/src/adapter/ezsp/driver/driver.ts#L357

zstack
https://github.com/Koenkk/zigbee-herdsman/blob/76d3de1e45e4bb61de3652ac016d5e7e75ab172e/src/adapter/z-stack/adapter/adapter-backup.ts#L275

CubeZ2mDeveloper: @Nerivec Yes, I have made the changes as you described.

Koenkk:
Thanks @CubeZ2mDeveloper !

PR #1282, opened 2024-12-30
Extend genBasic support

kennylevinsen:
Currently fails on coverage as we do not have any tests with frames containing `productCode`, `productLabel` or `productUrl` so the setters do not get exercised. The latter is quite common though.


Nerivec:
We should limit this to update cluster.ts only (the missing properties) for reference.
There is no use case for any of these properties, and most manufacturers don't set them at all (or worse, set to long null strings). They would bloat database/payloads for no reason.

kennylevinsen:
Why would unused properties bloat the database? It would be undefined and therefore not written, so it doesn't seem like something that we need to worry about.

Generic device type and product URL are generally set, but indeed I have yet to see e.g. productCode.

Nerivec:
> It would be undefined

In cases where the fields are filled with dummy/null characters that would not be the case, and could represent long useless strings of data. In cases where the fields are left to the firmware builder's defaults by the manufacturer, some values could be all wrong on top of it. Additionally, URLs could become (or already be) defunct, not to mention could since have been hijacked, etc., so, providing them to the end-user is a non-starter (I'm not even sure why the ZigBee spec bothered with these... 😕).

More importantly, they just have no use (they bring no actual value and are rarely even set -properly-), but they add code to maintain/cover with tests/be executed.

kennylevinsen:
I don't agree that the fields don't have value, but I do agree that when the fields are rarely if ever set that the gain is minimal and as such not something I plan on spending a whole lot of energy defending nor something I feel like crafting packets for. :)

I'll leave it at the cluster changes so that users at least get access to the attributes in the dev console.

Koenkk:
Thanks!

andrei-lazarov:
**Hi! I found `productCode` is filled correctly for all my IKEA devices.**
_SONOFF and TUYA give `unsupported_attribute`._

**I want it in `database.db` because it would allow setting the correct picture for IKEA plugs.**
Currently they report `INSPELNING Smart plug` and `TRETAKT Smart plug`.
We ended up with this mess: [E22xx.png](https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/E22xx.png) 😄 

Here is TRETAKT **E2204**, currently recognized as **E22x4**:
> Read result of 'genBasic': {"productCode":{"type":"Buffer","data":[69,50,50,48,52]}}

Here is STYRBAR **E2313**, currently recognized as **E2001/E2002**:
> Read result of 'genBasic': {"productCode":{"type":"Buffer","data":[69,50,51,49,51]}}

What do you think?

chris-1243:
This would be a way to differentiate some IKEA devices for sure.

I just had a look at the **E2214** and it shows:

> Read result of 'genBasic': {"productCode":{"type":"Buffer","data":[69,50,50,49,52]}}

@andrei-lazarov 

> We ended up with this mess: [E22xx.png](https://github.com/Koenkk/zigbee2mqtt.io/blob/master/public/images/devices/E22xx.png) 😄

There were no way to differentiate those devices so a generic name had to be defined. Not the best, I do agree

PR #1283, opened 2025-01-02
add manuSpecificSalus cluster

mkrle:
Feedback welcome!

While trying to figure out how to control presets on Salus FC 600 (https://github.com/Koenkk/zigbee-herdsman-converters/pull/8528) I was able to map out parts of a manufacturer specific cluster it uses.

I was able to probe attributes 0x004-0x007 and 0x000F (and their data types) but I could not figure out their purpose. I have labeled them `attr4`, `attr5` and so on using their decimal IDs, perhaps someone will figure out the purpose later on. Please let me know if this is not a good approach.

The command performs a soft reset (reboot) of the thermostat, and the thermostat never sends back the response. I don't know if and how it's possible to flag this, so that the command doesn't time out with an exception when in fact it was successful. 

If and when this is accepted, I'll look into exposing the presets for FC600.

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -5150,12 +5150,12 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
             attr7: {ID: 0x0007, type: DataType.UINT8},
             autoCoolingSetpoint: {ID: 0x0008, type: DataType.INT16},
             autoHeatingSetpoint: {ID: 0x0009, type: DataType.INT16},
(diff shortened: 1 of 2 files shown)


Nerivec:
New custom clusters should be added via zigbee-herdsman-converters
Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/20c0958f6e99343040b334898de494fb97f18aa8/src/devices/sonoff.ts#L1105

Koenkk:
Please make a pr for zigbee-herdsman-converters, will close this one.

mkrle:
Thanks, opened https://github.com/Koenkk/zigbee-herdsman-converters/pull/8551

PR #1284, opened 2025-01-02
Update Aqara specific attributes to support PowerOnBehaviour for Aqara T1M lamp (CL-L02D/lumi.light.acn032)

brainstormi:
```
Cluster: 0xFCC0 # Manufacturer Specific
Endpoint: 0x01
Attribute: 0x0517
Type: 0x20 # 8-bit int
Manufacturer Code: 4447
Value: 
    0x00 - Always On (default) - Will be turned on after power outage.
    0x01 - Preserve State - Will keep the state it was before power outage.
    0x02 - Always Off - Will be turned off after power outage.
```


Koenkk:
The intention is not to update the custom clusters here anymore but since you added it by [id here](https://github.com/Koenkk/zigbee-herdsman-converters/pull/8550/files#diff-11e8450decb6d1cc6caa9e7b7c76b5d53c34b34d5e6e30f1b6b55122a20c5e54R4467) this shouldn't be needed.

PR #1293, opened 2025-01-13
ZBOSS: bug in zboss - reverse extendedpanid

kirovilya:


src/adapter/zboss/driver.ts
```diff
@@ -171,7 +171,8 @@ export class ZBOSSDriver extends EventEmitter {
         const ieeeAddr = result.payload.ieee;
 
         result = await this.execCommand(CommandId.GET_EXTENDED_PAN_ID, {});
-        const extendedPanID = result.payload.extendedPanID;
+        // TODO: bug in extendedPanID - got reversed value
+        const extendedPanID = result.payload.extendedPanID.reverse();
```
Nerivec, line 175: No chance that payload is used again (after this call), right? Because reverse is "in-place".

kirovilya: I don't see where it can be used yet. When I checked earlier, the extpanid was made of identical elements, so I didn't notice the problem. And [here](https://github.com/Koenkk/zigbee2mqtt/issues/24705#issuecomment-2586719350) I suddenly discovered that a reset occurs at startup - the set and read values do not match.

Koenkk:
Thanks!

PR #1298, opened 2025-01-15
Update clusters for SprutDevice

BurchuMikhail:
- add the ability to change baud rate
- add a flag for the presence of a modbus connection
- add debug mode enable command


Koenkk:
Please use the custom clusters API instead (example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/ef7ac30befdfa980f139249ea62c8db7a985d9a5/src/devices/sonoff.ts#L1115)

PR #1299, opened 2025-01-16
Update cluster.ts

BurchuMikhail:
delete a nonexistent attribute


src/zspec/zcl/definition/cluster.ts
```diff
@@ -4663,9 +4663,7 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
     sprutDevice: {
```
Koenkk, line 4663: Can the complete cluster be removed?

PR #1300, opened 2025-01-16
fix: Fill out missing Touchlink cluster info

Nerivec:
I was investigating a Touchlink-related matter and noticed the spec was behind, so, added (most of) the missing stuff.
Figure this can start a bettering of the Touchlink controller in general (network commissioning & co), if someone is motivated 😉
_(I don't have any Touchlink, so I won't be much help there)_

It adds a new `ParameterCondition.FIELD_EQUAL` to match the Touchlink requirements for payload decoding.

@Koenkk I included a cleanup of some related tests, I don't believe copying the raw clusters in the tests brings any real value, so might as well reduce the amount of data (by a lot!)... let me know what you think.


Koenkk:
> @Koenkk I included a cleanup of some related tests, I don't believe copying the raw clusters in the tests brings any real value, so might as well reduce the amount of data (by a lot!)... let me know what you think.

Perfect ❤️

PR #1307, opened 2025-01-20
fix(ignore): cleanup logger, add waitress clear util

Nerivec:
- Remove lambda for `error` since it's always called anyway
- Add ISO date to default logger (_when `setLogger` not called, Z2M not concerned, but useful at least in testing ZH alone_)
- Add `clear` util fn to waitress (should avoid cascading errors in scenario where the waitress should no longer be expected to resolve, like a NCP crash)
  - TODO: need to figure out if should also reject every entry when called?
  - TODO: add calls wherever appropriate (separate PRs I think)


Koenkk:
Thanks!

PR #1301, opened 2025-01-18
Move Niko custom clusters to zigbee-herdsman-converters

svenjochems:
The niko specific clusters are moved to zigbee-herdsman-converters in https://github.com/Koenkk/zigbee-herdsman-converters/pull/8635

PR #1310, opened 2025-01-25
Update cluster.ts

fairecasoimeme:
Add new attribute to LixeeCluster

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -4529,6 +4529,7 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
             warnDIR2: {ID: 0x0007, type: DataType.UINT16},
             warnDIR3: {ID: 0x0008, type: DataType.UINT16},
             motDEtat: {ID: 0x0009, type: DataType.CHAR_STR},
+            tariffPeriod : {ID: 0x0010, type: DataType.CHAR_STR},
             currentPrice: {ID: 0x0200, type: DataType.CHAR_STR},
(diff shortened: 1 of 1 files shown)


Nerivec:
This cluster should be moved to custom clusters directly in
https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/lixee.ts

Koenkk:
Please use the custom cluster API instead as @Nerivec mentioned, see https://github.com/Koenkk/zigbee-herdsman-converters/blob/284e8557ac7bde72429daee2f168f89e0050d4cd/src/devices/sonoff.ts#L41 for an example. You can update this PR to remove the cluster entirely.

KipK:
@Koenkk , can this be merged until a new external converter using modernExtend customCluster is made ? 
It will take some time, seems to be a massive rewrite of lixee, and there's nobody confident enough with zhc codebase yet to do it ( and lack of dev documentation ) 

The other [PR ](https://github.com/Koenkk/zigbee-herdsman-converters/commit/284e8557ac7bde72429daee2f168f89e0050d4cd) on zhc related to this one has been merged and released in 2.1.0, but is not totally functional because this one is missing. 

Thanks

Nerivec:
The issue is only to move the custom LiXee cluster to ZHC. There are plenty of examples in recent PRs. Since the goal is to remove custom clusters from the ZigBee spec, they are all being moved whenever the occasion presents.
Here's an example:
https://github.com/Koenkk/zigbee-herdsman/pull/1301
https://github.com/Koenkk/zigbee-herdsman-converters/pull/8635 (ignoring new features also implemented by that PR)

KipK:
Ok I think I got it. Have to test first.

PR #1318, opened 2025-02-12
feat: Allow using custom clusters with groups

ccutrer:
But only if ALL devices in the group support the cluster. It's highly dependent on individual devices if they will support read/write/command for custom clusters. In my testing, my Inovelli switches are responding to writing attributes, but not commands (such as LED commands).

### DIFF
--- a/src/controller/model/group.ts
+++ b/src/controller/model/group.ts
@@ -2,6 +2,7 @@ import assert from 'node:assert';
 
 import {logger} from '../../utils/logger';
 import * as Zcl from '../../zspec/zcl';
+import {CustomClusters} from '../../zspec/zcl/definition/tstype';
 import ZclTransactionSequenceNumber from '../helpers/zclTransactionSequenceNumber';
 import {DatabaseEntry, KeyValue} from '../tstype';
 import Device from './device';
@@ -27,6 +28,7 @@ export class Group extends Entity {
     private databaseID: number;
     public readonly groupID: number;
     private readonly _members: Set<Endpoint>;
+    private _customClusters: CustomClusters | null = null;
     get members(): Endpoint[] {
         return Array.from(this._members).filter((e) => e.getDevice());
     }
@@ -164,11 +166,13 @@ export class Group extends Entity {
 
     public addMember(endpoint: Endpoint): void {
         this._members.add(endpoint);
+        this._customClusters = null;
         this.save();
     }
 
     public removeMember(endpoint: Endpoint): void {
         this._members.delete(endpoint);
+        this._customClusters = null;
         this.save();
     }
 
@@ -182,7 +186,7 @@ export class Group extends Entity {
 
     public async write(clusterKey: number | string, attributes: KeyValue, options?: Options): Promise<void> {
         const optionsWithDefaults = this.getOptionsWithDefaults(options, Zcl.Direction.CLIENT_TO_SERVER);
-        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, {});
+        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, this.customClusters);
         const payload: {attrId: number; dataType: number; attrData: number | string | boolean}[] = [];
 
         for (const [nameOrID, value] of Object.entries(attributes)) {
@@ -210,7 +214,7 @@ export class Group extends Entity {
                 'write',
                 cluster.ID,
                 payload,
-                {},
+                this.customClusters,
                 optionsWithDefaults.reservedBits,
             );
 
@@ -226,7 +230,7 @@ export class Group extends Entity {
 
     public async read(clusterKey: number | string, attributes: (string | number)[], options?: Options): Promise<void> {
         const optionsWithDefaults = this.getOptionsWithDefaults(options, Zcl.Direction.CLIENT_TO_SERVER);
-        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, {});
+        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, this.customClusters);
         const payload: {attrId: number}[] = [];
 
         for (const attribute of attributes) {
@@ -242,7 +246,7 @@ export class Group extends Entity {
             'read',
             cluster.ID,
             payload,
-            {},
+            this.customClusters,
             optionsWithDefaults.reservedBits,
         );
 
@@ -263,7 +267,7 @@ export class Group extends Entity {
 
     public async command(clusterKey: number | string, commandKey: number | string, payload: KeyValue, options?: Options): Promise<void> {
         const optionsWithDefaults = this.getOptionsWithDefaults(options, Zcl.Direction.CLIENT_TO_SERVER);
-        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, {});
+        const cluster = Zcl.Utils.getCluster(clusterKey, undefined, this.customClusters);
         const command = cluster.getCommand(commandKey);
 
         const createLogMessage = (): string => `Command ${this.groupID} ${cluster.name}.${command.name}(${JSON.stringify(payload)})`;
@@ -279,7 +283,7 @@ export class Group extends Entity {
                 command.ID,
                 cluster.ID,
                 payload,
-                {},
+                this.customClusters,
                 optionsWithDefaults.reservedBits,
             );
 
@@ -302,6 +306,32 @@ export class Group extends Entity {
             transactionSequenceNumber: undefined,
             ...(options || {}),
         };
+    }
+
+    /**
+     * Get custom clusters that all members share.
+     */
+    get customClusters(): CustomClusters {
+        if (this._customClusters) {
+            return this._customClusters;
+        }
+
+        if (this._members.size === 0) {
+            return {};
+        }
+
+        const firstMember = Array.from(this._members)[0];
+        const customClusters = firstMember.getDevice().customClusters;
+
+        const commonClusters: CustomClusters = {};
+        for (const clusterName in customClusters) {
+            if (Array.from(this._members).every((member) => clusterName in member.getDevice().customClusters)) {
+                commonClusters[clusterName] = customClusters[clusterName];
+            }
+        }
+
+        this._customClusters = commonClusters;
+        return this._customClusters;
     }
 }
 
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -5665,6 +5665,39 @@ describe('Controller', () => {
         expect(mocksendZclFrameToGroup.mock.calls[0][2]).toBeUndefined();
     });
 
+    it('Write to group with custom cluster', async () => {
+        await controller.start();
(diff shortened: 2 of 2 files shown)


src/controller/model/group.ts
```diff
@@ -27,6 +28,7 @@ export class Group extends Entity {
     private databaseID: number;
     public readonly groupID: number;
     private readonly _members: Set<Endpoint>;
+    private _customClusters: CustomClusters | null = null;
```
Nerivec, line 31: should use `undefined`, not `null`

src/controller/model/group.ts
```diff
@@ -303,6 +307,32 @@ export class Group extends Entity {
             ...(options || {}),
         };
     }
+
+    /**
+     * Get custom clusters that all members share.
+     */
+    get customClusters(): CustomClusters {
```
Nerivec, line 314: I can foresee sync issues with this logic.

src/controller/model/group.ts
ccutrer: Will do.

src/controller/model/group.ts
ccutrer: Just with members of the group being out of date, or something else? At first I didn't cache this at all, but I didn't want to run this calculation every time a read/write/command is sent. I could go back to that.

src/controller/model/group.ts
```diff
@@ -303,6 +307,32 @@ export class Group extends Entity {
...
+        if (this._members.size === 0) {
+            return {};
+        }
+
+        const firstMember = Array.from(this._members)[0];
+        const customClusters = firstMember.getDevice().customClusters;
+
+        const commonClusters: CustomClusters = {};
+        for (const clusterName in customClusters) {
+            if (Array.from(this._members).every((member) => clusterName in member.getDevice().customClusters)) {
```
Nerivec, line 328: Should optimize this for performance. And needs testing with large-ish groups, to see the impact on first exec (could "freeze" for extended periods, worse than it already does in some cases).

ccutrer: What would you recommend? I'm not a JS/TS guy. But I can definitely do performance testing here. I have a group with 85 members in my network that I can send a write to. What do you want me to measure? But not for a custom cluster, though it shouldn't matter, as currently structured custom clusters for the group are calculated regardless of the first write to the group. BUT... I could change that to _only_ calculate the custom clusters for the group if we couldn't find the cluster at first in a read/write/command. That would make this change essentially a no-op for 99.9% of users.

src/controller/model/group.ts
Nerivec: Definitely needs some kind of caching due to the amount of calls that could be generated, but it needs to be invalidated by some events. I can think of at least OTA updates (otherwise it breaks without a restart), and re-interview (and re-configure?), basically anything that can change the clusters of a device.

src/controller/model/group.ts
Nerivec: I'm not overly familiar with Z2M group logic, Koenkk should be able to answer that better than me.

I'd first look into optimizing the number of nested loops as much as possible (avoiding the array conversion would be a good start).
No-oping for majority of users is also always preferred when possible.

As for testing, we need to compare what the first execution adds (i.e. when it's first cached) to current time for the same function call, make sure it's not going to create a bottleneck (mostly for low resource scenario).

src/controller/model/group.ts
```diff
@@ -303,6 +307,32 @@ export class Group extends Entity {
             ...(options || {}),
         };
     }
+
+    /**
+     * Get custom clusters that all members share.
+     */
+    get customClusters(): CustomClusters {
+        if (this._customClusters) {
+            return this._customClusters;
```
Koenkk, line 316: This value should be re-calculated once the group members change. Maybe make a `private computeCustomClusters` which is called when first called and when group members change?

Nerivec: This goes back to what I mentioned here https://github.com/Koenkk/zigbee-herdsman/pull/1318#discussion_r1953282248

ccutrer: yes, that's exactly what it's already doing, but @Nerivec brought up a good point that the clusters available on device/endpoint could change without group membership changing, so I need to handle several of those cases as well

src/controller/model/group.ts
Nerivec: Another thought. We should also be looking out for potential race conditions since this is triggered from various places.

ccutrer: Is there any documentation on the threading model so I can get more familiar with it? If not, that's fine, I'm happy to go code diving and figure it out.

src/controller/model/group.ts
```diff
@@ -27,6 +28,7 @@ export class Group extends Entity {
     private databaseID: number;
     public readonly groupID: number;
     private readonly _members: Set<Endpoint>;
+    private _customClusters: CustomClusters | undefined;
     get members(): Endpoint[] {
         return Array.from(this._members).filter((e) => e.getDevice());
```
Nerivec, line 33: @Koenkk remember the reasoning behind this logic? (not related to the PR, but odd)

src/controller/model/group.ts
```diff
@@ -27,6 +28,7 @@ export class Group extends Entity {
     private databaseID: number;
     public readonly groupID: number;
     private readonly _members: Set<Endpoint>;
+    private _customClusters: CustomClusters | undefined;
```
Nerivec, line 31: I think we should name this `_commonCustomClusters` to avoid confusion (could be all clusters of all devices with current name).

src/controller/model/group.ts
```diff
@@ -303,6 +307,45 @@ export class Group extends Entity {
             ...(options || {}),
         };
     }
+
+    /**
+     * Calculate, store, and return custom clusters that all members share.
+     */
+    private calculateCustomClusters(): CustomClusters {
```
Nerivec, line 314: `calculate` doesn't seem to quite fit. Maybe `identifyCommonCustomClusters`?

src/zspec/zcl/utils.ts
Nerivec, line 1: This shouldn't need to be changed.

_Also, considering the changes, and the fact tests are still passing, looks like there are gaps in coverage that are not being properly detected (hence, could be misbehaving from previous logic and we wouldn't know)._

src/controller/model/group.ts
```diff
@@ -27,6 +28,7 @@ export class Group extends Entity {
     private databaseID: number;
     public readonly groupID: number;
     private readonly _members: Set<Endpoint>;
```
Nerivec, line 30: @Koenkk I think between the logic in `members` and now the new logic from this PR, it would be more efficient to just use an array here, and have a check in `add` instead of using a `Set` that ends up often converted to an array... or we get rid of Array-specific functions, use `for (const member of this._members)` and keep the `Set`.

ccutrer:
I'm sorry I haven't had time to get back to this. Other projects have taken priority. I still plan on getting back to it, but don't have a specific estimate as to when. Feel free to finish it yourself if you feel like it warrants it. Otherwise... I'll get to it sometime!

Nerivec:
Will finish https://github.com/Koenkk/zigbee-herdsman/pull/1330 first then we can cleanup this one.

slyglif:
Adding a comment to hopefully make this not stale.  I'm really hoping this functionality is included soon so I can keep a bunch of settings on dozens of Inovelli switches in sync.

ccutrer:
@Nerivec: were you planning on getting back to finishing this sometime soon?

Nerivec:
Made a couple attempts a while back, but nothing satisfactory. The logic is too convoluted, and there is a fatal flaw in the detection that requires some refactoring first to support this (a group would be considered as having a common custom cluster, even if the group doesn't have the specific device _endpoint_ that actually supports this custom cluster, which would make requests partially fail).

I also don't have any use case, so, I can check the logic, but not test any of this.

ccutrer:
I'm not sure I fully understand what the next steps are. The logic and fatal flaw... in this PR as-is... is too convoluted? Or you just don't think this feature should be implemented at all?

You're still planning on getting to this, but will need help testing? Or because you can't test, you no longer plan on working on this.

Basically... are you expecting me to pick it back up, and if so, can you give any pointers on what I should address first? Or is it just a "no, this isn't going to happen"?

Nerivec:
The convoluted part was about the logic to keep the common clusters in sync. _There should be a way to make this simpler._

The flaw part was about the fact we need custom clusters to be endpoint-aware before this can be implemented ("what custom clusters are supported on this endpoint on this device" so the group limits the checking to the endpoint actually in the group). Two PRs are needed, in ZH and in ZHC, so devices can declare which endpoints are concerned when custom clusters are added.

I don't have enough time to get into big PRs at the moment, but later in the year, sure.

Probably should first discuss the best approach for the endpoint-aware part. @Koenkk we only flew over the subject... not sure if you've had any ideas on this.

Koenkk:
> what custom clusters are supported on this endpoint on this device" so the group limits the checking to the endpoint actually in the group

I would expect that if a device has 2 endpoints where both endpoints support a manuspecific cluster with id 99, that both are the same the same cluster definition (so cluster definitions are always device wide, not endpoint specific). Or am I missing something?

Nerivec:
That's not the problem here.

A device could have 2, 3, 4... endpoints, and say only 1 supports the custom cluster.
If the device is included in the group, it would always assume that custom cluster is common, no matter what endpoint is actually in the group, and if that endpoint indeed supports that cluster or not.
That would result in partial failure, since the device will obviously not properly respond (I'm not even sure what would happen in this scenario, probably some stack-specific outcome...).
- If you send a group command to a non-existent custom cluster on a device's endpoint because the group assumed it was supported, you might bring troubles.
- Same goes if for example, that device's endpoint actually implements the same cluster ID but with a different structure/behavior (mismatching types, out of bounds, etc. that kind of parsing failure on the device is likely to crash lesser-quality ones...). _Actually this one might be a problem even without multi-endpoint in the mix._

At best it's a failed request and is silently dropped by the device, but it's something Z2M should not allow for safety.

ccutrer:
Indeed. In this case, the Inovelli switch has manuSpecificInovelli cluster as an _input_ on endpoint 1, and an _output_ on endpoints 2 and 3 (and not present on endpoint 242).

Koenkk:
I think a group should only support a manuspecific cluster if all endpoints in the group support it.

Nerivec:
That would be bring the number down to pretty much zero, since a lot of devices have GP 242 cluster that won't support custom. Also several have either "mistake" secondary clusters (bad/leftover firmware config), or a secondary that's just for a specific use (identify, etc.), that would have the same impact.

Koenkk:
What I mean is e.g.:

- Group 1: Inovelli device 1 ep 1 which supports cluster X, Inovelli device 2 ep 1 which supports cluster X -> group supports cluster X
- Group 2: Inovelli device 1 ep 1 which supports cluster X, Inovelli device 2 ep 1 which supports cluster X, some Ikea device ep 1 -> group **does not** supports cluster X

PR #1322, opened 2025-02-22
feat: Add manuSpecificPhilips3 cluster definition.

chrivers:
The Philips Hue ecosystem uses cluster 0xFC01 for "entertainment mode" packets. This is a manufacturer-specific extension, that allows high-frequency updates (tested up to 90 *per second*, but typically 10-50) for up to 10 lights.

Unfortunately, `manuSpecificLegrandDevices` already "owns" 0xFC01, making it impossible to send the required commands to make this work, even with the necessary patch to enable full control in `zcl_command`:

  https://github.com/Koenkk/zigbee-herdsman-converters/pull/8867

With this change, the `.cluster` field in `zcl_command` can be specified as `manuSpecificPhilips3`, and will be correctly identified by zigbee-herdsman.

These changes (this PR and the one mentioned above) is sufficient to allow Entertainment Mode control over Zigbee2MQTT!

The upcoming beta version of [Bifrost](https://github.com/chrivers/bifrost) will have rudimentary support for this.

A note on the command layout: Obviously, this is not the final, polished form of this, but it's a flexible way to allow Bifrost to construct the somewhat intricate commands, and allow more testing.

In the meantime, @danieledwardgeorgehitchcock is leading the effort to provide a "good citizen" version of this support, which integrates properly with Zigbee2MQTT. However, this work is a bit further from completion, and we would *really* appreciate being able to offer this much-sought-after feature as soon as possible.

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -4161,6 +4161,34 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
         },
         commandsResponse: {},
     },
+    manuSpecificPhilips3: {
+        ID: 0xfc01,
+        manufacturerCode: ManufacturerCode.SIGNIFY_NETHERLANDS_B_V,
+        attributes: {},
+        commands: {
+            command1: {
+                ID: 1,
+                parameters: [{name: 'data', type: BuffaloZclDataType.BUFFER}],
+            },
+            command2: {
+                ID: 2,
+                parameters: [{name: 'data', type: BuffaloZclDataType.BUFFER}],
+            },
+            command3: {
+                ID: 3,
+                parameters: [{name: 'data', type: BuffaloZclDataType.BUFFER}],
+            },
+            command4: {
+                ID: 4,
+                parameters: [{name: 'data', type: BuffaloZclDataType.BUFFER}],
+            },
+            command7: {
+                ID: 7,
+                parameters: [{name: 'data', type: BuffaloZclDataType.BUFFER}],
+            },
+        },
+        commandsResponse: {},
+    },
     manuSpecificSinope: {
         ID: 65281,
         manufacturerCode: ManufacturerCode.SINOPE_TECHNOLOGIES,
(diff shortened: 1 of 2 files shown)


Koenkk:
The intention is not add manuf specific cluster to zigbee-herdsman anymore, could you use the new custom cluster API instead? Example:

danieledwardgeorgehitchcock:
@Koenkk - that is the intention while also porting across and extending the existing functionality held in clusters.js to the new format too

chrivers:
@Koenkk yes - what @danieledwardgeorgehitchcock said :)

I hope we can convince you to make this one, minor, temporary addition, which will allow me to implement Entertainment Mode, and give @danieledwardgeorgehitchcock some time to improve things all around.

If you'd agree to merge this, we can remove it when we have updated the other (existing) manuSpecificPhilips* clusters. Deal? :-)

Nerivec:
It's even more important to move custom clusters that conflict to ZHC.

Koenkk:
I think it would be good to directly move it into ZHC (also for the cluster conflict). I think we can do it with a minimal amount of changes by just adding it to https://github.com/Koenkk/zigbee-herdsman-converters/blob/8dc1719896a1f12a532c2497f06d1059bfb1c06c/src/lib/philips.ts#L62 ?

chrivers:
Thanks for the feedback! I was hoping we could add this to the base clusters, but I respect that you have to make decisions that make things maintainable in the long run :+1: 

I wrote to you on discord, but I realized it's probably better to keep it here, so everybody can participate - so I'll just copy what I wrote there:

...

I'm trying to wrap my head around https://github.com/Koenkk/zigbee-herdsman/pull/1322#issuecomment-2676361563 - do you mean adding the changes from https://github.com/Koenkk/zigbee-herdsman/pull/1322/files#diff-b186f965ae516aa02794eaf5db244e40d3a6ee0ab33c7251590f7eafc8f4e3f6 into `philipsTz`, with some modifications?

I'm still trying to completely grok how this works 🙂

You points to line 62 in https://github.com/Koenkk/zigbee-herdsman-converters/blob/8dc1719896a1f12a532c2497f06d1059bfb1c06c/src/lib/philips.ts#L62 but I'm not quite sure how that would work. Do you mean like line 61? (where converters are pushed to `philipsTz`?)

chrivers:
I talked to @Koenkk and @danieledwardgeorgehitchcock on discord, and we've come to the conclusion that PR doesn't make sense in the current form, so I'll close it.

@danieledwardgeorgehitchcock will return with a better solution, in the style that @Koenkk suggests. Here's a preview of that:

https://gist.github.com/danieledwardgeorgehitchcock/eba5e6127f8f757131e4dba6f4a74daa

PR #1326, opened 2025-02-24
Move manuSpecificPhilips Clusters To modernExtends

danieledwardgeorgehitchcock:
Removing manuSpecificPhilips clusters as these will be replaced by modernExtends in Koenkk/zigbee-herdsman-converters#8882


Nerivec:
Have to remove the names from the list too:
https://github.com/Koenkk/zigbee-herdsman/blob/572c4b8f4580da9c0a353b50f9a155c4a360d19e/src/zspec/zcl/definition/tstype.ts#L107
_And might need a few tests update if any use these for example._

PR #1328, opened 2025-02-28
Fix Heiman device parsing by handling CHAR_STR as buffer

Pastor02:
**Workaround:** Parse Heiman CHAR_STR attributes as buffer for correct data handling

### DIFF
--- a/src/zspec/zcl/zclFrame.ts
+++ b/src/zspec/zcl/zclFrame.ts
@@ -220,7 +220,16 @@ export class ZclFrame {
                         if (entry.dataType === DataType.CHAR_STR && entry.attrId === 65281) {
                             // [workaround] parse char str as Xiaomi struct
                             options.dataType = BuffaloZclDataType.MI_STRUCT;
-                        }
+                        }else if (
+                            entry.dataType === DataType.CHAR_STR &&
+                            (entry.attrId === 61448 ||
+                              entry.attrId === 61449 ||
+                              entry.attrId === 61450) &&
+                            header.manufacturerCode == 4619
+                          ) {
+                            // [workaround] parse char str as heiman buffer
+                            options.dataType = BuffaloZclDataType.BUFFER;
+                          }
                     }
 
                     entry[parameter.name] = buffalo.read(parameter.type, options);
--- a/src/zspec/zcl/zclFrame.ts
+++ b/src/zspec/zcl/zclFrame.ts
@@ -220,16 +220,14 @@ export class ZclFrame {
                         if (entry.dataType === DataType.CHAR_STR && entry.attrId === 65281) {
                             // [workaround] parse char str as Xiaomi struct
                             options.dataType = BuffaloZclDataType.MI_STRUCT;
-                        }else if (
+                        } else if (
                             entry.dataType === DataType.CHAR_STR &&
-                            (entry.attrId === 61448 ||
-                              entry.attrId === 61449 ||
-                              entry.attrId === 61450) &&
+                            (entry.attrId === 61448 || entry.attrId === 61449 || entry.attrId === 61450) &&
                             header.manufacturerCode == 4619
-                          ) {
+                        ) {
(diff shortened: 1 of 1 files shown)


Nerivec:
Is there an issue linked to this, or can you provide some background?

It seems like something that should be in converters (utf8<>buffer handling), not in ZH.

Leo2442926161:
Hi @Nerivec , 
Thank you for point out.

yes, it is an issue for Heiman fall down sensor, considering the maximum of zcl packet, we have to fill the data in that attribute as much as possible, so the data type is 0x42(Cstring), but we actually filled the data as hex format in that attribute.

when we tried to parer the data in converter, the data is wrong because the data was misinterpret in ZH, the ZH thought the data is string, but it is not, so something is wrong at that part.

for the pr fix, we checked the manufacture code(0x120b Heiman) and the specific attributes for fall down detection, it won't affect any of other part.

Nerivec:
Workarounds in ZH should be kept to a minimum, especially in places called so very often, and for cases this specific.
The ZHC converter for the affected device(s) should be used to convert the string back to whatever format that device uses.
At first glance, from your comments, something like this `Buffer.from(data, 'utf8')` should do.
I'll let you guys create a PR in ZHC instead.

Leo2442926161:
Hi @Nerivec , thanks for clarifying, I did in ZHC but it didn't work, that is why we made this PR.

we tried all of encodings, none of them convert it correctly. would you mind take a deep look? thank you very much.
```
            const encodings = [
              "hex",
              "latin1",
              "utf8",
              "ascii",
              "base64",
              "binary",
              "ucs2",
              "utf16le",
            ];
```

the raw data is:
0a 30 f8 d0 07 30 f8 d0 07 f0 0a
where:
42: zcl Cstring type)
0a: length of payload
30 f8 d0 07 30 f8 d0 07 f0 0a: the data


we tried it with below encodings, we actually got data as below, none of them matched the raw data from fall detector sensor.
Encoding: hex, Buffer: <Buffer >
Encoding: latin1, Buffer: <Buffer 30 fd fd 07 30 fd fd 07 fd 0a>
Encoding: utf8, Buffer: <Buffer 30 ef bf bd ef bf bd 07 30 ef bf bd ef bf bd 07 ef bf bd 0a>
Encoding: ascii, Buffer: <Buffer 30 fd fd 07 30 fd fd 07 fd 0a>
Encoding: base64, Buffer: <Buffer d3>
Encoding: binary, Buffer: <Buffer 30 fd fd 07 30 fd fd 07 fd 0a>
Encoding: ucs2, Buffer: <Buffer 30 00 fd ff fd ff 07 00 30 00 fd ff fd ff 07 00 fd ff 0a 00>
Encoding: utf16le, Buffer: <Buffer 30 00 fd ff fd ff 07 00 30 00 fd ff fd ff 07 00 fd ff 0a 00>

Koenkk:
Could you try using the [custom cluster API](https://github.com/Koenkk/zigbee-herdsman-converters/blob/a32df6625f31f9e2d9cc6305971b6f5b022cd166/src/devices/sonoff.ts#L41) to override the data type?

Nerivec:
The problem with using the utf8 data type is that it enforces strict use of utf8. Since you are using it for non-utf8 data it won't work "as-is" (non-utf8 is automatically replaced by `U+FFFD` character by NodeJS, indicating non-compliance).
Solutions would be to either:
- as Koenkk suggested, set the data type using a custom cluster for the device, [example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/a32df6625f31f9e2d9cc6305971b6f5b022cd166/src/devices/bosch.ts#L2004-L2021)
- fix the device firmware to use proper data types before it is released (`0x41` is an octet string, see [ZCL 2.6.2.13 Octet String](https://github.com/Koenkk/zigbee-herdsman/wiki/References#csa-zigbee-alliance-spec), which has no specific format, hence always parsed as Buffer by ZH)

PR #1330, opened 2025-03-02
fix: endpoint device & group members typing

Nerivec:
- `endpoint.getDevice()` could potentially be undefined => log error/stack trace
- `group._members` required array conversion for use in some places => use array as base instead (with check on add)

@Koenkk I'm a little worried that this didn't trip coverage, smells like hidden coverage issues, particularly for `Group`.
The `members` getter for `Group` is a little awkward. It also results in lots of chained calls because more often than not, Z2M will chain the result with some other filters on top. And typing is crappy too, because `getDevice` shouldn't be undefined for resulting items since that's the whole purpose of the getter. See any easy way to change this?

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -119,7 +119,6 @@ export class Endpoint extends Entity {
 
     get configuredReportings(): ConfiguredReporting[] {
         const device = this.getDevice();
-        assert(device, `Cannot get configured reportings for unknown/deleted device ${this.deviceIeeeAddress}`);
 
         return this._configuredReportings.map((entry, index) => {
             const cluster = Zcl.Utils.getCluster(entry.cluster, entry.manufacturerCode, device.customClusters);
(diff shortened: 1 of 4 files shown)


src/controller/model/endpoint.ts
```diff
@@ -165,8 +168,8 @@ export class Endpoint extends Entity {
     /**
      * Get device of this endpoint
      */
-    public getDevice(): Device {
-        return Device.byIeeeAddr(this.deviceIeeeAddress)!; // XXX: no way for device to not exist?
+    public getDevice(): Device | undefined {
+        return Device.byIeeeAddr(this.deviceIeeeAddress);
```
Koenkk, line 172: Shouldn't we just throw an exception when `device === undefined`?

Nerivec: There are places where `undefined` was being checked against the return value, even though the type didn't allow it, so, I don't think we can just throw.

https://github.com/Koenkk/zigbee-herdsman/blob/c9e3fd12c771c5825fa2caa5d3742a9abe1bfac0/src/controller/model/group.ts#L31

Koenkk: And how can the device get `undefined`?

Nerivec: Should be only if deleted
https://github.com/Koenkk/zigbee-herdsman/blob/c9e3fd12c771c5825fa2caa5d3742a9abe1bfac0/src/controller/model/device.ts#L634-L638

Koenkk (approved):

PR #1331, opened 2025-03-02
Setting zboss adapter to always use LUMI manufacturer code

captainlettuce:
The fix for [this](https://github.com/Koenkk/zigbee2mqtt/issues/20740) seems to have been to set the manufacturer code to match the incoming code for the specific vendor.

Since the zboss library doesn't set the manufacturerCode anywhere i assume it's fine to just set it globally? 

According to the Nordic documentaation for ZBOSS the NCP code on the micro-controller transparently handles the node_desc messages without interaction from the host, so a case-by-case implementation would require custom firmware.

For now i think that just setting the ManufacturerCode is an okay workaround, yes?

Side-note: Some documentation in zigbee2mqtt on how to override the zigbee-herdsman version would be A+ from a dev-standpoint.


Nerivec:
The default manuf code would be set in firmware, ZH just doesn't appear to override it (likely not needed).
This could lead to other, hard-to-detect, bugs for other manufacturers. I'd say best to set it when a LUMI device joins and override it back on network close. See example for `ember`:
https://github.com/Koenkk/zigbee-herdsman/blob/915347fa706a5a4fe88fa1ce767d12f9568289ef/src/adapter/ember/adapter/emberAdapter.ts#L619-L639
https://github.com/Koenkk/zigbee-herdsman/blob/915347fa706a5a4fe88fa1ce767d12f9568289ef/src/adapter/ember/adapter/emberAdapter.ts#L1867-L1872

---

> Side-note

Package linking would be the easiest way.
https://pnpm.io/cli/link
But there are a lot of different setups out there, different approaches...

Koenkk:
@kirovilya please take a look

Nerivec:
I don't think this [last commit](https://github.com/Koenkk/zigbee-herdsman/pull/1331/commits/37718f909317df4c26b2a213ee3b4afce13029b8) works.
The code has to be set when a device requests joining, so that when the device requests the coordinator's node descriptor it fakes being a lumi hub (that request is between device<>coordinator, Z2M is not involved, doesn't even see it).
It should be set whenever the coordinator informs Z2M of `device joined`, and then can be reset later by some other appropriate event.

captainlettuce:
@Nerivec you are correct. Thanks for the hints, I'll give it a try. 

I pushed the commit to this branch for testing, I'll mark the PR as draft until I can verify it works.

captainlettuce:
image
image

Confirmed working and i can still join other devices as well

Nerivec:
You'll want to add the secondary IEEE prefix for lumi too.
See https://github.com/Koenkk/zigbee-herdsman/pull/1334/files

I also moved away from using permit join to disable, since it won't trigger automatically anymore (unless explicitly triggered by user), meaning it won't reset in a lot of cases. Any chance you have another event like ember (stack closed) that you can use?

captainlettuce:
I added the second manufacturer code as requested.

> I also moved away from using permit join to disable, since it won't trigger automatically anymore (unless explicitly triggered by user), meaning it won't reset in a lot of cases. Any chance you have another event like ember (stack closed) that you can use?

I'm not sure how the ember adapter works or how it differs from the zboss stack. From a cursory glance it seems that `ZIGBE_NETWORK_CLOSED` is when closing the actual network, not a specific device leaving, right?

I looked through the [zboss spec](https://wiki.homed.dev/files/9/95/ZBOSS_NCP_Serial_Protocol.pdf) and can't find anything similar, the only "events" (called indications in zboss) are for when devices leave, which we can not use since we don't know if we have more Lumi devices that are still connected. (?)

My thinking is that the node descriptor should mainly be requested when joining, and since we reset before allowing join it should not be any problems for currently joined devices if the manufacturer code is stuck as Lumi? 

If that is not the case please help me pinning down what actually needs to be done.

Nerivec:
It should mostly be fine I think. It creates a few edge cases, but hopefully not too problematic. Technically, proper devices should not use the manuf code as a barrier like these crappy Aqara devices... so, that, in itself, should limit the impact. It just creates a state that makes debugging a lot harder in case it is ever a problem.
@kirovilya can help better than me on that front though. I don't know the zboss stack well at all.

kirovilya:
In my opinion, it is enough to set the manufacturer's code only for the interview process - Lumi/Aqara devices do not require permanent of this code. Therefore, I consider @captainlettuce code quite logical.

Nerivec:
Current code will make the change permanent in a lot of cases, because it only resets when permit join is triggered by the user. If permit join is let to expire, it won't reset, meaning it will keep the workaround manuf code until the next permit join is triggered.
While that shouldn't create issue with proper devices, as mentioned, it creates a hard-to-debug state (floating state, race conditions, etc.).

Also, do these requests need to be queued in adapter class? I can see zdo/zcl to endpoint are queued, gcast/bcast aren't, so, not sure.

captainlettuce:
> If permit join is let to expire, it won't reset, meaning it will keep the workaround manuf code until the next permit join is triggered.

I'm not so sure this is the case. When a user clicks `Permit join` the device will reset its manufacturer code to the default. If a Lumi device joins it will set it until the next time `Permit join` is called. 

If (another) device were to stop working, the first thing I imagine the user doing is repairing, and thereby resetting the workaround. That in combination with the logs makes this hypothetical bug much less severe. Just my 2c.

I'm not sure what determines if a request needs to be queued or not, but i have confirmed this to be working with an actual adapter (nrf52840)

Nerivec:
You might also get cases of race conditions with the current logic, where the manuf code is reset and right around the same time, an Aqara already in pairing mode joins, and triggers the workaround. Since we have no control over when _exactly_ the device actually requests the node desc. from the coordinator, it could result in cases where the workaround doesn't... work. It's too dependent on processing speed, Z2M, serial, firmware...
In short, I think the reset in the permit join trigger is "too close" to a potential device join.
_It's an edge case for sure, but it's not worth the nightmares required to debug a scenario like this, best to properly implement the fix._

captainlettuce:
I understand your motivation and reasoning. However without any clear path forward and a current working solution I feel I have achieved what i set out to do: get my devices working, so unless someone can point me in a direction, my path ends here.

If you chose to not merge this code, I will just build my own image for my own use, so it's not really a problem for me if you chose to wait and do it yourselves.

Thanks for your time

Koenkk:
Thanks!

Issue #1342, opened 2025-03-12
Add support for encrypted Zigbee Green Power devices


Timu5:
Currently, the project does not support encrypted Zigbee Green Power (ZGP) devices. According to the ZGP specification v1.0a(pdf easly accesible online), when the NWK header contains 0b11 as the security level, the payload is encrypted. The decryption process is described in A.1.5.4.4, with a step-by-step example in A.1.5.5.4.

The encryption algorithm used is the same as the one already implemented in the project. Adding support for encrypted ZGP devices would enhance compatibility and allow more devices to function properly within the ecosystem.

Decryption might take place in onZclGreenPowerData or onZclPayload(before calling onZclGreenPowerData)
https://github.com/Koenkk/zigbee-herdsman/blob/5669c216cfc0a4575c7d6713b7f4a710f7626913/src/controller/greenPower.ts#L111
https://github.com/Koenkk/zigbee-herdsman/blob/4657ecf9c3f13dd9599560ba7a1d278540a0320c/src/controller/controller.ts#L799

This will allow at least bellow issues to be resolved:
https://github.com/Koenkk/zigbee2mqtt/issues/19405
https://github.com/Koenkk/zigbee2mqtt/issues/23618

PR #1349, opened 2025-03-16
fix: Green spring cleaning

Nerivec:
- https://github.com/Koenkk/zigbee-herdsman/commit/5b6fcb07bf17953d62fd435e4f90fa814ca4e32f
  - Improve Green Power logging
  - Improve typing
  - Fix Channel Request incorrect parameters
  - Fix missing/misused parameters in GP cluster
  - Fix misused parameters in GP cluster

- https://github.com/Koenkk/zigbee-herdsman/commit/03fe2bc4134206018d83476d1ce3daaf8d0fcf4b
  - Implement decrypting GPDF that come in as `UNPROCESSED` and with `FULLENCR`
  - Implement passing of correct GP `options` from deconz/ember
  - Fixes https://github.com/Koenkk/zigbee-herdsman/issues/1342

> [!IMPORTANT]
> FULLENCR mode is rarely used, and devices that do (namely Moes ZEU2S), appear to misbehave in other ways (non-fixed source ID?)... Implementation is purely based on data gathered from https://github.com/Koenkk/zigbee2mqtt/issues/19405#issuecomment-2727338024 (ember) and the GP spec. It needs testing with actual devices. Also, with the data being limited compared to having the full NWK GP frame, this will only work if the stack passes the proper information. The codepath is guarded as much as possible, so, worse case scenario, it just shouldn't trigger.

@Koenkk storing of the `gpSecurityKey` in `device.meta` isn't all that great, but since devices pass it during commissioning and it's not available in following `UNPROCESSED` frames, we need to store it somehow...?

CC: @chris-1243


Koenkk:
> but since devices pass it during commissioning and it's not available in following UNPROCESSED frames, we need to store it somehow...?

`.meta` is more meant for apps like z2m, maybe we can add a `.greenPowerData` to the device? (not saved to database.db when the object is not set)

Nerivec:
It's really just that key that's the problem, don't think it's worth adding another object. Maybe directly as `device.gpSecurityKey`?

Koenkk:
Also fine!

Koenkk:
🚀

PR #1352, opened 2025-03-19
Fixed problem with non-latin characters in the UTF8 string.

ptvoinfo:
Fixed problem with non-latin characters in the UTF8 string.


ptvoinfo:
@Koenkk  I'm not sure if I should fix the code style issue in ".github/workflows/ci.yml"

Koenkk:
Thanks!

PR #1359, opened 2025-03-22
fix: Green Power decryption when received via commissioning notif

Nerivec:
Based on feedback from zstack users with the Moes ZEU2S.

NOTE: Looks like either zstack is incorrectly wrapping the frames, or the GPP is not sending them properly (using `commissioning notification` instead of regular `notification`).
Guarded as much as possible to ensure no decryption unless we're reasonably certain it is needed (so, passed as many `options` as possible for other stacks too).

CC: @MaxwellJK


Koenkk:
Thanks!

PR #1369, opened 2025-03-26
fix: Green power parsing issue with payload size

Nerivec:
_Hijacking my own PR... Better to fix bugs than logging only_ 😬

Found an issue with GDP_FRAME parsing where, if the spec changes (like it did in 1.1.1, which apparently Silabs router firmware has support for), the parsing will result in scrambled footer (GPP data) because it was only offset by the option-based parsing instead of actual payload size.
GP spec 1.1.1 introduced the "switch info" field in commissioning, which was not supported by ZH yet, hence, bug triggered... _Note that only devices with ID 0x07 (like PTM216Z) actually are supposed to have this new field._
_Tests added to CI, and tested with a PTM216Z and a Silabs-coordinator-flashed-as-router._
Before:
```
zh:controller:greenpower: [COMMISSIONING] srcID=22369751 gpp=1282 rssi=2 linkQuality=Poor
```
After:
```
zh:controller:greenpower: [COMMISSIONING] srcID=22369751 gpp=57129 rssi=23 linkQuality=Excellent
```

Also fixed an issue in ZoH where the profile ID was not set properly, resulting in GPP ignoring commissioning mode commands.

---

- cleanup some unnecessary text in ezsp layer of `ember` driver (reduce log size a bit)
- hide security key for GP in logs
- standardize log string for GP
  - make it easier to identify direct/GPP comm
  - use source ID instead of relying on data from driver that isn't necessarily consistent from driver to driver (_also useful for sniffs, to correlate with logs_)
  - if GPP, make use of extra metadata

Should produce logs like this:
```
[COMMISSIONING] srcID=22410362 gpp=28566 rssi=23 linkQuality=Excellent
[PAIRING] srcID=22410362 gpp=28566 options=58696 (addSink=true commMode=2) wasBroadcast=true
Green power device '{"sourceID":22410362,"deviceID":2,"networkAddress":62586,"securityKey":{"type":"Buffer","data":HIDDEN}}' joined
New green power device '0x000000000155f47a' joined
```
```
[UNHANDLED_CMD/PASSTHROUGH] command=0x20 srcID=2777252112 gpp=24404 rssi=17 linkQuality=Excellent
```
```
[UNHANDLED_CMD/PASSTHROUGH] command=0x21 srcID=1496140231 gpp=NO
```

CC: @chris-1243

### DIFF
--- a/src/adapter/ember/ezsp/ezsp.ts
+++ b/src/adapter/ember/ezsp/ezsp.ts
@@ -2134,7 +2134,7 @@ export class Ezsp extends EventEmitter<EmberEzspEventMap> {
      * @param tokenAddress uint16_t The address of the stack token that has changed.
      */
     ezspStackTokenChangedHandler(tokenAddress: number): void {
-        logger.debug(`ezspStackTokenChangedHandler(): [tokenAddress=${tokenAddress}]`, NS);
+        logger.debug(`ezspStackTokenChangedHandler: tokenAddress=${tokenAddress}`, NS);
     }
 
     /**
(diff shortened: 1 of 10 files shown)


chris-1243:
It solved

https://github.com/Koenkk/zigbee2mqtt/issues/23959

PR #1376, opened 2025-03-30
fix: Fix Conbee adapter having no endpoints

Koenkk:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/26817

Not a nice solution, but for some reasons some don't respond causing failures in Z2M. I've checked with `0x26780700` (same as the one from the issue)


Nerivec:
It's getting pretty hard to keep track of what deconz version has what bugs or missing implementations...
Maybe should add 0xf2 to the list too (like the hardcoding before)?

Koenkk:
> Maybe should add 0xf2 to the list too (like the hardcoding before)?

This doesn't seem to be correct, in my case it has ep 1 and 80.

PR #1379, opened 2025-04-01
remove custom cluster heimanSpecificAirQuality

jeffmhastings:
This is related to https://github.com/Koenkk/zigbee-herdsman-converters/pull/9104 and removes the heimanSpecificAirQuality cluster from the zcl definitions. These are being moved over to zigbee-herdsman-converters since there are conflicts with that cluster ID.


jeffmhastings:
Yeah, I can do that too. I wasn't sure what the scope should be. I haven't had much time to look into this recently but I will try to update this in the next week or so.

PR #1380, opened 2025-04-01
fix: Enable serial port locking by default for ZiGate

puddly:
`lock: true` is the default for the serial port library and all other adapters.


Koenkk:
Apologies for the delay, somehow I missed it!

PR #1381, opened 2025-04-02
feat: add Tunneling cluster commands

avzasorin-sd:
Extend cluster.ts with mandatory commands of Tunneling cluster, which are described in ZCL specification ( ZCLv8 10.6.2.4 and 10.6.2.5 )

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -3452,7 +3452,7 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
                 response: 0,
                 parameters: [
                     {name: 'protocolId', type: DataType.ENUM8},
-                    {name: 'mfgCode', type: DataType.UINT16},
+                    {name: 'manufCode', type: DataType.UINT16},
                     {name: 'flowControl', type: DataType.BOOLEAN},
                     {name: 'mtuSize', type: DataType.UINT16},
                 ],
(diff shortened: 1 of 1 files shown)


src/zspec/zcl/definition/cluster.ts
```diff
@@ -3447,22 +3447,58 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
         ID: 0x0704,
         attributes: {},
         commands: {
+            requestTunnel: {
+                ID: 0,
+                response: 0,
+                parameters: [
+                    {name: 'protocolID', type: DataType.ENUM8},
```
Koenkk, line 3454:
```suggestion
                    {name: 'protocolId', type: DataType.ENUM8},
```

To keep it consistent with other `Id`s in this file (also change other occurrences below)

avzasorin-sd: There was `tunnelID` in description of `transferData` command already, it was added before my PR.
Should I change only `protocolID` to `protocolId`, because `tunnelID` key is used by easyiot.ts converter - and I was following this notation as an example.

Koenkk: Let's also update `easyiot.ts` for consistency purposes.

avzasorin-sd:
Done with naming style. Created related PR to ZHC, because otherwise it will break `easyiot.ts` converter
https://github.com/Koenkk/zigbee-herdsman-converters/pull/9118

Koenkk:
Thanks!

PR #1386, opened 2025-04-06
fix: improve Green Power processing

Nerivec:
- always use IEEE to lookup GP devices to prevent possible nwk address conflicts (mocked from source ID, so can't be sure)
- fix processing of non-command GP frames (commandsResponse, defaultRsp...), should not go through GP processing
- fix pairing options not derived from commissioning data
- add test to cover unsupported IEEE addressing
- debug log `defaultRsp` with non-SUCCESS status (_should allow identifying a lot more potential problems unrelated to GP too whenever responses are not explicitly checked_)
- rename `GreenPower.onZclGreenPowerData` to `GreenPower.processCommand`

Also:
- Fix typing in controller.tests.ts (was hiding a few outdated properties, and was just annoying 😁)

---

Following up on https://github.com/Koenkk/zigbee2mqtt/issues/25620 it appears the "proper" logic would be to not consider a GP device joined if `pairing` command returns non-success status through `Foundation.defaultRsp` when sent as unicast (the device appears to stay in commissioning and notification frames eventually get dropped when commissioning mode is exited). I did **not** add this behavior, as it would change the logic quite drastically and I don't have enough stacks/GP devices to confirm all-around impact (would need a GPP that does this in the first place...).

CC: @chris-1243

### DIFF
--- a/src/adapter/ember/ezsp/ezsp.ts
+++ b/src/adapter/ember/ezsp/ezsp.ts
@@ -8610,7 +8610,7 @@ export class Ezsp extends EventEmitter<EmberEzspEventMap> {
 
         const messageContents = Buffer.concat([gpdHeader, gpdCommandPayload]);
 
-        // XXX: BROADCAST currently hardcoded to match upstream codepath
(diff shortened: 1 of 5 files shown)


Koenkk:
Thanks!

PR #1396, opened 2025-04-25
fix: add support for generic science-related clusters

Nerivec:
Couple there at the bottom I marked, `pm1Measurement`, `pm10Measurement`, these are not ZCL clusters?


Koenkk:
Thanks!


> Couple there at the bottom I marked, pm1Measurement, pm10Measurement, these are not ZCL clusters?

I don't see them being used so can probably remove them (should be in there if they are not ZCL clusters)

PR #1408, opened 2025-05-19
feat: Add conditional fieldControl fields to genOta commands

Koenkk:
For https://github.com/Koenkk/zigbee-herdsman-converters/issues/9356

CC: @sjorge


Nerivec:
👍 

Got busy with 2.0 and forgot about this minimumBlockPeriod. I wanted to take a closer look, see if that could lessen the massive amount of timers that current OTA generates...
Probably should look into moving OTA to ZH too.

PR #1410, opened 2025-05-21
fix: remove global BigInt.toJSON override and introduce explicit jsonReplacer

kubens:
This change removes the global patch on BigInt.prototype.toJSON and replaces it with an explicit JSON replacer function (jsonReplacer) that herdsman can opt into whenever it needs to serialize BigInts. By doing so we eliminate side-effects on consumer code (e.g. Matter.js) that expect native BigInt behavior and manage their own serialization logic.

**Problem**

- Herdsman was monkey-patching BigInt.prototype.toJSON = () => this.toString(), which affects every JSON.stringify call in the process.
- Consumers that rely on raw BigInt values (or their own replacers/revivers) can no longer parse or stringify objects in a compatible way, leading to data corruption or errors.

**Solution**
- Remove the global override of BigInt.prototype.toJSON.
- Introduce a jsonReplacer helper that converts any BigInt values to strings.

### DIFF
--- a/src/zspec/zcl/zclFrame.ts
+++ b/src/zspec/zcl/zclFrame.ts
@@ -1,4 +1,4 @@
-import {jsonReplacer} from "src/utils/json";
+import {jsonReplacer} from "src/utils";
 
 import {BuffaloZcl} from "./buffaloZcl";
 import {BuffaloZclDataType, DataType, Direction, FrameType, ParameterCondition} from "./definition/enums";
--- a/src/zspec/zcl/zclFrame.ts
+++ b/src/zspec/zcl/zclFrame.ts
@@ -1,4 +1,4 @@
-import "../../utils/patchBigIntSerialization";
+import {jsonReplacer} from "src/utils/json";
 
 import {BuffaloZcl} from "./buffaloZcl";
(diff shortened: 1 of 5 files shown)


src/zspec/zcl/zclFrame.ts
```diff
@@ -1,4 +1,4 @@
-import "../../utils/patchBigIntSerialization";
+import {jsonReplacer} from "src/utils";
```
Nerivec, line 1: Should use relative path import.

Nerivec:
There are probably more places where this would have to be changed (lots of BigInt are hidden behind auto-parsing). And any place where this is not changed, and not try/catch'ed could hard-crash the Z2M process. i.e. this is a dangerous change that requires a full check across all repos.

kubens:
Thanks for the valid concern — you’re absolutely right that removing BigInt.prototype.toJSON outright can be risky, especially if there are still places in the codebase that rely on implicit BigInt serialization.

As a safer intermediate step, I’m thinking of keeping the patch for now but modifying it to include a deprecation warning when it’s used. That way, we can avoid hard crashes while also making it clear that this behavior is deprecated and will be removed in a future version.

In the meantime, I’ll continue working off my own fork to experiment with this change and map out the impact across repos before proposing full removal. Let me know if you’re okay with this direction.

PR #1418, opened 2025-06-06
feat: Make profileId overwritable per frame to support custom Shelly clusters

SvenKirschbaum:
This is a companion PR to [9468](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9468).

This PR exposes a new option, making it possible to override the profileId contained in the APS frame individually.
As zStack does not support changing the profileId directly, in case the profileId but no src endpoint is set, an endpoint with the correct profile is choosen automatically. If the requested profileId cannot be satisfied, an exception is thrown.

Please note, that i only own ember based adapters, so the changes to the other adapters are untested.

I have also added the following Shelly specific changes:
- Added the manufacturer Code
- Added a new endpoint with appprofid 0xc001
- Added profile 0xc001 to the ezspIncomingMessageHandler filter


src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -433,14 +434,23 @@ export class ZStackAdapter extends Adapter {
         disableResponse: boolean,
         disableRecovery: boolean,
         sourceEndpoint?: number,
+        profileId?: number,
     ): Promise<Events.ZclPayload | undefined> {
+        // If sourceEndpoint is not provided, try to select the source endpoint based on the profileId. If no profileId is provided, use endpoint 1.
+        const srcEndpoint = sourceEndpoint ?? Endpoints.find((e) => e.appprofid === profileId)?.endpoint ?? 1;
```
Koenkk: `Endpoints.find((e) => e.appprofid === profileId)?.endpoint` should only be done when `profileId !== undefined`

src/adapter/ember/ezsp/ezsp.ts
```diff
@@ -5305,7 +5305,10 @@ export class Ezsp extends EventEmitter<EmberEzspEventMap> {
         } else if (
             apsFrame.profileId === ZSpec.HA_PROFILE_ID ||
             apsFrame.profileId === ZSpec.WILDCARD_PROFILE_ID ||
-            (apsFrame.profileId === ZSpec.GP_PROFILE_ID && type !== EmberIncomingMessageType.BROADCAST_LOOPBACK)
+            (apsFrame.profileId === ZSpec.GP_PROFILE_ID && type !== EmberIncomingMessageType.BROADCAST_LOOPBACK) ||
+            // TODO: Better (dynamic) way to handle this?
+            // Shelly WifiSetupCluster uses profileId 0xc001
+            apsFrame.profileId === 0xc001
```
Nerivec: Should const that value somewhere, and import it where needed, avoid magic numbers spread everywhere.

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -1953,7 +1954,7 @@ export class EmberAdapter extends Adapter {
         }
 
         const apsFrame: EmberApsFrame = {
-            profileId: sourceEndpointInfo.profileId,
+            profileId: profileId ?? sourceEndpointInfo.profileId,
```
Nerivec: Should skip the `sourceEndpointInfo` logic entirely if profileId is present.

Nerivec:
We should add complete generic support and add the variables to group/broadcast too.
Not sure how these Shelly work, if they might need it or not, but at least it covers future needs (and keeps the API consistent).

src/adapter/z-stack/adapter/zStackAdapter.ts
SvenKirschbaum: I've changed the code flow accordingly, and also moved the logic into its own function, so it can be reused for group and broadcast frames

src/adapter/ember/ezsp/ezsp.ts
SvenKirschbaum: I have added a constant to `src/zspec/consts.ts`

src/adapter/ember/adapter/emberAdapter.ts
SvenKirschbaum: I have changed the code flow accordingly.

SvenKirschbaum:
> We should add complete generic support and add the variables to group/broadcast too. Not sure how these Shelly work, if they might need it or not, but at least it covers future needs (and keeps the API consistent).

Support for group and broadcast frames was added.

src/adapter/ember/ezsp/ezsp.ts
Nerivec: Is it a CSA-assigned profile ID? If not, I suggest we name it `CUSTOM_...`.
Like we do for manufacturer codes:
https://github.com/Koenkk/zigbee-herdsman/blob/2da206da4b4376c22b192d4d15a6e4bef389e996/src/zspec/zcl/definition/manufacturerCode.ts#L724-L729

(That goes for the manuf code that was added in this PR too, not sure if it's registered or just self-assigned by Shelly)

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -2090,10 +2101,16 @@ export class EmberAdapter extends Adapter {
...
-        const sourceEndpointInfo = FIXED_ENDPOINTS.find((epi) => epi.endpoint === sourceEndpoint) ?? FIXED_ENDPOINTS[0];
+        let resolvedProfileId = profileId;
+        if (resolvedProfileId === undefined) {
+            const sourceEndpointInfo = (sourceEndpoint && FIXED_ENDPOINTS.find((epi) => epi.endpoint === sourceEndpoint)) || FIXED_ENDPOINTS[0];
+            resolvedProfileId = sourceEndpointInfo.profileId;
+        }
+
         const apsFrame: EmberApsFrame = {
-            profileId: sourceEndpointInfo.profileId,
+            profileId: resolvedProfileId,
```
Nerivec: Should be able to one-line these:
```ts
profileId: profileId ?? ((sourceEndpoint && FIXED_ENDPOINTS.find((epi) => epi.endpoint === sourceEndpoint)) || FIXED_ENDPOINTS[0]).profileId;
```

src/adapter/ember/ezsp/ezsp.ts
SvenKirschbaum: Is there any publicly accessible source to confirm this? I assume the manufacturer code is assigned to them, as they have (Matter) certified devices under it. I guess the profile id is custom though, and renamed it as suggested.

src/adapter/ember/adapter/emberAdapter.ts
SvenKirschbaum: Changed.

src/adapter/deconz/adapter/deconzAdapter.ts
```diff
@@ -492,9 +493,9 @@ export class DeconzAdapter extends Adapter {
             requestId: transactionID,
             destAddrMode: PARAM.PARAM.addressMode.GROUP_ADDR,
             destAddr16: groupID,
-            profileId: 0x104,
+            profileId: profileId ?? 0x104,
             clusterId: zclFrame.cluster.ID,
-            srcEndpoint: 1,
+            srcEndpoint: sourceEndpoint || 1,
```
Koenkk:
```suggestion
            srcEndpoint: sourceEndpoint ?? 1,
```

src/zspec/consts.ts
```diff
@@ -16,6 +16,8 @@ export const GP_PROFILE_ID = 0xa1e0;
 export const TOUCHLINK_PROFILE_ID = 0xc05e;
 /** The profile ID used to address all the public profiles. */
 export const WILDCARD_PROFILE_ID = 0xffff;
+/** The profile ID used to access Shelly devices custom clusters. */
+export const CUSTOM_SHELLY_PROFILE_ID = 0xc001;
```
Koenkk, line 20: Since it is only used in the `ezsp`, let's keep it there ( we won't want other to use it, and EZSP implementation is deprecated anyway)

src/adapter/ember/ezsp/ezsp.ts
Nerivec: Databases are not publicly available (only to other members), but...
You can search for members here:
https://csa-iot.org/members/search/?m_keywords=shelly
And for products here (you usually get the vendor ID if you open one):
https://csa-iot.org/csa-iot_products/?p_keywords=shelly

For profile IDs, manuf-specific are mostly intended for internal use (breaks interoperability between devices), which would seem to be the case here (it's intended for use by Shelly software with Shelly products), some notes even suggest they are not intended for commercial use (only for testing/demo)... So, probably custom indeed. 😞

src/zspec/consts.ts
Nerivec: It's used in `ember`. Also should be used in `src/adapter/z-stack/adapter/endpoints.ts` (a lot of magic numbers to replace still).
I assume the ZHC PR will also use it.

Nerivec:
It's a little strange that `ember` and `zoh` tests are passing without changes. That new function parameter should require more coverage.

https://github.com/Koenkk/zigbee-herdsman/blob/master/test/adapter/ember/emberAdapter.test.ts#L2497-L3284

https://github.com/Koenkk/zigbee-herdsman/blob/master/test/adapter/zoh/zohAdapter.test.ts#L560-L810

Hedda:
@SvenKirschbaum Off-topic but if you are by any chance interested in also contributing support for this to Home Assistant's ZHA integration as well then  please see this discussion in the zigpy/zha community:

* https://github.com/zigpy/zha/discussions/470

Koenkk:
Sorry for taking so long to merge, thanks!

PR #1421, opened 2025-06-11
Add new cluster fallDetectionOwon

Pastor02:
Add cluster for OWON FDS315 Fall sensor


Koenkk:
Please use the custom cluster API in the definition instead, example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/0185eb943c3b8c1dfc424cae935f2516eaf6736a/src/devices/sonoff.ts#L45

PR #1424, opened 2025-06-15
fix: more support for ZCL non-value

Nerivec:
Add ZCL-specific wrappers to BuffaloZcl to handle non-values, per ZigBee spec.
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/27102

TODO:
- [x] Current behavior is to use NaN for number non-values (including boolean since using 0/1). This must be checked against uses, and support added where needed.
- [x] https://github.com/Koenkk/zigbee-herdsman-converters/pull/9754
- [ ] zboss adapter uses BuffaloZcl for frame reading, this might break that... @kirovilya 

CC: @oddlama


test/controller.test.ts
```diff
@@ -3776,7 +3776,7 @@ describe("Controller", () => {
                             "65281": {
                                 "1": 3285,
                                 "10": 0,
-                                "100": 0,
+                                "100": false,
```
Koenkk: Also need to update the Aqara converters to be able to deal with this.

Nerivec: I reverted this.
That change was a bit too dangerous (too much typing lacking in ZHC/ZCL payloads). We can do it in a separate PR if the need arises.

Nerivec:
For ZHC, I guess most of the code should hit:
- tz https://github.com/Koenkk/zigbee-herdsman-converters/blob/b2e45275958f5418e07924800d253156ece1b48b/src/lib/modernExtend.ts#L2441
- fz https://github.com/Koenkk/zigbee-herdsman-converters/blob/b2e45275958f5418e07924800d253156ece1b48b/src/lib/modernExtend.ts#L2464

Both of these should allow using `Number.NaN` to represent non-value.
The current workaround was to use [value_ignore](https://github.com/Koenkk/zigbee-herdsman-converters/blob/b2e45275958f5418e07924800d253156ece1b48b/src/lib/modernExtend.ts#L2395) (due to lack of support at parser level), though it's only used in a couple places.

@Koenkk how do you think non-values should translate (ignore, push the equivalent of "N/A", etc.)? _At least for tz, it should not ignore the request anymore._

Also have to deal with converters that are not using modern extend yet (haven't looked if any).

Nerivec:
As discussed, zboss being experimental, if this introduces undesired behavior, it will have to wait for someone able to fix it (not sure what zboss expects for ZCL non-values, should likely move to using its own Buffalo subclass)

Koenkk:
Thanks!

PR #1438, opened 2025-07-01
fix: deCONZ: Handle unsupported status code for read parameter response

manup:
This happens when older firmware doesn't support a parameter.
E.g.: Seen with firmware version 0x26580700 which doesn't support reading the frame counter.

Fixes: https://github.com/Koenkk/zigbee2mqtt/issues/27864


gromitn:
Sorry I know this is really bad general knowledge (how GitHub even works) in this field of Home Assistant - but once this fix has got to this stage of passed/merged etc, at what point generally does it come out included in home assistant as an actual version update?

PR #1441, opened 2025-07-03
fix: deCONZ: ZGP implementation

manup:
The old ZGP frame parser used some magic numbers which often shoot over  the actual received buffer. In the new adapter implementation frames are exposed with their actual size, which as side effect caused the frame parser to have invalid negative offsets.

The new implementation is a port from deCONZ code without fixed magic numbers.

Tested with a old Hue Tap and Friends of Hue switch.

Note that the old implementation likely didn't support "joining" without ZGP proxies since any frame length >= 30 was treated as commissioning command (0x04), which the controller rejected.

Issue: https://github.com/Koenkk/zigbee2mqtt/issues/27888

### DIFF
--- a/src/adapter/deconz/driver/frameParser.ts
+++ b/src/adapter/deconz/driver/frameParser.ts
@@ -16,6 +16,7 @@ import {
     ParamId,
     type ReceivedDataResponse,
     type Request,
+    ZgpConstants,
 } from "./constants";
 import {apsBusyQueue, busyQueue} from "./driver";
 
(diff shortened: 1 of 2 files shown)


src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
-            commandId = view.getUint8(17);
-            commandFrameSize = view.byteLength - 18 - 6; // cut 18 from begin and 4 (sec mic) and 2 from end (cfc)
-            commandFrame = Buffer.from(view.buffer.slice(18, commandFrameSize + 18));
+        let srcId = 0;
+        let frameCounter = 0;
+        let commandId = 0;
+        let commandFrameSize = 0;
+        let commandFrame = null;
+
+        const buf = new Buffalo(Buffer.from(view.buffer));
```
Nerivec, line 475: Is the buffer copy needed here?

src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
+
+        // GPD SrcID field
+        if (frameType === ZgpConstants.GpNwkDataFrame && applicationId === 0) {
+            srcId = buf.readUInt32();
+        } else if (
+            frameType === ZgpConstants.GpNwkMaintenanceFrame &&
+            nwkFrameControl & ZgpConstants.GpNwkFrameControlExtensionFlag &&
+            applicationId === 0
+        ) {
+            srcId = buf.readUInt32();
```
Nerivec: Should combine in one `if`, same inner code.

src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
+                    commandFrame = Buffer.alloc(commandFrameSize);
+                    for (let i = 0; i < commandFrameSize; i++) {
+                        commandFrame[i] = buf.readUInt8();
+                    }
+                }
+                break;
+
+            default:
+                commandId = 0;
+                break;
```
Nerivec: Not needed, already 0 from before the `switch`.
Also might as well turn the `switch` to `if`.

src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
+        if (nwkFrameControl & ZgpConstants.GpNwkFrameControlExtensionFlag) {
+            nwkExtFrameControl = buf.readUInt8();
+        }
+
+        const options = nwkExtFrameControl | (nwkFrameControl << 8);
+
+        const applicationId = nwkExtFrameControl & 7;
+        const securityLevel = (nwkExtFrameControl >> 3) & 3;
+
+        if (applicationId > 2) {
```
Nerivec: Might want to put a few `const enum` (stripped out at compile time) at the top for some of these to avoid magic numbers (and comments).
_Feel free to copy from https://github.com/Nerivec/zigbee-on-host/blob/e6a56389bb51002ee32a77ab7117851e4b85475a/src/zigbee/zigbee-nwkgp.ts#L30-L59 ._

src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
+                    commandFrameSize = buf.getBuffer().length - buf.getPosition() - micSize;
+                    //logger.debug(`GPD payload length: ${commandFrameSize}, mic size: ${micSize}`, NS);
+                    if (commandFrameSize < 0) {
+                        logger.error(`GPD payload length < 0: ${commandFrameSize}`, NS);
+                        return null;
+                    }
+                    commandFrame = Buffer.alloc(commandFrameSize);
+                    for (let i = 0; i < commandFrameSize; i++) {
+                        commandFrame[i] = buf.readUInt8();
+                    }
```
Nerivec: This should do the same:
```ts
commandFrame = Buffer.from(buf.readBuffer(commandFrameSize)); // copy
```

src/adapter/deconz/driver/frameParser.ts
```diff
@@ -440,41 +441,152 @@ function parseDeviceStateChangedNotification(view: DataView): number | null {
...
-            srcId = view.getUint32(9, littleEndian);
-            frameCounter = view.getUint32(13, littleEndian);
-            commandId = view.getUint8(17);
-            commandFrameSize = view.byteLength - 18 - 6; // cut 18 from begin and 4 (sec mic) and 2 from end (cfc)
-            commandFrame = Buffer.from(view.buffer.slice(18, commandFrameSize + 18));
+        let srcId = 0;
+        let frameCounter = 0;
+        let commandId = 0;
+        let commandFrameSize = 0;
+        let commandFrame = null;
```
Nerivec:
```ts
let commandFrame: Buffer | undefined;
```
more explicit typing wouldn't hurt here I think (also need to change the null check to undefined at bottom).

src/adapter/deconz/driver/frameParser.ts
manup: At least for now, the incoming packet handler currently provides a `DataView` I'd like to refactor this in a upcoming PR to use plain `Buffer` and `Buffalo` for all parser functions. The DataView is used since we didn't understand the Buffalo interface earlier.

Nerivec: Ah right, the `.buffer` is an ArrayBuffer, forgot about that!
Sounds good about the refactor.

manup:
@Nerivec thanks a lot for the review :+1: The updated PR hopefully addresses all comments expect the initial buffer copy which I'd like to address in another PR to get rid of `DataView`.

Nerivec (approved):
👍 

Note: `enum` != `const enum` (only the latter is inlined at compile time) https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

manup:
> 👍
> 
> Note: `enum` != `const enum` (only the latter is inlined at compile time) https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

Good to know, guess there are quite a few places where this can be applied in future PRs.

PR #1445, opened 2025-07-06
fix: deCONZ: Joining bug in some setups and endpoint configuration

manup:
This fixes pairing devices due a  bug in which ZGP endpoint wasn't configured in some setups which were around for a long time.

The old adapter didn't bail out since the error was silently ignored. Configuring the endpoints never worked prior to this PR. So only the shipping firmware default endpoints where available.

This also fixes sending out ZGP proxy commissioning commands for the setups in which no ZGP endpoint was configured before.

Tested with ConBee 1, 2 and 3.

Issue: https://github.com/Koenkk/zigbee2mqtt/issues/27902


Koenkk:
Thanks for all the quick fixes!

PR #1450, opened 2025-07-12
fix(ignore): add benchmarks with codspeed

Nerivec:
@Koenkk I'll let you setup [codspeed](https://codspeed.io/login?flow=get-started) for this repo, then we can uncomment the ci addition.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,9 @@
 name: CI
 
-on: [pull_request, push]
+on:
+  push:
+    branches: [master]
+  pull_request:
 
 permissions:
     contents: write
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -13,20 +16,33 @@ jobs:
...
               run: pnpm run build
+
             - name: Test
               run: pnpm run test:coverage -- --testTimeout=10000
+
+            # - name: Bench
+            #   uses: CodSpeedHQ/action@v3
+            #   with:
+            #     run: pnpm run bench
+            #     token: ${{ secrets.CODSPEED_TOKEN }}
```
Koenkk: I setup codspeed so should work now

.github/workflows/ci.yml
```diff
@@ -1,6 +1,9 @@
 name: CI
 
-on: [pull_request, push]
+on:
+  push:
+    branches: [master]
```
Koenkk: Why only for master?

Nerivec: Forgot the tags trigger for versions.
I think the new trigger should work for all needed cases?
This should avoid CI triggering twice in some cases (like when you -the owner- create a PR).

PR #1451, opened 2025-07-14
fix: improve perf when cloning cluster entries

Nerivec:


Koenkk:
Nice to see the improvements!

PR #1453, opened 2025-07-14
fix: remove duplicate logic to set reportable properties

Nerivec:
Removes the duplicate call to `attributeKeyValue` which is the costly part of receiving a ZCL message that triggers it.


Koenkk:
Thanks!

PR #1455, opened 2025-07-19
fix!: remove unnecessary throwing

Nerivec:
Throwing on `getAttribute` was used mostly for flow control, not great for perf. Also, most of the time, that was bypassed by the use of `hasAttribute` (now removed), which resulted in unnecessary dupe get logic.

Invalid attributes for endpoint/group read are now ignored instead of failing the whole request. Added some logging for it, and test coverage (wasn't covered before when it was throwing since the change didn't fail tests).

Note: will require checking a couple of places where `getAttribute` is used in ZHC & Z2M (will at least fail typing).

Also:
- cleaned up a lot of frame matching (use `Zcl.Frame.create` instead of writing the whole object by hand).
- updated vitest, for some reason it was quite behind.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
         "@vitest/coverage-v8": "^3.1.1",
         "rimraf": "^6.0.1",
         "typescript": "^5.8.3",
-        "vitest": "^3.1.1"
+        "vitest": "^3.2.4"
     },
     "homepage": "https://github.com/Koenkk/zigbee-herdsman",
     "keywords": [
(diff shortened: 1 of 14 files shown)


Koenkk (approved):

Nerivec:
Corrected a flaw in the benchmarking, so last edit from codspeed report is not representative (previous edit is the right one, albeit with the flaw).

Koenkk:
For future reference, this PR has the following breaking changes:
- `cluster. hasAttribute()` has been removed
- `cluster.getAttribute()` now returns `undefined` instead of throwing an exception when called with a non-existing attribute

PR #1456, opened 2025-07-21
fix: improve perf with ZCL getters, align manuf code behavior

Nerivec:
- [fix] cluster/command lookup possibly not matching between the command's body and the `Zcl.Frame.create` due to mismatching parameters
- [fix] groups not passing options in same way as endpoints, causing mismatch in parameters during lookup/frame creation
- [perf] use cluster/command directly when already looked up in command's body instead of redoing the lookup at frame creation
- cleanup some tests


Nerivec:
Note: bench regression appears to be a codspeed bug. Doesn't make sense and cannot replicate with local run.

bench

Koenkk:
Thanks!

PR #1449, opened 2025-07-06
feat: support custom clusters for groups

Nerivec:
Supersedes #1318

I think this approach should cover most of the issues described in above PR without requiring manually maintenance in ZHC. It uses the endpoint's in/out clusters to properly identify "common" custom clusters based on request.

Still a couple of questions though:
- What about identically named/IDed custom clusters but with different behavior (attrs/cmds)? These will pass "common" validation even though they should not and will trigger undetermined behaviors on the device (bad parsing = ?).
- This is expecting proper ACTIVE_ENDPOINTS/SIMPLE_DESCRIPTOR responses. Devices with bad ZDO data may incorrectly report common custom clusters. _An acceptable limitation I'd say, bad device is bad device._ 😅

CC: @ccutrer

_NOTE: I cannot test any of this with real devices, I don't have any that fit the bill._

PR #1458, opened 2025-07-22
fix: send proper manufacturerCode with group commands

ccutrer:
so that manufacturer specific commands will work

see also https://github.com/Koenkk/zigbee-herdsman/pull/1449#issuecomment-3100052008


src/controller/model/group.ts
```diff
@@ -358,7 +358,7 @@ export class Group extends Entity {
                 Zcl.FrameType.SPECIFIC,
                 optionsWithDefaults.direction,
                 true,
-                optionsWithDefaults.manufacturerCode,
+                cluster.manufacturerCode,
```
ccutrer, line 361: should I maybe do `optionsWithDefaults.manufacturerCode ?? cluster.manufacterCode` instead, so that if something explicitly passes in a manufacturerCode, it doesn't get overwritten?

Nerivec:
I'm reworking manuf codes in https://github.com/Koenkk/zigbee-herdsman/pull/1456

PR #1459, opened 2025-07-24
feat!: autotype ZCL clusters

Nerivec:
This PR introduces the first step in typing ZCL/Foundation related stuff.
The types are automatically derived from the `Clusters` and `Foundation` objects.

Should allow to introduce typing like this:

_I did not set it up in any kind of automation since this stuff rarely changes, we can deal with it manually as needed._

Note: This PR is breaking due to renaming of a few clusters to match convention. Requires fixing names in ZHC for the two fixed cluster names here: https://github.com/Koenkk/zigbee-herdsman-converters/pull/9679


Koenkk:
Thanks!

Nerivec:
@Koenkk fyi, I saw that `unknown2` param is [used in ZHC](https://github.com/Koenkk/zigbee-herdsman-converters/blob/b8cd9aa10bb6c0e8f21d9721ced34c398a4ab6d3/src/lib/philips.ts#L540) and it was clashing here with it being duped. Not sure if this will fix a bug, or possibly introduce one (if ZHC would need to use `unknown3` instead). I don't have the device, can't confirm which is the right one.

Koenkk:
@Nerivec should be `unknown3`, fixed it!

PR #1464, opened 2025-07-26
chore: import vitest instead of relying on globals

Nerivec:
Appears to sometimes cause issues with `node:test`, so, use explicit imports.


Koenkk:
Thanks!

PR #1466, opened 2025-07-30
fix: code cleanup

Nerivec:
- remove unnecessary non-null assertions in `Entity`
- cleanup `endpoint.sendRequest`


Koenkk:
Thanks!

PR #1468, opened 2025-07-31
fix: Allow BuffaloZclDataType for AttributeDefinition.type

Koenkk:
Already allow for [parameter](https://github.com/Koenkk/zigbee-herdsman/blob/164b6402d49895ac3e18f222ffcee6c57553c901/src/zspec/zcl/definition/tstype.ts#L43) and now also for the `AttributeDefinition`. Required for https://github.com/Koenkk/zigbee-herdsman-converters/pull/9711

### DIFF
--- a/src/zspec/zcl/definition/tstype.ts
+++ b/src/zspec/zcl/definition/tstype.ts
@@ -52,7 +52,7 @@ export interface Command {
 
 export interface AttributeDefinition {
     ID: number;
-    type: DataType;
+    type: DataType | BuffaloZclDataType;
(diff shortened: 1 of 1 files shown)


Nerivec:
Need to change `Attribute` too.
It opens up the code paths a little, though I suppose it all ends up in BuffaloZcl read/write. Might want to add a test or two just in case though, make sure calls with both types are covered.

Koenkk:
Closing this PR as this won't work for reading (only writing), when reading the type is included in the ZCL payload.

PR #1473, opened 2025-08-06
fix!: Expose ZclFrame on message event

Koenkk:
Expose the whole frame in just attributes separately, required for e.g. https://github.com/Koenkk/zigbee-herdsman-converters/pull/9711 where the raw frame is needed.


Nerivec:
I don't really like the fact it opens up ZHC (and external converters) to manipulate what should be only ZH/spec-derived data. It's very bug-prone, especially in cases where the frame might be used after by some other process and it comes up with unexpected formats (not sure if/how, but reference issues are very hard to track down/debug).

PR #1475, opened 2025-08-08
fix: Expose rawData on message event

Koenkk:
This is was actually was needed to fix https://github.com/Koenkk/zigbee-herdsman-converters/issues/9670


Nerivec:
This conflicts with https://github.com/Koenkk/zigbee-herdsman/pull/1473

Koenkk:
@Nerivec note that the `ZclFrame.payload` is something different than the `rawData`, the raw data is currently already exposed when ZclFrame parsing fails: https://github.com/Koenkk/zigbee-herdsman/blob/6f0b443e238ee0fa0e20648c6a91303a8b749aa0/src/controller/controller.ts#L995

PR #1480, opened 2025-08-11
fix: ZCL condition cleanup

Nerivec:
- cleanup conditions to make the switch simpler, move all `equal` types under same roof (_also put some air in the validation function, was unreadable_)
- add some missing conditions in `genOta`
- add some missing parameters in `genOta.imageNotify`
- add support for `genOta.queryDeviceSpecificFile` request/response
- improve error messages when failing to parse ZCL payload (include cmd name & parameter name)
- [workaround] ignore `minimumBlockPeriod` if not present in `genOta.imageBlockRequest` since apparently some devices have bad field control values (fixes https://github.com/Koenkk/zigbee2mqtt/issues/28217)


Koenkk:
Thanks!

PR #1485, opened 2025-08-18
chore: switch benchmarking tool

Nerivec:
Here's an example run: https://github.com/Nerivec/zigbee-on-host/pull/56
PR ci run also shows the results: https://github.com/Nerivec/zigbee-on-host/actions/runs/17053545311
Baseline branch ci run shows the "will-compare-to" results: https://github.com/Nerivec/zigbee-on-host/actions/runs/17054583777

We can adjust options (warmup, time, iterations...) according to early results (ci runner variance & co).
We can also customize the vitest reporter if needed (the action just posts that in the comment).
The action is [pretty simple](https://github.com/Nerivec/action-ci-bench/blob/main/src/index.ts) and easily customizable.

Note: it won't run in this PR since there is no baseline to compare to yet.

_Z2M benches will require a bit of rework to fit back to plain vitest._

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,9 @@ jobs:
     ci:
         runs-on: ubuntu-latest
         permissions:
-            id-token: write
+            id-token: write # for npm publish
+            actions: read # for Nerivec/action-ci-bench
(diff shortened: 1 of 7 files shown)


Koenkk:
Thanks!

PR #1488, opened 2025-08-20
added danfossSWStatusCode

AivCho123:
added danfossSWStatusCode


Koenkk:
Thanks for the PR!

Adding manufacturer attributes to cluster.ts is not allowed anymore, instead the [custom cluster API](https://github.com/Koenkk/zigbee-herdsman-converters/blob/c926b318cc2647aebacd540940b969455edd9991/src/devices/sonoff.ts#L151) should be used from the device definition.

AivCho123:
So I need to rewrite all danfoss.ts or just add extend ?
That parameter was found in official documentation

Koenkk:
All has to be added to `danfoss.ts` indeed, you can use the `sonoff.ts` from above as an example on how to use the custom cluster API.

AivCho123:
for clarification: I need to transfer even the existing parameters from cluster.ts from zigbee-herdsman to danfoss.ts ? 
cluster
also existing lines from ToZigbee.ts and fromZigbee.ts ?

Koenkk:
Lets also transfer those indeed

PR #1492, opened 2025-09-01
Fixes for ZNP INCOMING_MSG_EXT handling with huge data byte count

canardos:
Fixes for `ZNP INCOMING_MSG_EXT` handling as [discussed here](https://github.com/Koenkk/zigbee2mqtt/discussions/28395).

As the data retrieval requires repeated messages to the ZNP, it is an async operation, which differs from the prior handling. Please check this aspect, as I am not very familiar with the zh program flow. Specifically, the TI docs state:

> Note that multiple AF_INCOMING_MSG_EXT indications with huge data byte counts may be queued, and each
will be timed-out separately.

It's not entirely clear whether the ZNP waits until all data has been retrieved from one `AF_INCOMING_MSG_EXT` before sending the next, or whether it's only the huge data buffers that are queued and the second, third etc. message is sent immediately.

In the latter case, the code should retrieve all data from one message before beginning retrieval of the next. I believe that is what the current implementation will do, but please confirm as I am not familiar with the queue processing setup in zh.

EDIT:
FYI I see the ci/test is failing, but all tests completing successfully on the client side.


Koenkk:
> FYI I see the ci/test is failing, but all tests completing successfully on the client side.

The tests fail because it's missing coverage, you can run it locally by executing `pnpm test:coverage`

test/adapter/z-stack/adapter.test.ts
```diff
@@ -3381,13 +3431,69 @@ describe("zstack-adapter", () => {
...
+        expect(zclData.endpoint).toStrictEqual(20);
+        expect(zclData.groupID).toStrictEqual(12);
+        expect(zclData.linkquality).toStrictEqual(101);
+        expect(zclData.address).toStrictEqual(0x5678);
+        expect(zclData.groupID).toStrictEqual(12);
+        expect(zclData.data).toStrictEqual(Buffer.from([24, 100, 1, 0, 0, 0, 32, 2]));
+        expect(zclData.header.commandIdentifier).toBe(1);
+    });
+
+    it("Incoming message extended (inter PAN) - 64-bit address", async () => {
```
Koenkk: Since code is almost the same, could you merge this with the 16-bit test case? This should save a lot of code duplication. The same can be done for the huge bytes test cases.

Example on how to do this: https://github.com/Koenkk/zigbee-herdsman/blob/fe9651586c3ad27869e6e8ee735a40fc1fc0cd9a/test/zcl.test.ts#L1589

After that this is ready for merge.

Koenkk:
Thanks!

PR #1501, opened 2025-09-10
fix: RSSI Location cluster support

Nerivec:
_This cluster has no current use, so I renamed everything to fit convention._


Koenkk:
Thanks!

PR #1502, opened 2025-09-11
fix(ignore): update typescript to 5.9.2

Nerivec:
Looks like most of the previous issues have been ironed out... still a few explicit types required, but far less.


Koenkk:
Nice, thanks!

Issue #1498, opened 2025-09-07
`UINT8` doesn't allow `0xff` (255) as value anymore which breaks code


DerDreschner:
With in the latest releases of Z2M, there must have been a change in the code regarding processing of messages with `UINT8`. This leads to the situation, that 255 isn't a valid value anymore, although it is valid, and it's been used multiple times ([see for example](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/toZigbee.ts#L769)).

I realized that when I implemented [this pull request](https://github.com/Koenkk/zigbee-herdsman-converters/pull/9941), I didn't get any value regarding the light settings as well as the `maximumBrightness` attribute, which allows 255 as a valid value as well (it's actually the default value). The `meta` attribute contains the correct byte values in that case, but I'm not sure if that was after some changes in `zigbee-herdsman` or not. I'm not the only person who realized the issue. See https://github.com/Koenkk/zigbee2mqtt/issues/28506 for example.

Is this a bug in `zigbee-herdsman` or do all converters that need to process `0xff` as a valid value have to be changed to the `raw` message type?

Koenkk:
If I understand correctly, this happens when reading the brightness right? (because setting should still work). 

For reference from the spec:
<img width="801" height="205" alt="Image" src="https://github.com/user-attachments/assets/5b82ec02-85e2-495d-b39f-9e9be22597e9" />

CC: @Nerivec

DerDreschner:
That's correct. Setting `0xff` does work correctly. Just reading is broken.

Nerivec:
> When a variable value is required to designate an unknown, invalid, null, or undefined data value and there
is no obvious data value (e.g. zero), that is within the valid range, then the Non-Value in the data type table
(see 2.6.2.2) MAY be used.

See [Table 2-11 (page 98)](https://github.com/user-attachments/files/19044883/07-5123-08-Zigbee-Cluster-Library.pdf) for details. But basically almost all types have FF (or equivalent) as "non-value", it's a valid value, but signals "undefined".

Should the "default"s in specific tables not be considered non-values even if they are the non-values for the types?
I'm not quite sure what is proper spec here, but we may have to introduce a new param that's passed as option to the `BuffaloZcl` to expand capabilities?
```ts
currentLevel: {ID: 0, type: DataType.UINT8, default: 0xff}, // i.e. "if value is default, use it as plain"
// or
[...]

PR #1503, opened 2025-09-15
fix: enhance ZCL specification

Nerivec:
Introduces new typing for ZCL spec, with initial "auto update" using data from CSA ZAP (https://github.com/project-chip/zap)
Should eventually allow much better processing, e.g.:
- proper processing of non-values & default _per attribute_
- allow extending support (e.g. legacy devices missing a command parameter must rely on default to avoid failures)
- allow clamping numeric values as appropriate, but most importantly detect values out of range (which would at least be useful to log for debug purposes)
- allow better error handling (e.g. don't consider failure when not required)
- allow far better frontend validation & UX (prevent "doomed-to-fail" & co)

Notes:
- using incremental _per commit_ updates of Clusters for easier checking
- using a couple of "reports" for auto updates of Clusters for easier checking (present in same commit)

TODO:

- [x] update cluster typing
- [x] update clusters with ZAP data (auto)
- [ ] check clusters from auto update
- [ ] update clusters with spec data (manual, _for clusters not available with auto update_)
- [ ] create logic to handle default/invalid (will handle https://github.com/Koenkk/zigbee-herdsman/issues/1498)
- [ ] [optional/future] create logic to handle other metas
- [ ] [optional/future] add touchlink-specific types to parse from Buffalo
- [ ] move 3 sprut clusters to custom clusters (uses IDs not intended for custom)
- [ ] found spec typos (error-prone): `installedClosedLimitTiltDdegree`, `acCollTemp` (needs cross-repo check)
- [ ] remove `clusters-old.ts` reference (used by scripts for reporting)
- [ ] remove no longer useful scripts

CC: @sjorge @kirovilya


sjorge:
Big change, I had a quick look over it so far, I love the switch to constantly using hex makes things so much easier when reading the ZCL.

So looking at the scripts, is there an XML source of all the known clusters/attribute and they are now being parsed?

Nerivec:
Yes, it's in the ZAP repo I linked ([zap/zcl-builtin/dotdot](https://github.com/project-chip/zap/tree/master/zcl-builtin/dotdot)). Though it does not have _all_ spec clusters for some reason, so some will have to be done manually.

_ZAP could technically be used to generate the whole ZCL Clusters object for ZH, that's one of its purposes, but the problem is backwards compat with name changes (& customs), that would create a massive amount of changes needed in ZHC (& external converters). So, had to build a script to merge data instead._

Nerivec:
1213 pages later... 😅 

Important stuff to check:
- [changelog](https://github.com/Nerivec/zigbee-herdsman/blob/zap/scripts/clusters-changes.log) of existing data changes in `scripts/clusters-changes.log` (e.g. datatype changes, attribute additions, etc.)
  - see notes in first post (re. importance of some datatype changes)
  - _as a (much) shorter alternative to re-checking the whole spec, these highlight changes to prior behavior only (+ additions)_
- main logic changes in `zspec/zcl/utils.ts` (with specific tests added in `test/zspec/zcl/utils.test.ts`) & `zspec/zcl/buffaloZcl.ts` (mostly cleanup)
- `zspec/zcl/definition/datatypes.ts` is auto-generated from ZAP (we can deprecated some of the manually defined stuff eventually)
- keeping prior behavior:
  - types for Foundation are not defined specifically to uint, so moved everything to data (no non-values) - _more of a judgment guess really, didn't find an explicit reference_
  - custom parameters with non-values were marked with min/max to skip non-values (can be cleaned up later as needed)
    - e.g. `type: DataType.UINT8, max: 0xff`
  - custom attributes were all marked `writable` (except when comments explicitly specified otherwise) and with min/max to skip non-values (can be cleaned up later as needed)
    - e.g. `type: DataType.UINT8, writable: true, max: 0xff`
    - e.g. `type: DataType.INT8, writable: true, min: -128`

_Note: I skipped a few of the massive clusters (mostly SE-related) that are not currently implemented and have no known use (added comments about them though)._

Nerivec:
Fixed an issue with new options fields added to most control cmds for several clusters. Many devices don't implement this (yet), so we don't want to fail if not (assumed `0` if not present). These cmds should not be _received by coordinator_, so should never be a problem, but... :sweat_smile:

We should however update ZHC to send these two attributes (even if it's `0` for now), as it does fail payloads for e.g. in Wireshark if not present (doesn't appear to check/default to zero, so, failure).

scripts/clusters-changes.log
```diff
@@ -0,0 +1,585 @@
+Found 582 changes.
+
+[REMOVED] Clusters > pm1Measurement: 1068 (0x042c)
+[REMOVED] Clusters > pm10Measurement: 1069 (0x042d)
```
Koenkk, line 4: Do this one and the one above need to be added as custom clusters in zhc?

scripts/clusters-changes.log
```diff
@@ -0,0 +1,585 @@
...
+[CHANGED] Clusters > genCommissioning > commandsResponse > resetStartupParamsRsp > parameters > status > type: UINT8 -> ENUM8
+[CHANGED] Clusters > genOta > commands > queryNextImageRequest > parameters > fieldControl > type: UINT8 -> BITMAP8
+[CHANGED] Clusters > genOta > commands > imageBlockRequest > parameters > fieldControl > type: UINT8 -> BITMAP8
+[CHANGED] Clusters > genOta > commands > imagePageRequest > parameters > fieldControl > type: UINT8 -> BITMAP8
+[CHANGED] Clusters > genOta > commands > upgradeEndRequest > parameters > status > type: UINT8 -> ENUM8
+[CHANGED] Clusters > genOta > commandsResponse > imageNotify > parameters > payloadType > type: UINT8 -> ENUM8
+[CHANGED] Clusters > genOta > commandsResponse > queryNextImageResponse > parameters > status > type: UINT8 -> ENUM8
+[CHANGED] Clusters > genOta > commandsResponse > imageBlockResponse > parameters > status > type: UINT8 -> ENUM8
+[CHANGED] Clusters > genOta > commandsResponse > upgradeEndResponse > parameters > currentTime > type: UINT32 -> UTC
+[CHANGED] Clusters > genOta > commandsResponse > upgradeEndResponse > parameters > upgradeTime > type: UINT32 -> UTC
```
Koenkk, line 68: Is this breaking? Maybe UTC can be a native date object?

src/controller/model/endpoint.ts
```diff
@@ -429,7 +431,11 @@ export class Endpoint extends ZigbeeEntity {
             const attribute = cluster.getAttribute(nameOrID);
 
             if (attribute) {
-                payload.push({attrId: attribute.ID, attrData: attributes[nameOrID], dataType: attribute.type});
+                // TODO: refs are not processed (pre-loop?)
```
Koenkk: What does this mean?

src/zspec/zcl/buffaloZcl.ts
```diff
@@ -35,197 +39,34 @@ const EXTENSION_FIELD_SETS_DATA_TYPE: {[key: number]: DataType[]} = {
...
-
-    private readZclUInt8(): number {
+    private readLengthUInt8(): number {
         const value = this.readUInt8();
 
-        return value;
-
-        // See UINT8_TMP_FIX
-        // return value === 0xff ? Number.NaN : value;
+        return value === UINT8_NON_VALUE ? Number.NaN : value;
```
Koenkk, line 46: Will this not result in incorrect NaN values again?

scripts/clusters-changes.log
Nerivec: It's just the proper alias. All uint32 behind.
https://github.com/Koenkk/zigbee-herdsman/blob/6384ef24d23d777a9e08dbdd7878dbeaee5e918c/src/zspec/zcl/buffaloZcl.ts#L1065-L1071

src/controller/model/endpoint.ts
Nerivec: Refs would technically need to be run through the checks before being passed (otherwise they might still contain e.g. out of bounds values).

I'm wondering if we should just drop refs altogether. It's not a really big gain, most cases are around ms clusters, and they are not useful to us (intended for firmware config, not writable).
Like this:
```ts
            measuredValue: {
                ID: 0x0000,
                type: DataType.UINT16,
                reportRequired: true,
                required: true,
                minRef: "minMeasuredValue",
                maxRef: "maxMeasuredValue",
            },
            minMeasuredValue: {ID: 0x0001, type: DataType.UINT16, required: true, max: 65533, maxExclRef: "maxMeasuredValue"},
            maxMeasuredValue: {ID: 0x0002, type: DataType.UINT16, required: true, min: 1, max: 65534, minExclRef: "minMeasuredValue"},
```

For the few that are writable, I think ZHC can handle that as needed by devices?
Makes the logic cleaner overall to drop this.

src/zspec/zcl/buffaloZcl.ts
Nerivec: This is for the `length` field of various attrs/params. This one should be enforced.

scripts/clusters-changes.log
Nerivec: Have to check.
I had mentioned this in a previous cleanup, these don't exist in the Zigbee spec, but we hadn't removed them yet.

EDIT: found no use in ZHC neither by name/ID/hex ID

src/controller/model/endpoint.ts
Nerivec: I've removed the logic:
- in most cases, the refs are meant for ZAP use in device firmware config (i.e. not for our use case)
- in most cases, the payload (attributes) needed for the refs logic to actually be used won't be complete or there at all (i.e. unused)

_If needed we can revisit this in a future PR, needs a bigger refactor of the stack logic to have ref values always present when needed._

Koenkk: I agree, let's keep it simple especially if it's not used.

Nerivec:
@Koenkk once good to go, the following script should take care of all ZHC custom clusters (will set behavior as "backwards compatible"):
https://gist.github.com/Nerivec/b3f514201f41eb8218f8dd3fb78217fd
They can be further edited in future PRs according to their respective specs by users with the knowledge to do so :wink:

Nerivec:
Took this out of draft, though I suggest we wait for next release, since we're now closer to it than not :grin: 
Let me know if you see anything else needed before merging.

Koenkk:
@Nerivec perfect!

Nerivec:
Cleaned up a bit, figured extra bytes saved with shorter property names in `Clusters` was worth it since this is published over MQTT.
Also updated the script linked above to match (just `writable`=>`write`).
Removed `genLevelCtrlForLightning` in favor of always using `genLevelCtrl` (to have just 1 cluster with same ID). It's a bit less restrictive than `genLevelCtrlForLightning` in metadata, but ZHC can do the rest as needed (with so many off-spec devices, probably best to use the base version instead of the derived one anyway). _Added comments in-place for stuff that would be different._ Should avoid making a mess with cache, scenes & co.

I left the added scripts, even though most can no longer be used as-is (was meant for first-trigger), they might be useful in future...

burmistrzak:
@Nerivec Looking sweet! 🙌

Figuring out the correct metadata for custom clusters will be... fun. 😅
I wouldn't be surprised if some devices don't handle out-of-bound values correctly.

Nerivec:
@burmistrzak in most cases, it will be around `write`, which can easily be tested with the equivalent command (to see if it fails). For the min/max/default, that's a bit more complicated, can try to write the type's non-value, then read the attr to see what was actually saved, same for bounds, but less likely to provide an actionable answer (just "it works at X"), unless the device ends up actually writing it as non-value (which would point to "it doesn't work at X" :grin:).

It's mostly an enhancement though, no doubt plenty of custom clusters will keep the default of `write: true, min|max: {of-type}` A.K.A. "whatever" :sweat_smile: 

It should be nice for the dev console in frontend, keep users from wondering why a read/write didn't work, at least with spec clusters. And should also allow to detect improper converters that may have been silently ignoring errors in the past, which may have resulted in incorrect assumptions.

burmistrzak:
@Nerivec Regarding custom cluster enhancements in the frontend, would it be possible to put them first (or prefix them w/ the custom cluster's name) in the attribute dropdown?

Koenkk:
Thanks!

sjorge:
A bit late but congrats on landing this! <3

PR #1507, opened 2025-09-20
fix: improve ZCL TSN handling

### LINKED ISSUE #1506: Setting a zero TSN in a commandResponse fails
Hi,

I know I should do a pull request for this but I'm not really a javaScript programmer, sorry.  I could have a go but I'd probably get it wrong and you'd alter and fix it anyway :smile:   It's also a simple one for you to fix please.

I'm using custom commands (using C++ and ESP-IDF on an ESP32-C6) with zigbee2mqtt to pass serial data back and forth.  The commandResponse in this file needs updating please:

`controller/model/endpoint.ts`

On or around line 831 is this
```
        transactionSequenceNumber = transactionSequenceNumber || zclTransactionSequenceNumber.next();
```

The issue I'm having is that when I provide a valid TSN of 0 it skips the || and uses the next() tsn sequence number and does not use my provided value of zero.  All other values 1 to 255 work as expected. It is taking 0 as boolean false and skipping it when it should use 0 as a valid value.

Can I suggest that the line / code is updated to allow 0 to 255 as valid tsn, and only use next() when null or undefined.
Thanks

Nerivec:
- fix couple places where zero wasn't allowed due to `||` check instead of `??`
- fix some adapters not properly matching against TSN zero
- range of ZCL TSN helper class always starts at zero and wraps after 255

Fixes #1506 
Might need some test tweaks in Z2M.

_I wonder if biome has a rule to prevent `if (!num)` or `num || ...` when `typeof num === "number"`. Would be useful as a warning at least._

### DIFF
--- a/src/adapter/deconz/adapter/deconzAdapter.ts
+++ b/src/adapter/deconz/adapter/deconzAdapter.ts
@@ -874,7 +874,7 @@ export class DeconzAdapter extends Adapter {
             payload.header &&
                 (!matcher.address || payload.address === matcher.address) &&
                 payload.endpoint === matcher.endpoint &&
-                (!matcher.transactionSequenceNumber || payload.header.transactionSequenceNumber === matcher.transactionSequenceNumber) &&
+                (matcher.transactionSequenceNumber === undefined || payload.header.transactionSequenceNumber === matcher.transactionSequenceNumber) &&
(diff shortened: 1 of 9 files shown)


Koenkk:
thanks

PR #1510, opened 2025-09-22
fix: Fix uint8 `NaN` value https://github.com/Koenkk/zigbee-herdsman/issues/1498

Koenkk:
A temp fix until https://github.com/Koenkk/zigbee-herdsman/pull/1503 is ready.

Fixes:
- https://github.com/Koenkk/zigbee-herdsman/issues/1498
- https://github.com/Koenkk/zigbee-herdsman-converters/pull/10065

### DIFF
--- a/src/zspec/zcl/buffaloZcl.ts
+++ b/src/zspec/zcl/buffaloZcl.ts
@@ -43,7 +43,12 @@ export class BuffaloZcl extends Buffalo {
     private readZclUInt8(): number {
         const value = this.readUInt8();
 
-        return value === 0xff ? Number.NaN : value;
+        return value;
+
+        // TEMP: return value instead of Number.NaN.
+        // Will be replaced by https://github.com/Koenkk/zigbee-herdsman/pull/1503
+        // https://github.com/Koenkk/zigbee-herdsman/issues/1498
+        // return value === 0xff ? Number.NaN : value;
     }
 
     private writeZclUInt16(value: number): void {


src/zspec/zcl/buffaloZcl.ts
```diff
@@ -200,14 +210,20 @@ export class BuffaloZcl extends Buffalo {
         if (value) {
             this.writeUInt8(value.length);
             this.writeBuffer(value, value.length);
-        } else {
```
Nerivec, line 203: Should not change that one.

src/zspec/zcl/buffaloZcl.ts
```diff
@@ -200,14 +210,20 @@ export class BuffaloZcl extends Buffalo {
...
+        //     this.writeUInt8(0xff); // non-value
+        // }
     }
 
     private readOctetStr(): Buffer {
         const length = this.readZclUInt8();
-        return Number.isNaN(length) ? Buffer.from([]) : this.readBuffer(length);
+
+        // See UINT8_TMP_FIX
+        return this.readBuffer(length);
```
Nerivec: Have to keep the check here, but without NaN use:
```ts
        return length < 0xff ? this.readBuffer(length) : Buffer.from([]); // non-value
```
same for char_str, and long versions (with 0xffff).

https://github.com/Koenkk/zigbee-herdsman/pull/1424/files#diff-fc7e6b2e70437c37018dafdc034cd36e90e974b6745faf828ed43cd34ab4cc1fL142

src/zspec/zcl/buffaloZcl.ts
Koenkk: Done, had to add no coverage marks

src/zspec/zcl/buffaloZcl.ts
Koenkk: Updated `readCharStr` and `readOctetStr`. I don't think `readLongOctetStr` has to be updated because it's reading a uint16? (which will still return NaN)

PR #1511, opened 2025-09-28
fix: Fix writing empty (long) char str

Koenkk:
This PR fixes the writing of empty (long) char strings. Currently in case of an empty string (`""`) the non-value is written instead of a zero length string. This is incorrect, instead the length should be set to zero, from the spec:

Screenshot 2025-09-28 at 22 14 45

This causes e.g. scene add command with an empty group name to fail.

Fixes:
- https://github.com/Koenkk/zigbee2mqtt/issues/28475
- https://github.com/Koenkk/zigbee2mqtt/issues/28532
- And maybe also https://github.com/Nerivec/zigbee2mqtt-windfront/issues/232 ?

### DIFF
--- a/src/zspec/zcl/buffaloZcl.ts
+++ b/src/zspec/zcl/buffaloZcl.ts
@@ -227,7 +227,10 @@ export class BuffaloZcl extends Buffalo {
     }
 
     private writeCharStr(value?: string | number[]): void {
-        if (value) {
+        // In case of an empty string, send 0 length, from the spec:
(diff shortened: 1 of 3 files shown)


Nerivec:
Another case of "too-wide" `if` :sweat_smile: 
Looks like the same thing applies to OCT_STR & LONG.

Koenkk:
> Looks like the same thing applies to OCT_STR & LONG.

In case of an empty string, since it's a `number[]`, a `[]` should be given which does not evaluate to `false`, so the current changes should be enough (?) (if not, I'll submit a second PR)

PR #1516, opened 2025-09-30
Add tuyaWeatherSync and tuyaWeatherRequest commands to manuSpecificTuya cluster

l1berat0r:
Hi,

I have added tuyaWeatherSync and tuyaWeatherRequest commands to manuSpecificTuya cluster.
Main motivation was to support weather information on Tuya M8Pro device. Although I tried the code to be as universal as possible it has only been tested with the mentioned device which has some quirks in how it implements the protocol.
More on that [here](https://github.com/Koenkk/zigbee-herdsman/commit/61b7d4613264b096c6c61384ea4ec4495d15a5de#diff-b186f965ae516aa02794eaf5db244e40d3a6ee0ab33c7251590f7eafc8f4e3f6R5236)

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -5118,6 +5118,14 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
                     {name: "payload", type: DataType.UINT8},
                 ],
             },
+
+            /**
+             * Weather forecast synchronization (check requestWeatherInformation)
+             */
+            tuyaWeatherSync: {
+                ID: 0x61,
+                parameters: [{name: "payload", type: BuffaloZclDataType.BUFFER}],
+            },
         },
         commandsResponse: {
             /**
@@ -5220,6 +5228,18 @@ export const Clusters: Readonly<Record<ClusterName, Readonly<ClusterDefinition>>
             mcuGatewayConnectionStatus: {
                 ID: 0x25,
                 parameters: [{name: "payloadSize", type: DataType.UINT16}],
+            },
+
+            /**
+             * Device can request weather forecast information and expects response respecting given parameters.
+             * This command ID seem to be device speciffic, because there is simmilar structure documented in Tuya Serial Communication Protocol,
+             * but with different ID (0x3a and 0x3b respectively). In this case, I'm not sure if the name should reflect the one from
+             * docs or be also speciffic (providing space for the implementation of the correct one in the future)?
+             *
+             */
+            tuyaWeatherRequest: {
+                ID: 0x60,
+                parameters: [{name: "payload", type: BuffaloZclDataType.BUFFER}],
             },
         },
     },
--- a/src/zspec/zcl/definition/clusters-types.ts
+++ b/src/zspec/zcl/definition/clusters-types.ts
@@ -6194,6 +6194,11 @@ export interface TClusters {
                 /** Type: UINT8 */
                 payload: number;
             };
+            /** ID: 96 */
+            tuyaWeatherRequest: {
+                /** Type: BUFFER */
+                payload: Buffer;
+            };
         };
         commandResponses: {
             /** ID: 1 */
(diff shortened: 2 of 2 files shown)


Nerivec:
Just gave a quick look, but it looks like something that should be in converters, not in ZH?

markajm:
I've just started playing with an M9Pro.  Seeing the same 0x60, 0x61 commands, so hopefully this will work for me too :)

l1berat0r:
> Just gave a quick look, but it looks like something that should be in converters, not in ZH?

@Nerivec - this device uses command ID that has not yet been implemented in ZH, for weather information sync. It has simmilar structure as the one documented in Tuya Serial Sommunication Protocol standard however has different command id for some reason.
There is also a PR for a [converter](https://github.com/Koenkk/zigbee-herdsman-converters/pull/10119)

@markajm I suppose it should. It seems it's the same manufacturer. Please let me know if you make any progress. It's also in my drawer waiting to ba taken care of :)

Nerivec:
The command definitions would go in `cluster.ts` (although it is a custom cluster that should be moved to ZHC eventually too), but the rest is entirely custom, so that would be in converters (ZH would just use BUFFER for types and ZHC would take care of the rest). _ZH should remain about Zigbee spec (apart from a few "impossible" quirks)._

Koenkk will confirm though.

l1berat0r:
I tried to follow what I already seen in the code. 
From what I could find, there were already less complex protocols put to [buffaloZcl.ts](https://github.com/Koenkk/zigbee-herdsman/blob/v6.1.5/src/zspec/zcl/buffaloZcl.ts).
To point a few:
- https://github.com/Koenkk/zigbee-herdsman/blob/v6.1.5/src/zspec/zcl/buffaloZcl.ts#L445
- https://github.com/Koenkk/zigbee-herdsman/blob/v6.1.5/src/zspec/zcl/buffaloZcl.ts#L750
- https://github.com/Koenkk/zigbee-herdsman/blob/v6.1.5/src/zspec/zcl/buffaloZcl.ts#L785

I couldn't define the weather protocol, using just types from DataType and BuffaloZclDataType. I contains variable length fields terminated by special characters. This is why I thought it would be best to add distinct data type and the parser functions.
Would that qualify as an "impossible" quirk, or is there another, correct way to address parsing of this kind of data?

Nerivec:
The Mi stuff is pretty old, was before we started moving non-spec stuff outside ZH, like the remaining custom clusters that need to be moved to ZHC. The themo transition is Zigbee spec.

BUFFER should be used for this case, and the logic handled in ZHC. E.g.:
```ts
{ name: "payload", type: BuffaloZclDataType.BUFFER }
```
This is already done by many other converters for custom logic. ZHC builds the custom payload, ZH receives the final `Buffer` payload. The only change needed in ZH is adding the two commands to `cluster.ts` (and regenerating the types with the script).

Koenkk:
I agree with @Nerivec, the changes to `cluster.ts` are fine but the parsing can move to zigbee-herdsman-converters. In order to get the raw buffer there replace: `{name: "payload", type: BuffaloZclDataType.TUYA_WEATHER_REQUEST}` with `{name: "payload", type: BuffaloZclDataType.BUFFER}`

You are right that some of these protocol are in zigbee-herdman (like the Mi stuff) but it's really old (6+ years or something). Nowadays we want as little as possible manufacturer specific stuff in zigbee-herdsman.

l1berat0r:
Ok, understood. I reverted the protocol changes and left only new commands deffinitions.
I'll work on moving the logic to ZHC and hopefully push new commit later today or tomorrow.

Koenkk:
Thanks!

markajm:
> 
> @markajm I suppose it should. It seems it's the same manufacturer. Please let me know if you make any progress. It's also in my drawer waiting to ba taken care of :)

Hi @l1berat0r , I've put together a basic converter for the M9 (https://github.com/Koenkk/zigbee-herdsman-converters/issues/10161).  All the basic functions working.  Haven't tested your new weather code yet.  I'm new to zigbee and haven't got my head around much of it yet :)

Issue #122, opened 2020-01-07
How to respond to `timeStatus` requests?


markuspf:
I have a device (https://github.com/Koenkk/zigbee2mqtt/issues/2648) that I am trying to hack up support for.

I asks for `timeStatus" in the `genTime` cluster. There is already a response to `time`, and I tried hacking in a response to `timeStatus` by extending the code for `time`.

I don't seem to be doing it right, as it doesn't seem the device is satisfied with the answer (and I can't really verify whether the answer is sent out correctly). I'd be grateful for any pointers how to do this correctly.

Koenkk:
You can respond to it here: https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/controller.ts#L509 and very if using sniffing https://www.zigbee2mqtt.io/how_tos/how_to_sniff_zigbee_traffic.html

markuspf:
Ok, maybe lets first find out whether I am doing the code bit correctly, as that place is exactly where I was trying to respond before; 

I added:

```
if(data.includes("timeStatus")) {
  debug.log("responding to timeStatus: ");
  const resps = 5; // master, master for time zone and DST
  try { 
    yield endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, {"timeStatus": { value: resps, type: 24 }} );
   }
  catch (error) {
    debug.log(error); 
    debug.error(`genTime response to ${device.ieeeAddr} failed`);
  }
}
```

In particular I am not sure what the structure of the response object passed to `readResponse` is supposed to be.

Koenkk:
It should be 

```js
    yield endpoint.readResponse(frame.Cluster.ID, frame.Header.transactionSequenceNumber, {timeStatus: resps} );
```

sjorge:
I think this is supported now?

Koenkk:
Yes

PR #1517, opened 2025-10-01
feat: Add support for daylight saving time in `genTime` cluster

DerDreschner:
Currently, there is no support for daylight saving time in zigbee-herdsman. Some devices ask the coordinator repeatedly this information if there is no answer or just the default value as answer ([looking at you, Bosch!](https://github.com/Koenkk/zigbee-herdsman-converters/pull/5599)). To avoid that, there was the need for an `customReadResponse` function, which isn't that great.

This pull request changes that and includes support for daylight saving time. As the whole topic can be pretty complex, I've settled on the basics: daylight saving time is well-established in the timezone. This is important, as countries can leave or introduce daylight saving time ([happened a lot over the past 10-20 years](https://www.bbc.com/future/article/20240308-daylight-savings-reasons-history-health-hazards-countries-that-abolished)). In that case, I leave the default values, which should match the current behavior.

This is a follow-up to #1515.

### DIFF
--- a/src/utils/timeService.ts
+++ b/src/utils/timeService.ts
@@ -77,7 +77,7 @@ function recalculateTimeData(currentDate: Date) {
                     start: new Date(currentYear + 1, 1),
                     end: new Date(currentYear + 1, 12),
                 });
-                const hasRegularDstNextYear = dstChangesThisYear.length === 2;
+                const hasRegularDstNextYear = dstChangesNextYear.length === 2;
 
                 if (hasRegularDstNextYear) {
                     timeZoneDifferenceToUtc = dstChangesThisYear[0].offset * 60;


src/controller/model/device.ts
```diff
@@ -16,11 +17,6 @@ import type {DatabaseEntry, DeviceType, KeyValue} from "../tstype";
 import Endpoint from "./endpoint";
 import Entity from "./entity";
 
-/**
- * @ignore
- */
-const OneJanuary2000 = new Date("January 01, 2000 00:00:00 UTC+00:00").getTime();
```
Nerivec, line 22: Should not remove this pre-computed value since it's still used.

src/controller/model/device.ts
```diff
@@ -348,19 +344,99 @@ export class Device extends Entity<ControllerEventMap> {
             await endpoint.command("ssIasZone", "enrollRsp", payload, {disableDefaultResponse: true});
         }
 
-        // Reponse to read requests
+        // Response to read requests
         if (frame.header.isGlobal && frame.isCommand("read") && !this._customReadResponse?.(frame, endpoint)) {
-            const time = Math.round((Date.now() - OneJanuary2000) / 1000);
+            const currentTime = Date.now();
```
Nerivec: This is too much logic to run with every read frame I think.
We need to find a better way to do this _only if it's actually needed_.

Nerivec:
For ref, here's a code snippet of the handling from Silabs:
https://github.com/SiliconLabs/simplicity_sdk/blob/sisdk-2025.6/protocol/zigbee/app/framework/plugin/time-server/time-server.c

src/controller/model/device.ts
DerDreschner: Hmm, do you know where? I looked for other usages aside the one I do as well and didn't found anything. That's why I moved it into the function that uses it. Or is it about preventing a possible re-compute every time I use it?

src/controller/model/device.ts
DerDreschner: Gotcha, I'm working on it right now.

src/controller/model/device.ts
Nerivec: > aside the one I do as well 

Yes that's it.

> Or is it about preventing a possible re-compute every time I use it?

Yes. It's a static value, might as well not re-compute it a half dozen times every time the code block executes.

src/controller/model/device.ts
DerDreschner: Moved everything into a singleton module. Hope that meets the requirements now?

src/controller/model/device.ts
DerDreschner: Done!

src/controller/model/device.ts
```diff
@@ -348,23 +344,19 @@ export class Device extends Entity<ControllerEventMap> {
...
-                        lastSetTime: time,
-                        validUntilTime: time + 24 * 60 * 60, // valid for 24 hours
-                    },
-                },
             };
 
+            const isTimeReadRequest = dataPayload.clusterID === Zcl.Clusters.genTime.ID;
+            if (isTimeReadRequest) {
+                attributes.genTime = {
+                    attributes: timeService.getTimeClusterAttributes(),
```
Koenkk, line 356: It's not done in the original implementation, but since we are adding more attributes now, we shouldn't respond with all but only the requested ones.

DerDreschner: Isn't that the case already? https://github.com/Koenkk/zigbee-herdsman/blob/master/src/controller/model/device.ts#L371

Koenkk: Ah yes, my apologies!

DerDreschner: No problem :smile: I've send non-requested data during development on the `customReadResponse` change in the Bosch lib once. The device wasn't happy about that, that's why I checked that pretty strict here. :sweat_smile:

src/utils/timeService.ts
```diff
@@ -0,0 +1,165 @@
...
+            const dstStartDelay = currentTime - dstStart;
+            setTimeout(() => {
+                dstActive = true;
+            }, dstStartDelay);
+        }
+
+        if (currentTime < dstEnd && currentTime > dstStart) {
+            dstActive = true;
+            const dstEndDelay = currentTime - dstEnd;
+            setTimeout(() => {
```
Koenkk: I don't really understand what the setTimeout does, it seems to rely on the fact that `getTimeClusterAttributes` is executed between  `dstActive = true;` and ` dstActive = false;`?

test/controller.test.ts
```diff
@@ -2873,50 +2875,136 @@ describe("Controller", () => {
...
-            address: 129,
-            clusterID: frame.cluster.ID,
-            data: frame.toBuffer(),
-            header: frame.header,
-            endpoint: 1,
-            linkquality: 19,
-            groupID: 10,
-        });
+    it.each([
+        {
```
Koenkk: Since this test has only 1 parameter, I think it makes more sense to put the values in the test.

DerDreschner: That's true. Done! :+1:

src/utils/timeService.ts
DerDreschner: I've changed this part a bit to make it more clear (and avoid multiple timers tripping for the same event).

In this case, I use `setTimeout()` to ensure the returned `localTime` is always accurate when the daylight saving starts or ends (it's used [here](https://github.com/Koenkk/zigbee-herdsman/pull/1517/files#diff-d9d18280867ec10c89b5146531eec910729edb53a836c4a89b6250e89322fb5dR58) ). I didn't want to refresh the whole cached information in that case for performance reasons. I now only set this timer when the start or end moment is before the next update.

Without the timer, [this unit test fails](https://github.com/Koenkk/zigbee-herdsman/pull/1517/files#diff-ff16bdd20c13abe2046371fc793c2ce577c4eb809505eeea82691440a37e70dbR154).

Nerivec: I don't think we should be using timers. Better to not cache.

DerDreschner: Alright, I just tried to prevent as many calculations in `getTimeClusterAttributes()` as possible for performance reasons.

I'll remove the timer and put the logic into the `getTimeClusterAttributes()` function.

Nerivec: It would create dangling timers. That's a big source of problems.
We can add a couple benchmarks to check how well the "whole thing" is performing once the PR is nearly done, and see if there are any bottlenecks to fix.

DerDreschner: Done. I've included a comment on why it has to be ">=" for dstStart and "<" for dstEnd as it confused myself during development.

test/controller.test.ts
```diff
@@ -2894,20 +2929,68 @@ describe("Controller", () => {
...
-        // attrId 9 is not supported by controller.ts therefore should not be in the response
-        expect(message.payload.length).toBe(3);
-        expect(message.payload[0].attrId).toBe(0);
-        expect(message.payload[0].dataType).toBe(226);
-        expect(message.payload[0].status).toBe(0);
-        expect(message.payload[0].attrData).toBeGreaterThan(600822353);
-        expect(message.payload[1].attrId).toBe(1);
-        expect(message.payload[1].dataType).toBe(24);
-        expect(message.payload[1].status).toBe(0);
+        expect(message.payload.length).toBe(10);
```
Nerivec: Should use `toStrictEqual` (`toBe` has some limitations that I can never remember, so I avoid it, even when asserting on the type isn't strictly necessary).

src/utils/timeService.ts
```diff
@@ -0,0 +1,145 @@
...
+        standardTime: standardTime,
+        localTime: localTime,
+        lastSetTime: cachedTimeData.lastSetTime,
+        validUntilTime: cachedTimeData.validUntilTime,
+    };
+}
+
+function recalculateTimeData() {
+    const currentTime = Date.now();
+    const currentDate = new Date(currentTime);
```
Nerivec, line 76:
```ts
const currentDate = new Date();
const currentTime = currentDate.getTime();
```

test/controller.test.ts
DerDreschner: Done.

src/utils/timeService.ts
```diff
@@ -0,0 +1,143 @@
...
+    }
+
+    return Math.round((timestamp - OneJanuary2000) / 1000);
+}
+
+export function clearCachedTimeData() {
+    cachedTimeData = <CachedTimeData>{};
+}
+
+export function getTimeClusterAttributes(): TimeClusterAttributes {
```
Nerivec: Return type should be `TClusterAttributes<"genTime">` (or wrapped with `Partial<..>` if not all needed).

Nerivec:
I pushed a couple simple benchmarks in master so we have a ref on the overall impact of this PR.
I rebased your branch to let the benchmarks run here too (make sure to pull for changes locally before pushing new changes). You can track the results in every CI run, e.g. https://github.com/Koenkk/zigbee-herdsman/actions/runs/18234767632?pr=1517

Looks like it doesn't have any noticeable impact :+1: 
`tzScan` appears to be well optimized.

I've also shuffled the order of functions, just to be a bit clearer (declared before called), and fixed the cache to avoid re-creating the object (since 0 can be used for detection, might as well).

src/utils/timeService.ts
```diff
@@ -0,0 +1,131 @@
...
+    return {
+        time: currentTime,
+        timeStatus: 3,
+        timeZone: cachedTimeData.timeZoneDifferenceToUtc,
+        dstStart: cachedTimeData.dstStart,
+        dstEnd: cachedTimeData.dstEnd,
+        dstShift: cachedTimeData.dstShift,
+        standardTime: standardTime,
+        localTime: localTime,
+        lastSetTime: cachedTimeData.lastSetTime,
```
Nerivec: Not sure about this being pulled from the cache? Is that what the spec says? Otherwise, I'd be inclined to set it to `currentTime`.

DerDreschner: In our case, it should technically contain the timestamp of the last time synchronization with external sources over NTP/PTP/NTS. Since we don't know that moment, I abuse it as "last time the cached data were updated". This isn't technically correct, but it's okay to do that in my opinion.

image

What's your opinion on that?

Nerivec: > most recent time that the `Time` attribute was set

So I'd go with always setting it to `currentTime`. Can remove `lastSetTime` from the cached data completely I'd say.

src/utils/timeService.ts
```diff
@@ -0,0 +1,131 @@
...
+    cachedTimeData.dstShift = dstShift;
+    cachedTimeData.lastSetTime = timestampToZigbeeUtcTime(currentTime);
+    cachedTimeData.validUntilTime = timestampToZigbeeUtcTime(validUntilTime);
+}
+
+export function getTimeClusterAttributes(): TClusterAttributes<"genTime"> {
+    const currentTime = timestampToZigbeeUtcTime(Date.now());
+
+    if (currentTime >= cachedTimeData.validUntilTime) {
+        recalculateTimeData();
```
Nerivec: There's a little bit of a discrepancy here, as we're using different `currentTime` between this function and `recalculateTimeData()`, we should pull the `currentDate`/`currentTime` out of recalc and pass them to it, so it matches properly.
_It's an edge-case really, as this should happen fast enough, but might as well..._

DerDreschner: Yeah, that's true.  I've changed that. Wasn't able to write a unit test for that edge case sadly, but I doubt it's even possible as the time might be static when using `setSystemTime`.

src/utils/timeService.ts
DerDreschner: Yeah, true. It's more to spec to assume "calling `new Date()` = time attribute was internally set". Changed that.

src/utils/timeService.ts
```diff
@@ -0,0 +1,133 @@
...
+            dstStart = dstChangesThisYear[0].date.getTime();
+            // tzScan returns the first second on which a new shift has to be applied.
+            // This is fine for dstStart, but dstEnd has to be the last second of the
+            // DST period. Otherwise, the following equation copied from the ZCL spec
+            // isn't met anymore:
+            //
+            // Local Time = Standard Time + DstShift (if DstStart <= Time <= DstEnd)
+            //
+            // Therefore, we have to remove one second on all dstEnd times below.
+            dstEnd = dstChangesThisYear[1].date.getTime() - 1000;
```
DerDreschner, line 57: I've read the specifications again and that should be correct and necessary. I checked the returned time attributes against what the Bosch Smart Home Controller II returns before, but looks like they didn't implemented it correctly either...

src/utils/timeService.ts
```diff
@@ -0,0 +1,138 @@
...
+        localTime = standardTime + cachedTimeData.dstShift;
+    }
+
+    return {
+        time: currentTime,
+        // Bit 0: Master clock
+        // Bit 1: Synchronized
+        // Bit 2: Master for Time Zone and DST
+        // Bit 3: Time synchronization SHOULD be superseded
+        timeStatus: 0x1111,
```
DerDreschner: I'm not sure if I should set bit 3 to 1 or not. What's your opinion on that, @Nerivec / @Koenkk ?

src/utils/timeService.ts
```diff
@@ -0,0 +1,137 @@
...
+    cachedTimeData.timeZone = timeZoneDifferenceToUtc;
+    cachedTimeData.dstStart = timestampToZigbeeUtcTime(dstStart);
+    cachedTimeData.dstEnd = timestampToZigbeeUtcTime(dstEnd);
+    cachedTimeData.dstShift = dstShift;
+    cachedTimeData.validUntilTime = timestampToZigbeeUtcTime(validUntilTime);
+}
+
+export function getTimeClusterAttributes(): TClusterAttributes<"genTime"> {
+    const currentDate = new Date();
+    const currentTime = timestampToZigbeeUtcTime(currentDate.getTime());
```
Nerivec: I'm thinking we should rename this const `currentZigbeeTime` so as to not confuse JS UTC and ZB UTC throughout the file.
currentDate => JS Date
currentTime => JS UTC
currentZigbeeTime => ZB UTC

src/utils/timeService.ts
Nerivec: `0x` => `0b` for bits.

src/utils/timeService.ts
DerDreschner: Is it more clear now?

src/utils/timeService.ts
Nerivec: > The Master and Synchronized bits together provide information on how closely the Time attribute conforms to the time standard.
> 
> The Master bit specifies whether the real time clock corresponding to the Time attribute is internally set to the time standard. This bit is not writeable – if a value is written to the TimeStatus attribute, this bit does not change.
> 
> The Synchronized bit specifies whether Time has been set over the network to synchronize it (as close as MAY be practical) to the time standard (see 3.12.1). This bit must be explicitly written to indicate this – i.e., it is not set automatically on writing to the Time attribute. **If the Master bit is 1, the value of this bit is 0.**
> 
> If both the Master and Synchronized bits are 0, the real time clock has no defined relationship to the time standard (e.g., it MAY record the number of seconds since the device was initialized).
> 
> The MasterZoneDst bit specifies whether the TimeZone, DstStart, DstEnd and DstShift attributes are set internally to correct values for the location of the clock. If not, these attributes need to be set over the network. This bit is not writeable – if a value is written to the TimeStatus attribute, this bit does not change.
> 
> Devices SHALL synchronize to a Time server with the highest rank according to the following rules, listed in order of precedence:
> - A server with the Superseding bit set SHALL be chosen over a server without the bit set.
> - A server with the Master bit SHALL be chosen over a server without the bit set.
> - The server with the lower short address SHALL be chosen (note that this means **a coordinator with the Superseding and Master bit set will always be chosen as the network time server**).
> - A Time server with neither the Master nor Synchronized bits set SHOULD not be chosen as the network time server.

Looks like the value would be `0b1101`, but that thing about synchronized is confusing, and previous value was `0b0011`...
Since it's a reply, I'm assuming Master and MasterZoneDst bits don't matter, since they can't be written (used to determine precedence of server).

So, I'd go with `0b1111`.

DerDreschner: Ahh, true! Didn't looked into that as I assumed the value before was correct (Master clock, synchronized - value 0b0011). Changed.

Nerivec: Sorry, edited the reply above. I think safest is `0b1111`.

src/utils/timeService.ts
Nerivec: Don't need to be explicit for JS, that's assumed. It's the Zigbee one that's not.

src/utils/timeService.ts
DerDreschner: > Looks like the value would be `0b1101`, but that thing about synchronized is confusing, and previous value was `0b0011`... Since it's a reply, I'm assuming Master and MasterZoneDst bits don't matter, since they can't be written (used to determine precedence of server).

My way of understanding is that the Synchronized bit has to be 0, as the value wasn't set over *Zigbee*. Network refers to "via Zigbee", not "via NTP/etc". Thus `0b0011` isn't correct, as it can't be set over Zigbee when the Master bit is set (thus not writable via Zigbee by spec; it isn't in our implementation anyway).

src/utils/timeService.ts
DerDreschner: Ahh, yeah, gotcha! Sorry for my confusion :smile:

src/utils/timeService.ts
Nerivec: Since the attributes are described "as they are on the device", not "as they are transmitted by the server", I'm not sure how this applies, really.

Nerivec: Did you check what the Bosch controller sends when a device does a `read` on it (since you appear to have one)?

DerDreschner: It only replies with the requested attributes. The BTH-RM/-RM230Z asks the most attributes (doesn't calculate the `localTime` by itself and ask the coordinator, different to the BTH-RA), yet that doesn't include the `TimeStatus` attribute. As they reply with a wrong `dstEnd` value (has to be the last second of the DST period to match spec) as well as `dstShift` (it's set to 0x00000080 for whatever reason?) and devices used in i.e., Iceland continuously ask the coordinator for DST info as the firmware *always* expect to be used in a country with DST and don't accept the send default value, I wouldn't count on it that much anyway.

This is the answer when setting the time zone to Iceland on the Bosch SHC II:

image

This is the answer when setting the time zone to Germany on the Bosch SHC II:

image

The only devices from other manufacturers I have that read from the `genTime` cluster are from Aqara and they only read the `time` and `timeZone` attribute. I might power up my Aqara controller, my guess would be that I should add something like "device hasn't requested timezone info? Then include dstShift in timeZone".

But that doesn't solve the question here. :smile:

DerDreschner: The introduction to the `genTime` cluster includes the following sentence:

> The clock time MAY be read and also written, in order to synchronize the clock (as close as practical) to a time standard.

So, I think it's correct to leave the synchronized bit on `0`.

DerDreschner: > Since the attributes are described "as they are on the device", not "as they are transmitted by the server", I'm not sure how this applies, really.

I think the word "device" refers to coordinator/server here anyway as all attributes from the `genTime` cluster are written under "3.12.2 Server" and clients don't contain any attributes.

DerDreschner: The native Aqara hub doesn't reply with any useful `TimeZone` at all for the device (smart plug) and uses time === local time.

image

Nerivec: `80` (with according zero-padding) is used as non-value for INTx in the ZCL spec. It's the equiv of `FF` for UINTx.
So `-0x80000000` is the non-value for INT32.
I'm going to take a wild guess, and say that Bosch did not intend to sell these devices outside a specific set of countries, that have proper configs, the rest don't (i.e. using non-values).

I expect most firmware just silently ignore values that they didn't request/don't expect, so, can send more without worrying too much about it.

> word "device" refers to coordinator/server

It was more about the "stored" vs "transmitted". From the spec, I'd say it's likely that the timeStatus doesn't need to be included at all during this `readRsp` (which would fit with the Bosch SHC frames). Seems to be more about explicit reading/writing.
Note: The previous code is very old (> 5y).

Nerivec: Unrelated note: the 0x1111 mistake from before should have failed some tests. This would have been trying to write a value above the BITMAP8 limit (4369 > 255) during frame serialization. From what I can see, they all passed even before the fix? _Something's not right somewhere, I'd expect the controller to at least have 1 test that takes this codepath and serializes the `readRsp`._

DerDreschner: Regarding the unrelated note: It didn't fail, that's right. JavaScript only supports `0b` since ES6, that may be the reason why it's handled like UINT8 in the code (and why I originally used `0x1111` as I've rarely seen `0b` annotation for BITMAP8 values throughout the code base): https://github.com/Koenkk/zigbee-herdsman/blob/master/src/zspec/zcl/buffaloZcl.ts#L819

Nerivec: `0x1111` is not the same as `0b1111`
`0x` is for hexadecimal, `0b` is for binary
`0x1111` => `4369`
`0b1111` => `15`
It's just "pretty display" though, underneath is the actual number. Hence, 0x1111 should have failed at least one test, since writeUInt8 does not allow writing values above 255:
```
> b.writeUInt8(0x1111)
Uncaught:
RangeError [ERR_OUT_OF_RANGE]: The value of "value" is out of range. It must be >= 0 and <= 255. Received 4369
    at writeU_Int8 (node:internal/buffer:747:11)
    at Buffer.writeUInt8 (node:internal/buffer:757:10) {
  code: 'ERR_OUT_OF_RANGE'
}
```

But since no test failed, we'd have to assume that this codepath is never fully taken (lack of test coverage somewhere).

DerDreschner: > `80` (with according zero-padding) is used as non-value for INTx in the ZCL spec. It's the equiv of `FF` for UINTx. So `-0x80000000` is the non-value for INT32.

Ahh, thanks! Learned something.

> I'm going to take a wild guess, and say that Bosch did not intend to sell these devices outside a specific set of countries, that have proper configs, the rest don't (i.e. using non-values).

Yeah, that might be true as the website only offers localized variants for EU countries. But the app/native coordinator supported those time zones. I didn't to any "magic" with this response for Iceland. So, I would have expected these cases to be handled properly. As I don't understand the typical "manager-thinking", guess other people would seem that as sensible way to cut costs.

> I expect most firmware just silently ignore values that they didn't request/don't expect, so, can send more without worrying too much about it.

We only send the requested attributes as well. @Koenkk proposed this variant already in https://github.com/Koenkk/zigbee-herdsman/pull/1517/files#r2399833169. This is the behavior that the ZCL specifies:

> The command frame SHALL contain a read attribute status record for each attribute identifier specified in the original Read Attributes or Read Attributes Structured command.

Source: [ZCL spec, site 66](https://zigbeealliance.org/wp-content/uploads/2021/10/07-5123-08-Zigbee-Cluster-Library.pdf)

So yeah, I would also just keep it in case it's being requested by any device.

> It was more about the "stored" vs "transmitted". From the spec, I'd say it's likely that the timeStatus doesn't need to be included at all during this `readRsp` (which would fit with the Bosch SHC frames). Seems to be more about explicit reading/writing. Note: The previous code is very old (> 5y).

~~Ahh, that's what you meant. My guess is that it's specified in the ZCL, but rarely used by any devices.~~ As we throw away unrequested attributes anyway, I would follow the spec here and settle on `0b1101` as `timeStatus` and `time = lastSetTime`.

DerDreschner:
I've looked into the converters to see how other devices handle `genTime` requests apart from what is being done in zigbee-herdsman. Here are a few findings:

- Usage of standard UNIX timestamp instead of Zigbee UTC time by [Sonoff in at least one case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sonoff.ts#L2006).
- The behavior I found above for Aqara is already [being implemented and used in this case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/lumi.ts#L2080) and [that strangely global case](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/index.ts#L662) (index.ts being used).
- [Multiple devices](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/heiman.ts#L870) that [don't request the time](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/danfoss.ts#L14) during interview and expect it to be set by [the coordinator](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/sunricher.ts#L155). Regarding the discussion above, it would be interesting to see how the `timeStatus` looks like on these devices. Some devices [expect the time regularly as well](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/lib/modernExtend.ts#L701). My bet would be that these devices expect it every time their `ValidUntilTime` expires or is about to expire.
- Multiple devices try to read the time from the coordinator, but the [read request is being ignored](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/fromZigbee.ts#L5158) for some reason.
- Some devices seem to use their [`Time`](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/slacky_diy.ts#L2289) and `LastSetTime` attributes (*not the server side we discuss here*) as "[last time since reboot](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/mindy.ts#L48)". I don't know if that is because these devices expect the time to be explicitly set as well or if that's the way it should be. Considering [this abuse](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/devices/bacchus.ts#L543) of the `dstStart` and `dstEnd` fields, my bet would be the latter...

It might be sensible to expose the timeService to the converters for this cases, as a lot of them just copied the previous code we now replace. Maybe we should implement a `utils.time` file there or some methods in `modernExtend.ts` to handle such non-specification-compliant cases (or at least a subset of commonly used cases) in a central place.

src/utils/timeService.ts
Nerivec: > We only send the requested attributes

Right, I forgot about this. That also explains why the tests did not fail before with 0x1111, since the attribute is likely dropped, so it was never serialized.

> As we throw away unrequested attributes anyway, I would follow the spec here and settle on 0b1101 as timeStatus

I checked a few different codebases ([e.g.](https://github.com/dresden-elektronik/deconz-rest-plugin/blob/master/time.cpp#L256)), the correct value would appear to be `0b1101` (`0x0d`, `13`) indeed.
Thing is, it was explicitly set to 3 for a device looks like, so turning off that bit could have undesired consequences :
- https://github.com/Koenkk/zigbee-herdsman/issues/122
- https://github.com/Koenkk/zigbee-herdsman/pull/126

And quite a few devices appear to actually request read of `timeStatus` attribute:
- https://github.com/Koenkk/zigbee2mqtt/issues?q=is%3Aissue%20%22read%22%20%22timeStatus%22

In ZHC, writes don't appear to be very consistent:
- https://github.com/search?q=repo%3AKoenkk%2Fzigbee-herdsman-converters%20timeStatus&type=code

@Koenkk what do you think? On one hand, current value could be the source of bugs (the crazy hard to find kind...), on the other, fixing it could introduce undesired behavior (for _??_ devices).

test/controller.test.ts
```diff
@@ -2890,29 +2927,77 @@ describe("Controller", () => {
...
+        expect(message.payload[0].attrId).toStrictEqual(Zcl.Clusters.genTime.attributes.time.ID);
+        expect(message.payload[0].dataType).toStrictEqual(Zcl.DataType.UTC);
+        expect(message.payload[0].status).toStrictEqual(Zcl.Status.SUCCESS);
+        expect(message.payload[0].attrData).toStrictEqual(expectedTime);
+
+        // TimeStatus
+        expect(message.payload[1].attrId).toStrictEqual(Zcl.Clusters.genTime.attributes.timeStatus.ID);
+        expect(message.payload[1].dataType).toStrictEqual(Zcl.DataType.BITMAP8);
+        expect(message.payload[1].status).toStrictEqual(Zcl.Status.SUCCESS);
+        expect(message.payload[1].attrData).toStrictEqual(expectedTimeStatus);
```
DerDreschner, line 2939: @Nerivec : Shouldn't this unit test have failed with the `0x1111` case as this is the response? Or is this just the prepared response and it's being serialized (and should have failed) in the adapter code?

Nerivec: Should be fixed for this test (though no doubt other tests would also need the logger check).
Since `Device` is catching the error, and just logging it, if readResponse fails, tests can never fail on this.

src/utils/timeService.ts
DerDreschner: > Thing is, it was explicitly set to 3 for a device looks like, so turning off that bit could have undesired consequences

Looks like https://github.com/markuspf tried to implement support for the device in question before https://github.com/dmpolukhin did it. In Markus code, he would have set the correct value for the synchronized bit: https://github.com/Koenkk/zigbee-herdsman/issues/122#issuecomment-573187288

So, it appears to be a mistake to use 3 there and it just spread throughout various places from there.

test/controller.test.ts
DerDreschner: Thanks! :+1:

src/index.ts
```diff
@@ -7,6 +7,7 @@ export type * as Events from "./controller/events";
 export type * as Models from "./controller/model";
 export type * as Types from "./controller/tstype";
 export {setLogger} from "./utils/logger";
+export {getTimeClusterAttributes} from "./utils/timeService";
```
DerDreschner, line 10: I hope this export is okay. It's to prevent duplicate code in the converters once this PR is being merged (see my comment here: https://github.com/Koenkk/zigbee-herdsman/pull/1517#issuecomment-3368641166 ).

src/utils/timeService.ts
Koenkk: > @Koenkk what do you think? On one hand, current value could be the source of bugs (the crazy hard to find kind...), on the other, fixing it could introduce undesired behavior (for ?? devices).

Let's try to fix it and monitor for issues causing this with the next release.

Koenkk:
> Multiple devices try to read the time from the coordinator, but the [read request is being ignored](https://github.com/Koenkk/zigbee-herdsman-converters/blob/master/src/converters/fromZigbee.ts#L5158) for some reason.

It's an ignore converter to suppress the "No converter available.." message, but since it's not logged anyway for these [clusters](https://github.com/Koenkk/zigbee2mqtt/blob/4d247f3ae129ee7318eeede877ffdbbd27444693/lib/extension/receive.ts#L121) we can remove this. This converter does not prevent zigbee-herdsman from responding, it was purely there to suppress the log  message (before `ignoreClusters` was introduced). I'll cleanup the converters.

Nerivec:
> I'll cleanup the converters.

Mighty nice cleanup!

DerDreschner:
@Nerivec: Are there any open tasks for me to do that block a possible merge?

Koenkk:
Thanks!

PR #1526, opened 2025-10-08
feat: automatically backup on `reset`

Nerivec:
Prevention for cases like https://github.com/Koenkk/zigbee2mqtt/issues/28850#issuecomment-3380912045

Instead of just backing up the db, this would backup everything from data path in case `Adapter.start` returns `reset`.
Technically, this would be better in Z2M, since we're not handling just herdsman-specific files. However, that requires a bit more refactoring, where we move the stuff from `Controller.start` that's using `startResult` into a separate function (e.g. `onPostStart`), so Z2M can do its backup after `start`, but before `onPostStart`.

@Koenkk what do you think? I'm leaning towards doing this properly in Z2M, but that's a breaking change for ZH (API).

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert";
 import events from "node:events";
 import fs from "node:fs";
+import path from "node:path";
 import mixinDeep from "mixin-deep";
 import {Adapter, type Events as AdapterEvents, type TsType as AdapterTsType} from "../adapter";
 import {BackupUtils, wait} from "../utils";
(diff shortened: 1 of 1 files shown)


Koenkk:
> I suppose we could only backup database.db & coordinator-backup.json to fit with ZH layer (and uses outside Z2M), but that leaves configuration.yaml not backed up.

I would prefer this, Zigbee2MQTT doesn't change anything in `configuration.yaml` when `startResult == 'reset'`, so backing-up these two files is enough.

src/controller/controller.ts
```diff
@@ -344,6 +345,22 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
         Group.resetCache();
     }
 
+    private backupData(): void {
```
Koenkk, line 348: I think similar to `databaseBackupPath`, the client should indicate the backup path (in case one is generated), maybe we can have e.g. `dataBackupPath = 'data/backup/TIMESTAMP' provided by z2m.

Nerivec: So we only use `dataBackupPath` (where both files are copied if `reset` triggered) and use a settings migration to get rid of `databaseBackupPath` on Z2M side?

Koenkk: No setting migration is needed, as the `databaseBackupPath` is not a setting: `databaseBackupPath: data.joinPath("database.db.backup"),`

PR #1530, opened 2025-10-18
fix: Add missing parameters to `initTestMode` command

DerDreschner:
The `initTestMode` command definition do miss the payload described in the ZCL specification:

image

This pull request adds the necessary parameters so I don't have to declare them myself anymore in the Bosch lib.

### DIFF
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -2404,11 +2404,7 @@ describe("Controller", () => {
                         ],
                         name: "enrollRsp",
                     },
-                    initNormalOpMode: {
-                        ID: 1,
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

Nerivec:
All the missing spec params/attrs should already be in https://github.com/Koenkk/zigbee-herdsman/pull/1503

DerDreschner:
Yep, these parameters are indeed already being included in #1503. As I overwrite these in the Bosch lib, I wanted to have them right now to prevent duplicate and unclear code.

PR #1534, opened 2025-10-23
fix: Add missing parameters for `initTestMode` to `clusters-types.ts`

DerDreschner:
In the previous pull request (#1530), I forgot to add the parameters to the `clusters-types.ts` file. Therefore, the parameters where not usable with the `endpoint.command()` function.

This is being fixed with this pull request.


Koenkk:
Thanks!

PR #1535, opened 2025-10-26
fix: improve GP join handling

Nerivec:
- do not assume successful join when UNICAST pairing (GPP could return non-`SUCCESS` default rsp (e.g. `INSUFFICIENT_SPACE`)
- ember: use UNICAST when commissioning with coordinator

CC: @chris-1243


Nerivec:
@Koenkk keeping this as draft until next release. The ember change might need some tweaking, I don't have one of the "weirder" GP devices...

Koenkk:
Thanks!

PR #1538, opened 2025-10-28
fix: Improve backup corruption message

Koenkk:
Based on feedback given in https://github.com/Koenkk/zigbee2mqtt/discussions/11767

### DIFF
--- a/src/adapter/utils.ts
+++ b/src/adapter/utils.ts
@@ -9,10 +9,12 @@ export function isTcpPath(path: string): boolean {
 }
 
 export function parseTcpPath(path: string): {host: string; port: number} {
-    const str = path.replace("tcp://", "");
+    // built-in extra validation
+    const info = new URL(path);
+
     return {
-        host: str.substring(0, str.indexOf(":")),
-        port: Number(str.substring(str.indexOf(":") + 1)),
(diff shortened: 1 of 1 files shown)


src/adapter/utils.ts
```diff
@@ -0,0 +1,32 @@
...
+export function readBackup(path: string): Models.UnifiedBackupStorage | Models.LegacyBackupStorage | undefined {
+    if (!existsSync(path)) {
+        return undefined;
+    }
+
+    try {
+        return JSON.parse(readFileSync(path).toString());
+    } catch (error) {
+        throw new Error(
+            "[BACKUP] Coordinator backup is corrupted. This can happen due to filesystem corruption. " +
```
Nerivec: No need for the `+`, should just backtick everything.

Nerivec:
Somewhat related, should `LegacyBackupStorage` be deprecated (removal in 3.0.0)?

Should also look into moving the entire "port" (serial or tcp) out of drivers, we can init it from Adapter, and send it to the driver.

Koenkk:
Agree with both points, I think the only con of removing `LegacyBackupStorage` is that users that update from very old z2m versions will get an error.

Ready for merge?

PR #1542, opened 2025-11-05
fix: update zigbee-on-host to 0.2.0

Nerivec:
> [!IMPORTANT]
> Breaking change. Older saves will no longer be supported. The API of ZoH drastically changes.

https://github.com/Nerivec/zigbee-on-host/commits/main/?since=2025-10-02

CC: @chris-1243


Nerivec:
This needs tons of testing in various scenario, merging it so interested parties can easily test it in current Z2M dev!

PR #1544, opened 2025-11-07
feat: support custom stack config for ZoH

Nerivec:
Uses same mechanism as ember for now (`zoh_config.json` stored in Z2M data folder).
```ts
interface StackConfigJSON {
    /** if true, trigger serial dtr/rts flipping to skip bootloader (meant for TI hw) */
    tiSerialSkipBootloader: boolean;
    /** EUI64 used for the adapter -- 0x${hex} format */
    eui64: string;
    /** @see https://nerivec.github.io/zigbee-on-host/types/spinel_spinel.StreamRawConfig.html */
    ccaBackoffAttempts: number;
    /** @see https://nerivec.github.io/zigbee-on-host/types/spinel_spinel.StreamRawConfig.html */
    ccaRetries: number;
    /** @see https://nerivec.github.io/zigbee-on-host/types/spinel_spinel.StreamRawConfig.html */
    enableCSMACA: boolean;
}
```
Defaults are:
```json
{
    "tiSerialSkipBootloader": false,
    "eui64": "0x4d325a6e6f486f5a",
    "ccaBackoffAttempts": 1,
    "ccaRetries": 4,
    "enableCSMACA": true
}
```

`tiSerialSkipBootloader` should allow TI hw that default to bootloader on plug in to work properly. Due to the myriad of possible adapters compatible with ZoH, we can't use an automatic fallback trigger like zstack (could end up triggering undesired behavior) and have to rely on manual config (_unless someone has a better idea?_).

CC: @chris-1243

### DIFF
--- a/src/adapter/zoh/adapter/zohAdapter.ts
+++ b/src/adapter/zoh/adapter/zohAdapter.ts
@@ -52,9 +52,9 @@ interface StackConfig {
     enableCSMACA: boolean;
 }
 
-export interface StackConfigJSON extends Omit<StackConfig, "eui64"> {
+export interface StackConfigJSON extends Partial<Omit<StackConfig, "eui64">> {
     /** 0x${hex} format */
-    eui64: string;
+    eui64?: string;
 }
 
 const DEFAULT_REQUEST_TIMEOUT = 15000;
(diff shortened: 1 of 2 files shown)


Koenkk:
Thanks!

PR #1547, opened 2025-11-09
chore: LiXee ZLinky: remove cluster definition (moved to ZHC)

sbraz:
See
https://github.com/Koenkk/zigbee-herdsman-converters/pull/10704.

Also update tests to use another cluster.


sbraz:
Obviously do not merge until the converter works properly with the local cluster definition :)

test/controller.test.ts
```diff
@@ -4037,30 +4037,6 @@ describe("Controller", () => {
         }).rejects.toThrow(`Invalid attribute '99999' for cluster 'genBasic'`);
     });
 
-    it("Should replace legacy configured reportings without manufacturerCode", async () => {
```
sbraz, line 4040: Please let me know if dropping this test was the right decision or if it needs to be fixed to use another cluster.

Koenkk: If possible, replace it with a different cluster.

sbraz: Any of those that are still present in this repo?

Koenkk: I think this test can use any cluster which has a manufacturer code, e.g. https://github.com/Koenkk/zigbee-herdsman/blob/6384ef24d23d777a9e08dbdd7878dbeaee5e918c/src/zspec/zcl/definition/cluster.ts#L4915

sbraz: Thanks, done.

Koenkk:
Thanks!

PR #1549, opened 2025-11-10
fix: zoh: use unicast for GP

Nerivec:
@Koenkk the `wasBroadcast` prop was previously used to gate against using unicast when GP proxy was not present (no other use). Since the code now handles this properly upstream (use coord), I'm pretty sure it should be completely removable.


Koenkk:
Thanks!

PR #1551, opened 2025-11-10
feat!: archive data before "overwrite" situations

Nerivec:
Supersedes https://github.com/Koenkk/zigbee-herdsman/pull/1526

Creates a "backup-2025-01-01.01-01-01" folder (with database.db & coordinator-backup.json) before:
- adapter "reset" triggers a wipe of previous data from controller
- controller `backup`
  - when triggered by `stop`

Removes `databaseBackupPath` Controller options (no longer used).
Removes `supportsBackup` Adapter function in favor of a plain variable (defaults to true, adapters without can override, same as `hasZdoMessageOverhead`).

TODO: 
- [ ] test coverage
- [ ] update Z2M to match new API

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -1,7 +1,7 @@
 import assert from "node:assert";
 import events from "node:events";
 import fs from "node:fs";
-import {basename} from "node:path";
+import {basename, join} from "node:path";
 import mixinDeep from "mixin-deep";
 import {Adapter, type Events as AdapterEvents, type TsType as AdapterTsType} from "../adapter";
 import type {ZclPayload} from "../adapter/events";
@@ -43,6 +43,8 @@ interface Options {
      */
     acceptJoiningDeviceHandler: (ieeeAddr: string) => Promise<boolean>;
 }
+
+const ONE_WEEK_MS = 7 * 24 * 60 * 60 * 1000;
 
 const DefaultOptions: Pick<Options, "network" | "serialPort" | "adapter"> = {
     network: {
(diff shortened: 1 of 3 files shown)


src/controller/controller.ts
```diff
@@ -356,21 +355,51 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
...
+            try {
+                const backup = await this.adapter.backup(this.getDeviceIeeeAddresses());
+                const unifiedBackup = BackupUtils.toUnifiedBackup(backup);
+                const tmpBackupPath = `${this.options.backupPath}.tmp`;
+
+                fs.writeFileSync(tmpBackupPath, JSON.stringify(unifiedBackup, null, 2));
+                fs.renameSync(tmpBackupPath, this.options.backupPath);
+                logger.info(`Wrote coordinator backup to '${this.options.backupPath}'`, NS);
+            } catch (error) {
+                logger.error(`Unable to backup ${error}`, NS);
```
Koenkk: I don't think we should suppress the exception here, e.g. [this](https://github.com/Koenkk/zigbee2mqtt/blob/5797b56871215b8fac37cfeb4f702f73a09c4ae0/lib/extension/bridge.ts#L316) will now appear as succeeded even while it could have failed in reality.

Nerivec: Was mainly to avoid interruption of the shutdown flow [here](https://github.com/Koenkk/zigbee-herdsman/blob/61531ad912eec493fa179f449d76a7b7dcbf02f4/src/controller/controller.ts#L336). I suppose we could try/catch just that call instead.

Koenkk: Yes I agree, so we can do a try/catch there similar like for permit join.

Nerivec:
How do we handle rotating this stuff? I figure we don't want to be keeping 100 folders :sweat_smile:
Should we use a subfolder as default (like Z2M logs)? Static const, use a Z2M config for rotation params? Should it run on start? Have to keep in mind that a bad shutdown could end up doing a few retries (watchdog), each could end up "backing up" bad files, we want to avoid overwriting the last "good one"... as much as possible.

Sidenote: not sure how old coordinator backups will behave in all stacks with frame counters & such.

Koenkk:
Maybe do it time based? e.g. keep for a week

Nerivec:
Cleaning up... never was really happy with this one. Will re-work it all.

PR #1568, opened 2025-11-25
fix: Support ZBT-2 discovery

Koenkk:
Add definition + support setting default `rtscts` and `baudRate`

### DIFF
--- a/src/adapter/tstype.ts
+++ b/src/adapter/tstype.ts
@@ -6,7 +6,7 @@ export type UsbAdapterFingerprint = {
     productId: string;
     manufacturer?: string;
     pathRegex: string;
-    serialPortOptions?: Pick<SerialPortOptions, "baudRate" | "rtscts">;
+    options?: Pick<SerialPortOptions, "baudRate" | "rtscts">;
 };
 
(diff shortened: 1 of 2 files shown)


src/adapter/tstype.ts
```diff
@@ -6,6 +6,7 @@ export type UsbAdapterFingerprint = {
     productId: string;
     manufacturer?: string;
     pathRegex: string;
+    serialPortOptions?: Pick<SerialPortOptions, "baudRate" | "rtscts">;
```
Nerivec: Probably should make the props optional for future use (several adapters should have `rtscts:true` in ember section, but baudRate is rather "up-in-the-air" since multiple build sources exist).

Also probably `options` is enough for the name (this is already in USB namespace).

Koenkk: > Probably should make the props optional for future use

They are already (because they are optional in `SerialPortOptions`), so e.g. `options: {baudRate: 460800},` doesn't give an error, I guess that is what you mean?

> Also probably options is enough for the name (this is already in USB namespace).

done

PR #1569, opened 2025-11-27
fix(ignore): proper configure typing

Nerivec:
To go with https://github.com/Koenkk/zigbee2mqtt/pull/29854 (not needed since `any` typed, but still better :wink: )


Koenkk:
Thanks!

PR #1573, opened 2025-12-02
fix: check status when handling groups

Nerivec:
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/29913

CC: @CubeZ2mDeveloper


Koenkk:
Thanks!

PR #1576, opened 2025-12-04
fix: lower db save interval, remove save on ep report attr

Nerivec:
- lower DB save rate to 5min (instead of 1h) so new features (e.g. scenes) don't get too many sync problems in case of issues
- remove save on cluster attr report (too many triggers, especially when bad devices are involved)


Koenkk:
Thanks!

PR #1580, opened 2025-12-06
fix: guard against a `undefined` device on data receipt

### LINKED ISSUE #1574: `onZclData` throws `Cannot read properties of undefined (reading 'manufacturerName')`
On pairing a new device, I saw this as the device left the network prematurely:

```
warning:         z2m: Device '0x...' left the network
error:         zh:controller:endpoint: Tried to get unknown/deleted device 0x... from endpoint 1.

TypeError: Cannot read properties of undefined (reading 'manufacturerName')
    at Device.onZclData (.../zigbee-herdsman/src/controller/model/device.ts:431:64)
    at Controller.onZclPayload (.../zigbee-herdsman/src/controller/controller.ts:1017:13)
```

The only reference to `manufacturerName` in `onZclData` is on line 417:

https://github.com/Koenkk/zigbee-herdsman/blob/4dd94cac7e153a0cadd243344a4e15453f929c16/src/controller/model/device.ts#L413-L420

I'm happy to put a PR together which investigates typing [`getDevice()`](https://github.com/Koenkk/zigbee-herdsman/blob/4dd94cac7e153a0cadd243344a4e15453f929c16/src/controller/model/endpoint.ts#L192-L192=) to return `undefined` - it looks like I hit exactly the edge-case mentioned in https://github.com/Koenkk/zigbee-herdsman/pull/1330#discussion_r1978156103

bobrippling:
for example, if we're deleting or delete-pending a device

closes #1574

### DIFF
--- a/src/controller/controller.ts
+++ b/src/controller/controller.ts
@@ -991,7 +991,7 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
         }
 
         if (!device) {
-            if (typeof payload.address === "number") {
+            if (typeof payload.address === "number" && !Device.isDeletedByNetworkAddress(payload.address)) {
                 device = await this.identifyUnknownDevice(payload.address);
             }
 
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -323,6 +323,11 @@ export class Device extends Entity<ControllerEventMap> {
     }
 
     public async onZclData(dataPayload: AdapterEvents.ZclPayload, frame: Zcl.Frame, endpoint: Endpoint): Promise<void> {
+        if (!Device.devices.has(this.ieeeAddr)) {
+            // prevent race conditions where device gets deleted during processing
+            return;
+        }
+
         // Respond to enroll requests
         if (frame.header.isSpecific && frame.isCluster("ssIasZone") && frame.isCommand("enrollReq")) {
             logger.debug(`IAS - '${this.ieeeAddr}' responding to enroll response`, NS);
@@ -414,7 +419,7 @@ export class Device extends Entity<ControllerEventMap> {
         const commandHasResponse = frame.command.response !== undefined;
         const disableDefaultResponse = frame.header.frameControl.disableDefaultResponse;
         /* v8 ignore next */
-        const disableTuyaDefaultResponse = endpoint.getDevice().manufacturerName?.startsWith("_TZ") && process.env.DISABLE_TUYA_DEFAULT_RESPONSE;
+        const disableTuyaDefaultResponse = this.manufacturerName?.startsWith("_TZ") && process.env.DISABLE_TUYA_DEFAULT_RESPONSE;
         // Sometimes messages are received twice, prevent responding twice
         const alreadyResponded = this._lastDefaultResponseSequenceNumber === frame.header.transactionSequenceNumber;
 
@@ -601,6 +606,22 @@ export class Device extends Entity<ControllerEventMap> {
         }
 
         return devices;
+    }
+
+    /** Check if a device is explicitly deleted */
+    public static isDeletedByIeeeAddr(ieeeAddr: string): boolean {
+        Device.loadFromDatabaseIfNecessary();
+
+        return Device.deletedDevices.has(ieeeAddr);
+    }
+
+    /** Check if a device is explicitly deleted */
+    public static isDeletedByNetworkAddress(networkAddress: number): boolean {
+        Device.loadFromDatabaseIfNecessary();
+
+        const ieeeAddr = Device.nwkToIeeeCache.get(networkAddress);
+
+        return ieeeAddr ? Device.deletedDevices.has(ieeeAddr) : false;
     }
 
     /**
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -2615,6 +2615,51 @@ describe("Controller", () => {
         expect(events.message.length).toBe(0);
     });
 
+    it("Receive zclData with device deleted during processing (prevent race condition)", async () => {
+        const frame = Zcl.Frame.create(
+            Zcl.FrameType.GLOBAL,
+            Zcl.Direction.CLIENT_TO_SERVER,
+            true,
+            undefined,
+            40,
+            "read",
+            "genTime",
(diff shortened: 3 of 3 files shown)


bobrippling:
The endpoint is currently safe - on the issue (#1574) there was the idea to skip endpoint creation here and instead return early. I can include that in this PR too if desired

https://github.com/Koenkk/zigbee-herdsman/blob/db634fdbc6b79904012c41570b7734c0f8d46c77/src/controller/controller.ts#L1021-L1028

Nerivec:
Should check against `=== undefined` not null here.
This would not fix all possible race conditions though, just the specific issue linked. I think this needs to be done entirely in Controller (caller), we should not call the Device class when the device is deleted, and we should short-circuit the bottom processing in Controller (no trying to identify, no endpoint creation, no emitting, nothing really).

bobrippling:
> Should check against `=== undefined` not null here.

Thanks for the quick review - I went for `== null` since it's equivalent to `=== undefined || === null`, to guard against future changes which might return `null` (since we don't have this protection from the types). But I think the new approach (`.isDeleted()`) is better, do you agree?

> This would not fix all possible race conditions though, just the specific issue linked. I think this needs to be done entirely in Controller (caller), we should not call the Device class when the device is deleted, and we should short-circuit the bottom processing in Controller (no trying to identify, no endpoint creation, no emitting, nothing really).

Sounds good, I've made those changes. Will sort out linting/format/tests if this change is more along the right lines

Nerivec:
Have to use `device.isDeleted` (no parens, it's a getter). It's also called too late.

But I think we need a more complex mechanism to properly check the state.
The RCs happen because we retrieve the device, so we have a ref to it, but by the time we make a call (like on an endpoint), the device is deleted (ref still valid though). So, a pre-emptive check alone would do no good, technically, it's already done (`find`, `by` methods don't return a deleted device unless specifically requested).

There are several things to fix though:
- if a device is deleted `identifyUnknownDevice` should return early, it's not unknown (have to add a method in Device to check against the static maps since it can be by ieee or nwk address).
- should not create endpoint if `isDeleted`
- should not call any other device/endpoint function if device `isDeleted` and said function is not appropriate for deleted

For the RCs, we need to come up with a good way to test this, so we can catch at least the major edge-cases, have a better idea of where things can go wrong exactly. i.e. write tests to find the bugs
RCs are hard to catch :cold_sweat:
We could splatter `isDeleted` checks everywhere, but that's rather ugly :grimacing:

@Koenkk any ideas?

Koenkk:
I think it will be very hard to prevent all the race conditions unless we decide to also return deleted device in https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/model/device.ts#L585, maybe that's the more safe and simple solution? (I did not really check the impact of what that change would be)

Nerivec:
Can't return deleted device there, we'd end up triggering logic for devices we don't want to.

In Controller ZCL listener, we have a valid Device:
https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/controller.ts#L998-L1000
we then grab the Endpoint (or [create it](https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/controller.ts#L1027)):
https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/controller.ts#L1019
we call the ZCL listener for the Device, we pass a valid Endpoint:
https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/controller.ts#L1127

**At some point after first snippet but before here, the device moved from `Device.devices` to `Device.deletedDevices`**

we call `getDevice()` from the Endpoint:
https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/model/device.ts#L417
`getDevice()` only grabs non-deleted Device BUT assumes Device is always present (save for the extra logging we added):
https://github.com/Koenkk/zigbee-herdsman/blob/dec6584c3ee46ebce6e83fb35514ecc1135bce77/src/controller/model/endpoint.ts#L192-L203

Since Endpoint should never be present without a Device, it's a good assumption, except when we carry an Endpoint ref for too long without checking if the Device got deleted.

---

Of course the simplest of fix for this scenario, would just be to (not sure why it isn't like that in the first place btw):
```diff
- endpoint.getDevice().manufacturerName
+ this.manufacturerName
```
But it won't solve the RC problem (nor the other possible RCs). It won't fail, but it will keep processing stuff for a deleted device (and potentially make a mess in Z2M/ZHC since they expect emitted stuff & co to be from valid devices, not to mention network requests to a device that no longer is present).

Maybe we start by fixing the first two points I mentioned in previous post, plus add a return `if (!Device.devices.has(this.ieeeAddr))` at the top of `Device.onZclData`.
We can review other cases as they show up. I think the problem will be largely limited to emit scenario (where a ref can stay for extended period).

Koenkk:
> we'd end up triggering logic for devices we don't want to.

What kind of logic?

Nerivec:
I'm not sure I understood what you meant, because this would be everything really, from unnecessary/invalid MQTT publishing, to requests to devices that left the network (i.e. 100% fail)...

Nerivec:
I think this should cover most cases. It's not perfect, still probably several ways to RC, but it should better handle the case of `Device.onZclData` and the identify.

Note: `Device.onZclData` should be refactored a bit too (another PR), seems it's checking a bunch of `if` needlessly.

bobrippling:
> Since Endpoint should never be present without a Device, it's a good assumption, except when we carry an Endpoint ref for too long without checking if the Device got deleted.

I guess this isn't really something we can enforce either - I had thought about forcing access to go through a sync function:

```diff
-device.getEndpoint(...)
+device.withEndpoint(endpoint => { // can't be async by the typesystem
+    ...
+})
```

but the same problem exists in that the `endpoint` ref can leak away. But it might help people think twice?

Koenkk:
Thanks!

PR #1583, opened 2025-12-08
fix: ember: ASH layer optimizations

Nerivec:
This is mostly nitpicking, but hopefully can draw a bit more speed out of the low levels...

@Koenkk can you take a look? Changes at this level... :cold_sweat:
Code changes should be 1:1 in behavior with previous (except the emit), each commit with a specific focus (easier to read).


Koenkk:
Thanks!

PR #1584, opened 2025-12-10
fix: biome floating promises detection

Nerivec:
I `catch` them all! :grin: 
There's also https://biomejs.dev/linter/rules/use-await-thenable/ that's coming in next version, should be added then.

@manup if you can confirm for deconz that we don't need to await the two calls.

### DIFF
--- a/biome.json
+++ b/biome.json
@@ -12,7 +12,7 @@
         "bracketSpacing": false
     },
     "files": {
-        "includes": ["**", "!package.json"]
+        "includes": ["**", "!package.json", "!!**/dist", "!!**/coverage"]
     },
     "linter": {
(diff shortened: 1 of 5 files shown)


Koenkk (approved):

manup:
Hi so these are for promises which may reject but there is no catch handler? Good catch ;)
I think it's fine to merge, but I need to check to add proper handling here to not just silently discard errors.

Nerivec:
https://biomejs.dev/linter/rules/no-floating-promises/#description
Yes, should either be `await`ed so they throw in-place, or with a `.catch`. Can do it in a follow-up PR, you'll need to check which is best (`await` would require changing some function signatures but is usually easier to read for flow - and no callback).

Koenkk:
Thanks!

PR #1587, opened 2025-12-10
fix: use cheaper async mutex in place of 1-queue

Nerivec:
Should be much cheaper/faster to run than `Queue(1)`, no job object creation, no filtering needed, simple shift op.

@Koenkk what do you think?


Koenkk (approved):
Looks good! Feel free to merge

Nerivec:
Did you do a run with zstack just in case? (I already did with ember.)

Koenkk:
yes, looks good!

PR #1598, opened 2025-12-16
fix: onZclPayload crash when invalid post read

Nerivec:
Can't fail in onZclPayload, so instead ignore attributes that don't pass validation.
@Koenkk you're more familiar with converters, thoughts?

Cf https://github.com/Koenkk/zigbee2mqtt/issues/30218 (not a full fix, as the ZHC Lixee converter will need to take care of the fact the siteId apparently comes in invalid, per spec, and with this PR would be omitted from response payload)

Also fixed a mock device that was invalid (well-buried and hiding coverage :sweat_smile:).


Koenkk:
Looks good, thanks!

Issue #1600, opened 2025-12-18
Using IPv6 address in port configuration prevents zigbee-herdsman from starting


spike77453:
When using an IPv6 address (ULA or GUA), `zigbee-herdsman` fails with error "No such file or directory...":

```
Starting Zigbee2MQTT without watchdog.
[2025-12-18 16:58:18] info:         z2m: Logging to console, file (filename: log.log)
[2025-12-18 16:58:18] info:         z2m: Starting Zigbee2MQTT version 2.7.1 (commit #unknown)
[2025-12-18 16:58:18] info:         z2m: Starting zigbee-herdsman (7.0.4)
[2025-12-18 16:58:18] info:         zh:zstack:znp: Opening SerialPort with {"path":"tcp://[fd73:5e46:b9f0:0:80b5:4eff:dead:beef]:6638","baudRate":115200,"rtscts":false,"autoOpen":false}
[2025-12-18 16:58:18] error:         z2m: Error while starting zigbee-herdsman
[2025-12-18 16:58:18] error:         z2m: Failed to start zigbee-herdsman
[2025-12-18 16:58:18] error:         z2m: Check https://www.zigbee2mqtt.io/guide/installation/20_zigbee2mqtt-fails-to-start_crashes-runtime.html for possible solutions
[2025-12-18 16:58:18] error:         z2m: Exiting...
[2025-12-18 16:58:18] error:         z2m: Error: Error: No such file or directory, cannot open tcp://[fd73:5e46:b9f0:0:80b5:4eff:dead:beef]:6638
```

The configuration file looks like this:

```
$ grep serial -A 2 configuration.yaml 
serial:
  port: tcp://[fd73:5e46:b9f0:0:80b5:4eff:dead:beef]:6638
  adapter: zstack
```

**Note**: This does not seem to be an IPv6 issue. The socket can be opened successfully when using a hostname with a AAAA resource record.

After creating a corresponding entry in `/etc/hosts` (or ensuring there's a corresponding AAAA record on the DNS server) and changing the configuration accordingly

```
$ echo "fd73:5e46:b9f0:0:80b5:4eff:dead:beef foo" >> /etc/hosts
$ sed -i 's|\(port: tcp://\)\(.*\)|\1foo:6638|g' configuration.yaml 
$ grep serial -A 2 configuration.yaml 
serial:
  port: tcp://foo:6638
  adapter: zstack
```

`zigbee-herdsman` starts up just fine:

```
[...]

PR #1601, opened 2025-12-20
fix: allow literal (RFC2732) IPv6 addresses in TCP URI

spike77453:
Allows [RFC2732](https://www.rfc-editor.org/rfc/rfc2732) formatted IPv6 address to pass test in `isTcpPath`.

It also wraps the URL object constructed from the URI given in the configuration file with [urlToHttpOptions](https://nodejs.org/api/url.html#urlurltohttpoptionsurl) so the resulting hostname is formatted as expected by [socket.connect](https://nodejs.org/api/net.html#socketconnectport-host-connectlistener)

See also #1600

### DIFF
--- a/test/adapter/adapter.test.ts
+++ b/test/adapter/adapter.test.ts
@@ -8,6 +8,7 @@ import {DeconzAdapter} from "../../src/adapter/deconz/adapter/deconzAdapter";
 import {EmberAdapter} from "../../src/adapter/ember/adapter/emberAdapter";
 import {EZSPAdapter} from "../../src/adapter/ezsp/adapter/ezspAdapter";
 import {SerialPort} from "../../src/adapter/serialPort";
+import {parseTcpPath} from "../../src/adapter/utils";
 import {ZStackAdapter} from "../../src/adapter/z-stack/adapter/zStackAdapter";
 import {ZBOSSAdapter} from "../../src/adapter/zboss/adapter/zbossAdapter";
 import {ZiGateAdapter} from "../../src/adapter/zigate/adapter/zigateAdapter";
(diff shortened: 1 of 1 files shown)


src/adapter/utils.ts
```diff
@@ -1,19 +1,20 @@
...
 }
 
 export function parseTcpPath(path: string): {host: string; port: number} {
     // built-in extra validation
-    const info = new URL(path);
+    const info = urlToHttpOptions(new URL(path));
 
     return {
-        host: info.hostname,
+        host: String(info.hostname),
```
Nerivec: `hostname` is always a string, no need for the wrapping here.

src/adapter/utils.ts
```diff
@@ -1,19 +1,20 @@
...
+import {urlToHttpOptions} from "node:url";
 import type * as Models from "../models";
 
 export function isTcpPath(path: string): boolean {
     // tcp path must be:
     // tcp://<host>:<port>
-    const regex = /^(?:tcp:\/\/)[\w.-]+[:][\d]+$/gm;
+    const regex = /^(?:tcp:\/\/)\S+?[:]\d+$/gm;
     return regex.test(path);
 }
```
Nerivec, line 15: I'm wondering if we should just try/catch a `new URL` call here, and let internal mechanism validate it. Should be more thorough.
```ts
try {
    const url = new URL(path);

    return true;
} catch {
    return false;
}

File: src/adapter/utils.ts (hunk 1 above)
Comment (spike77453) [reply src/adapter/utils.ts]: That unfortunately doesn't work for me:
```
src/adapter/utils.ts:17:9 - error TS2322: Type 'string | null | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

17         host: info.hostname,
           ~~~~

  src/adapter/utils.ts:12:46
    12 export function parseTcpPath(path: string): {host: string; port: number} {
                                                    ~~~~
    The expected type comes from property 'host' which is declared here on type '{ host: string; port: number; }'
```

src/adapter/utils.ts
spike77453: Works for me. I've updated the branch accordingly.

Koenkk:
Could you also add a test case for this?

Koenkk:
Thanks!

PR #1605, opened 2025-12-26
Add Schneider Electric specific thermostat parameters to clusters

svhelge:
Add attributes for Schenider Electric thermostats, see https://github.com/user-attachments/files/16744437/RTC-6754012INT-cFMT_16A_F_Zigbee-29.pdf


svhelge:
@Koenkk, I have no idea why the check failed. Doesn't seem to be related to the changes I did.

svhelge:
@Koenkk, I got a Biome error. Copilot tells me the following:
- Biome reports an internal error: "unusually large amount of types which exceeded the limit of 200,000" for two files:
  - src/adapter/ember/uart/queues.ts
  - scripts/zap-update-clusters.ts
- This is a Biome analyzer limitation/bug triggered by very large/generated TypeScript data structures; the check exits non‑zero.
Anything I should do in order to correct this?

Koenkk:
Could you add these through the new custom cluster API instead? We want to move all custom cluster to zigbee-herdsman-converters. Example: https://github.com/Koenkk/zigbee-herdsman-converters/blob/5e318afb70528f83a0ec136fccd2197ae31f351b/src/devices/sonoff.ts#L213

svhelge:
@Koenkk , how should custom attributes be added to standard clusters? The Sonoff example was for custom cluster as I could understand (m.deviceAddCustomCluster)

Koenkk:
Here you can find an example on how to extend a standard cluster: https://github.com/Koenkk/zigbee-herdsman-converters/blob/93513f46403fa70ed83bf616a7319ee653b7d81a/src/devices/wirenboard.ts#L651

PR #1609, opened 2026-01-03
fix: transitionTime not always present in genScenes recall

Nerivec:
I added this in the scenes PR, but seems I forgot to backport it in ZCL PR...
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/30431


Koenkk:
Thanks!

PR #1613, opened 2026-01-08
fix: Allow unbinding from non-existing group and bind

Koenkk:
Some devices are bind by default to certain groups, e.g. the [Shelly BLU Remote Control ZB](https://github.com/Koenkk/zigbee2mqtt/issues/29104) is by default bind to group 1 till 4. This is not desired as upon joining it starts controlling those groups. In the `configure()` we should unbind it from those groups however this is currently not possible because:
- The group might not exist in zigbee-herdsman
- The bind is not present in the `endpoint.binds` list.

This PR introduces the `force` parameter to bypass those checks.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -736,25 +736,26 @@ export class Endpoint extends ZigbeeEntity {
         this.getDevice().save();
     }
 
-    public async unbind(clusterKey: number | string, target: Endpoint | Group | number): Promise<void> {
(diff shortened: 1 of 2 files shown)


Nerivec:
For that particular problem, `clearAllBindings` would be more efficient (1 request)?
https://github.com/Koenkk/zigbee-herdsman/blob/ef9438b8d35407c0002e7eb335971fc3795dd4a5/src/controller/model/device.ts#L1257-L1279

Koenkk:
Ah much better and avoid hacking, I'll try!

PR #1617, opened 2026-01-19
Allow omitting optionsMask/optionsOverride for strict ZCL devices

mchacher:
Some Zigbee devices strictly follow the ZCL specification and reject commands when optional parameters are present, even if set to their default value (0).

This commit allows converters to omit the optionsMask and optionsOverride parameters from Level Control cluster commands when these parameters are not provided in the payload.

These parameters have MINIMUM_REMAINING_BUFFER_BYTES conditions in the ZCL specification, making them optional. However, some device manufacturers (e.g., Schneider Electric) implement strict validation and reject commands containing these optional parameters.

Fixes: Koenkk/zigbee2mqtt#30701


Nerivec:
Looks like this should be fixed in the converters for these devices.
Per spec:
> The OptionsMask & OptionsOverride fields SHALL both be present.

mchacher:
Thanks for the review @Nerivec 

You're absolutely right that the ZCL spec states "The OptionsMask & OptionsOverride fields SHALL both be present." However, the Schneider Electric WDE002386 dimmers implement strict validation that **rejects commands when these optional parameters are present with value 0**.

This creates a conflict between:
- **ZCL spec**: Says fields SHALL be present
- **Device behavior**: Rejects commands when fields are present with default values

The current behavior is that standard converters (like `m.light()`) always send `optionsMask: 0, optionsOverride: 0`, which causes these strict devices to reject all dimming commands with "Parameter 'optionsMask' is missing" errors.

**This PR proposes treating `value 0` as omission** during ZCL frame serialization for these two specific parameters, allowing:
1. ✅ Standard converters to work with strict devices (no custom code needed)
2. ✅ Backward compatibility for devices that expect these parameters (non-zero values still included)

The complementary PR #11337 simplifies the WDE002386 converter to use standard `m.light()` instead of requiring custom converter logic.

**Physical testing confirms:**
- Dimming works correctly with brightness levels 30, 75, 100, 200
- Tested on 2 physical dimmers (0x14b457fffe9b5e78 and 0x14b457fffe9b3ba5)
- All 1572 tests passing

Would you prefer a different approach to handle these strict devices?

Nerivec:
We cannot remove them (even if zero), as that would go against the proper spec.
Handling fevices that don't follow the spec should always go in ZHC unless specific generic problems need solving here.

I suppose we can check if the parameter has a condition, and depending on that condition, we can alter the resulting payload (similar to how it is done on the receiving side). So, if the payload does not contain a parameter _at all_, and it passes the condition, `Zcl.Frame` doesn't throw, and just ignores the parameter (a more generic version of this PR). That way would allow ZHC to handle this as needed for devices with quirks.
_But for this specific issue with optionsMask/optionsOverride, the default should remain passing these parameters as zero._

Let's see what Koenkk thinks.

PR #1624, opened 2026-01-24
fix: performance nitpicks

Nerivec:
- remove unnecessary (old) dep
- use fast-deep-equal instead of `parse(stringify)`: x3 perf
- use runtime lookup for ZCL clusters by ID: x2 perf (a bit more verbose, but worth it considering the number of calls, can shed some after custom clusters are moved out)
- perf change for Queue should be significant as well (no traversal + array creation for each `executeNext` call)
- use lambda for all logging with stringify calls

@Koenkk what do you think about the Controller 3 default params? Technically unnecessary (type-enforced)... I ignored errors for now, but I don't think we should keep the ignores, too dangerous, especially there.


Koenkk:
>  but I don't think we should keep the ignores, too dangerous, especially there.

I would propose to remove it

Nerivec:
I think we might want to enforce the others in another PR as well, should be cleaner.

PR #1630, opened 2026-01-27
fix: OTA detection & filesystem index

Nerivec:
Hotfix for v9 (technically breaking, changes return payload of `checkOta`):
- Previous scheme of using numbers for `available` not longer worked as intended with the new `source` payload being passed around: `checkOta` should just return true/false based on `source.downgrade`
- handle index JSON as filesystem path (rel/abs)


Koenkk:
Thanks!

PR #1636, opened 2026-02-07
fix: OTA padding skip

Nerivec:
`indexOf` is checking as uint8 if passed a number.
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/30976
Also refactored the tests a bit to behave more like real code in image loading.


Koenkk:
Thanks!

PR #1644, opened 2026-02-20
feat(deconz): add NWK_UPDATE energy scan API support

brendan-carroll:
## Summary
- add controller-level API support for NWK_UPDATE-based network energy scans
- implement deCONZ adapter handling for scan request/response mapping
- expose scanned channel mask + energy entry list consistently in controller result
- include a deCONZ driver robustness fix for TCP close handler context binding

## Motivation
This enables adapter-level energy scan diagnostics required by higher-level coordinator tooling and management UIs.

## Validation
- unit/integration checks on local branch for controller/deconz scan flow
- hardware-backed verification with deCONZ coordinator through Zigbee2MQTT extension path
- verified response includes per-channel energy values and channel mask

### DIFF
--- a/src/adapter/deconz/driver/driver.ts
+++ b/src/adapter/deconz/driver/driver.ts
@@ -921,7 +921,7 @@ class Driver extends events.EventEmitter {
             });
 
             // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-            this.socketPort!.once("close", this.onPortClose);
+            this.socketPort!.once("close", this.onPortClose.bind(this));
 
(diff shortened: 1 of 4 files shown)


Nerivec:
This is not "deconz-only" (actually not deconz at all).

About the feature itself, if I remember correctly, I took a stab at implementing this a while back (when refactoring ZDO stuff), but plenty of devices don't support this (at all or properly), and it can make a big mess because devices will change channel during the scan, and some devices don't change back to proper channel after... which means re-pairing would be mandatory.
Not even sure all coordinator types would support this (some might, but with special handling, not via generic ZDO request).

PR #1645, opened 2026-02-20
fix: perf: ensure manuf code unique

Nerivec:
- use single for-loop with early bail-out if will-fail
- regroup tests and improve coverage


Koenkk:
Thanks!

PR #1650, opened 2026-02-23
Remove elkoOccupancySettingClusterServer, already implemented in ZHC

svhelge:
The `elkoOccupancySettingClusterServer `seems to be implemented in [schneider_electric.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/ce520abf31550c81220b57766cfa7fcda58d77be/src/devices/schneider_electric.ts#L423) with `schneiderElectricExtend.addOccupancyConfigurationCluster()`, cluster `occupancyConfiguration`. Different manufacturerCode (ManufacturerCode.ADEO vs Zcl.ManufacturerCode.SCHNEIDER_ELECTRIC) , but same cluster ID 0xff19. Could not find any reference to (or use of) the cluster elkoOccupancySettingClusterServer in ZHC.

### DIFF
--- a/src/zspec/zcl/definition/clusters-types.ts
+++ b/src/zspec/zcl/definition/clusters-types.ts
@@ -7898,20 +7898,6 @@ export interface TClusters {
         commands: never;
         commandResponses: never;
     };
-    elkoOccupancySettingClusterServer: {
-        attributes: {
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1652, opened 2026-02-23
Remove elkoSwitchConfigurationClusterServer, not used in ZHC

svhelge:
The `elkoSwitchConfigurationClusterServer` is not used in ZHC. `manuSpecificSchneiderLightSwitchConfiguration` has the same cluster ID (`0xff17`) and the same attributes.  `manuSpecificSchneiderLightSwitchConfiguration` is used in [schneider_electric.ts](https://github.com/Koenkk/zigbee-herdsman-converters/blob/7eb0b699e500bb945a45bd3533fe0b5a47c721e6/src/devices/schneider_electric.ts#L127). 
Hence, I propose to remove `elkoSwitchConfigurationClusterServer`.

### DIFF
--- a/src/zspec/zcl/definition/clusters-types.ts
+++ b/src/zspec/zcl/definition/clusters-types.ts
@@ -7898,38 +7898,6 @@ export interface TClusters {
         commands: never;
         commandResponses: never;
     };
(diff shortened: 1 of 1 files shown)


Koenkk:
Thanks!

PR #1655, opened 2026-02-24
Move manuSpecificDoorman from ZH to ZHC

svhelge:
manuSpecificDoorman is not used in ZHC, but the cluster definition is added to yale.ts and commented out.


Koenkk:
Thanks!

PR #1656, opened 2026-02-24
fix: pass abort signal to controller start

Nerivec:
cf https://github.com/Koenkk/zigbee2mqtt/pull/31130

@Koenkk I think it's fine to throw here? Can handle this in Z2M catch block.
I think I've covered the main ops, but let me know if you see other places where we should call this.
Param is optional, so not a breaking change.

Also optimized the database stuff a bit (use Map, use clear/reset instead of loops).


Koenkk:
>  I think it's fine to throw here? Can handle this in Z2M catch block.

yes that's fine, thanks!

PR #1659, opened 2026-02-25
Remove schneiderSpecificPilotMode from ZH

svhelge:
The cluster schneiderSpecificPilotMode has a diferent name in ZHC ([schneider_electric.ts](https://github.com/svhelge/zigbee-herdsman-converters/blob/3646e25875c1a7f791787ab34d860e08d2ab233b/src/devices/schneider_electric.ts#L505)). Implemented schneider_pilot_mode with moderextend and updated devices in ZHC.


Koenkk:
Thanks!

PR #1664, opened 2026-02-26
Implemented fix for dresden elektronik ota updates

fst-dresden-elektronik:
This fix targets older dresden elektronik devices, such as the FLS-PP, which run on an outdated Zigbee stack. On these devices, OTA updates only work with a maximum data size of 40.

A change like this could be implemented as a configurable feature within the zigbee-herdsman-converters device definition files. For example, it could allow setting OTA-related parameters such as the Image Block Request timeout or maximum data size per device.

This approach would better support legacy devices from various manufacturers without hardcoding limitations directly into ota.ts, as done in this PR. While newer devices typically handle a maximum data size of 50 bytes without issue, some older models require smaller limits to function correctly, because of theire stack.

### DIFF
--- a/src/controller/helpers/ota.ts
+++ b/src/controller/helpers/ota.ts
@@ -307,8 +307,11 @@ function buildImageBlockPayload(
 ) {
     let dataSize = baseDataSize;
 
-    if (requestPayload.manufacturerCode === Zcl.ManufacturerCode.INSTA_GMBH) {
-        // Insta devices, OTA only works for data sizes 40 and smaller (= manufacturerCode 4474).
+    if (
+        requestPayload.manufacturerCode === Zcl.ManufacturerCode.INSTA_GMBH ||
+        requestPayload.manufacturerCode === Zcl.ManufacturerCode.DRESDEN_ELEKTRONIK_INGENIEURTECHNIK_GMBH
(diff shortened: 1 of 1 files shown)


Nerivec:
> `as a configurable feature within the zigbee-herdsman-converters device definition`

OTA doesn't go through converters, however it is fully configurable, per-request via:
https://github.com/Koenkk/zigbee-herdsman/blob/7394cf7073b24f6d1c917afbb3fdbb180386faf7/src/controller/tstype.ts#L62-L66
(which Z2M uses with either the [request-given settings](https://www.zigbee2mqtt.io/guide/usage/ota_updates.html#override-ota-settings), or the [default OTA settings](https://www.zigbee2mqtt.io/guide/usage/ota_updates.html#advanced-configuration)).
But it would be good to move these hardcoded values out of ZH indeed! Just not sure there's any easy way to do this without a big refactor. Probably the easiest at the moment would be via metadata in the zigbee-OTA repo... Maybe Koenkk will have a better idea :grin:

Koenkk:
> Probably the easiest at the moment would be via metadata in the zigbee-OTA repo.

That's the cleanest indeed, but IMO this is also fine for the time being, thanks!

PR #1663, opened 2026-02-26
fix!: improve performance with ZCL clusters handling

Nerivec:
Been wanting to investigate this... It's proving worthwhile, 2-6x "effective" perf, on stuff that's called so very often! :open_mouth: 
It also has the benefit of relieving the pressure on the GC (many, many short-lived objects were previously created).

The "core" of this is the removal of the runtime cluster generation (clone-like behavior), which was the expensive part.

Added `Readonly` on all related types to ensure nothing can mutate the `Clusters` object, since we no longer have clones (don't think anything was mutating the clones before either, but this should catch it if any).
Replaced in-object functions with plain utils to get attrs & cmds. [breaking change]

Also cleaned up the logic for finding custom clusters by ID:
- no spread needed in Utils (for merged custom with ZCL, the addCustomCluster takes care of it)
- removed the fallback-to-ZCL codepath on partial custom cluster match, fairly certain this is unused
  - see removed test `by ID ignoring same custom ID if Zcl is better match with manufacturer code`

@Koenkk if you can think of anything that would block this approach that we might need to investigate in more details... :grimacing:
I haven't done the migration to add `name` to Clusters entries yet (used init loop, with some type-ignore for now). It will add a bit of duplication (property name + name property), but I think it's worth it to avoid object-spread on module load and avoid any double memory scenario (i.e. to not re-create the whole object with `name` props)... _Can just write a quick script to do the update automatically once reviewed. Could keep the init loop, but it's not great, since have to type-assert and type-ignore, plus it's a big object to loop through at module load..._

At quick glance, the ZHC migration should be minimal:
- adding `name` prop to custom cluster entries (the bulk of it)
- changing `ClusterDefinition` => `Cluster` typing
- replace couple of cluster getters

Bench with CI runner:
https://github.com/Koenkk/zigbee-herdsman/actions/runs/22799356124#summary-66138674829

And a simple `getCluster` call bench:
```
 ✓ test/index.bench.ts > Benchmarks 50834ms
     name             hz     min     max    mean     p75     p99    p995    p999     rme   samples
   · old      285,659.43  0.0031  3.6109  0.0035  0.0033  0.0072  0.0179  0.0233  ±0.24%   1428298
   · new   12,752,366.31  0.0001  0.3682  0.0001  0.0001  0.0001  0.0002  0.0005  ±0.08%  63761832

 BENCH  Summary

  new - test/index.bench.ts > Benchmarks
    44.64x faster than old
```

[...]

PR #1667, opened 2026-02-27
Move manuSpecificAssaDoorLock cluster to ZHC

svhelge:


svhelge:
@Nerivec , some of the tests failed since the test code is looking for the cluster I removed...

Nerivec:
I think for most cases, we should be able to keep the tests by either:
- using `device.addCustomCluster` (if test is higher level, like controller)
- using `customClusters` arg (if test is lower level, like zcl)

You can probably create a top-level const for zcl.test.ts with a stripped version, should be fine, I think it's just matching, no need for the content:
```ts
const MANU_SPE_CUSTOM_CLUSTERS = {
    manuSpecificAssaDoorLock: {
        ID: 0xfc00,
        attributes: {},
        commands: {},
        commandsResponse: {},
    },
};
```
then you can pass it to the tests that need it.

Nerivec:
Removing this one removes the multi-ID cluster conflict in-ZCL, so I cleaned up that logic as well (else would always fail coverage). Will work out nicely with https://github.com/Koenkk/zigbee-herdsman/pull/1663 for perf :grin:

svhelge:
@Nerivec , I didn't add any ManufacturID to the manuSpecificAssaDoorLock cluster in ZHC - https://github.com/Koenkk/zigbee-herdsman-converters/pull/11612.  
Should it be added?

Nerivec:
That one doesn't have a manuf code, that's why it was nasty to deal with and had specific tests for it.
cf https://github.com/Koenkk/zigbee-herdsman/pull/653/changes#r1092484629

Koenkk:
Thanks!

PR #1669, opened 2026-02-28
fix: missing `write` on some closuresWindowCovering attrs

Nerivec:
window-covering

@Koenkk can remove:
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/642a41597118ed2758ba3508f8336806c68bec21/src/lib/lumi.ts#L2223-L2232
- https://github.com/Koenkk/zigbee-herdsman-converters/blob/642a41597118ed2758ba3508f8336806c68bec21/src/lib/ubisys.ts#L421-L426


Koenkk:
Thanks!

PR #1674, opened 2026-03-01
fix: set serialport DTR_CONTROL_DISABLE on Windows

Nerivec:
Based on Sonoff investigation, this would appear to be the cause of some adapters not working on Windows.

However, need to test the impact in various scenario.
TODO:
- [x] problem is confirmed on both SI & TI adapters using CP210x serial chip, so need to test both cases, at least
- [ ] adapters that worked before, still work as expected
- [ ] adapters that use DTR for bootloader control still work as expected (e.g. CC2652/CC1352 that require bootloader skipping on start)

_I do not have a Windows machine on hand at the moment, so can't test this at all._

CC: @humingchun

### DIFF
--- a/src/adapter/serialPort.ts
+++ b/src/adapter/serialPort.ts
@@ -1,5 +1,6 @@
 /* v8 ignore start */
 
+import {platform} from "node:os";
 import {type AutoDetectTypes, autoDetect, type OpenOptionsFromBinding, type SetOptions} from "@serialport/bindings-cpp";
 // This file was copied from https://github.com/serialport/node-serialport/blob/master/packages/serialport/lib/serialport.ts.
 import {type ErrorCallback, type OpenOptions, SerialPortStream, type StreamOptions} from "@serialport/stream";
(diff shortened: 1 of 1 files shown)


humingchun:
Hi @Nerivec, thank you very much for your work. We have tested your solution and SONOFF ZBDongle-E, Dongle-LMG21/-PMG24/-M all work well on Windows. <s>but ZBDongle-P (TI CC2652 chip) doesn't. We are investigating it right now.</s>

humingchun:
Sorry, my bad. ZBDongle-P also works with this fix.

Koenkk:
@humingchun so this PR can be merged?

Nerivec:
Need more feedback, this affects all adapters, we need at least feedback on other cases (where the fix is not needed, or DTR is used by bootloader).

Nerivec:
About this one, I'm not getting any feedback on the remaining tests... not enough users on Windows, and the ones that are, it's through 3rd party... I think we might have to merge it, then see if it breaks for some users (likely only after a release)...
It's limited to Windows, so, "tiny potential break surface" at least. :sweat_smile:

Koenkk:
@Nerivec agree, thanks!

humingchun:
> About this one, I'm not getting any feedback on the remaining tests... not enough users on Windows, and the ones that are, it's through 3rd party... I think we might have to merge it, then see if it breaks for some users (likely only after a release)... It's limited to Windows, so, "tiny potential break surface" at least. 😅

Hi, we have just finished testing this patch with more Zigbee adapters on both Windows and Ubuntu. All works well, here are details:

- SONOFF ZBDongle-P, ZBDongle-E, Dongle-LMG21/-PMG24/-M on Ubuntu
- Home Assistant Connect ZBT-1 (SkyConnect), ZBT-2 on both Windows and Ubuntu
- SMLight SLZB-06M, SLZB-MR1U on both Windows and Ubuntu

Glad to see this pull request has been merged.

@Nerivec @Koenkk Thank you for your excellent work.

PR #1682, opened 2026-03-04
Move manuSpecificPhilips and manuSpecificPhilips2 from ZH to ZHC

svhelge:
Ref https://github.com/Koenkk/zigbee-herdsman-converters/pull/11648

### DIFF
--- a/test/zcl.test.ts
+++ b/test/zcl.test.ts
@@ -1945,9 +1945,9 @@ describe("Zcl", () => {
     });
 
     it("Zcl utils get cluster with manufacturerCode", () => {
-        const cluster = Zcl.Utils.getCluster(0xfc00, 0x100b, MANU_SPE_CUSTOM_CLUSTERS);
-        expect(cluster.ID).toBe(0xfc00);
-        expect(cluster.name).toBe("manuSpecificPhilips");
+        const cluster = Zcl.Utils.getCluster(0xfc57, 0x1217, MANU_SPE_CUSTOM_CLUSTERS);
+        expect(cluster.ID).toBe(0xfc57);
+        expect(cluster.name).toBe("manuSpecificAmazonWWAH");
     });
 
     it("Zcl utils get cluster manufacturerCode", () => {


Nerivec:
Should change that failing test to use the AmazonWWAH one since it will stay in ZH and has a manuf code.

svhelge:
> Should change that failing test to use the AmazonWWAH one since it will stay in ZH and has a manuf code.

But the test that failed are related to clusters with the same ID (`0xfc00`). Shouldn't all these tests be removed? 


Maybe update `MANU_SPE_CUSTOM_CLUSTERS` :

```
const MANU_SPE_CUSTOM_CLUSTERS = {
    manuSpecificAmazonWWAH: {
        ID: 0xfc57,
        attributes: {},
        commands: {},
        commandsResponse: {},
    },
};

```
And have a test? "Zcl utils get cluster with/without manufacturerCode". 
The tests below could be removed, together with the 3 other manuSpecificAssaDoorLock tests? `
```
    it("Zcl utils get cluster without manufacturerCode", () => {
        const cluster = Zcl.Utils.getCluster(**0xfc00**, undefined, MANU_SPE_CUSTOM_CLUSTERS);
        expect(cluster.ID).toBe(0xfc00);
        expect(cluster.name).toBe("manuSpecificAssaDoorLock");
    });

    it("Zcl utils get cluster with manufacturerCode", () => {
        const cluster = Zcl.Utils.getCluster(_0xfc00_, 0x100b, MANU_SPE_CUSTOM_CLUSTERS);
        expect(cluster.ID).toBe(0xfc00);
        expect(cluster.name).toBe("manuSpecificPhilips");
    });

    it("Zcl utils get cluster manufacturerCode", () => {
        const cluster = Zcl.Utils.getCluster(0xfc00, 0x10f2, MANU_SPE_CUSTOM_CLUSTERS);
        expect(cluster.ID).toBe(0xfc00);
        expect(cluster.name).toBe("manuSpecificAssaDoorLock");
    });

    it("Zcl utils get cluster manufacturerCode wrong", () => {
        const cluster = Zcl.Utils.getCluster(0xfc00, 123, MANU_SPE_CUSTOM_CLUSTERS);
        expect(cluster.ID).toBe(0xfc00);
        expect(cluster.name).toBe("manuSpecificAssaDoorLock");
    });
```

Nerivec:
No they test the behavior, doesn't matter the custom cluster involved. Just need to replace with a currently viable one. No need to touch the Assa ones, just change `manuSpecificPhilips` with `manuSpecificAmazonWWAH` and change to the appropriate ID. Should make the test pass.

_We'll eventually remove `zcl.test.ts` entirely (coverage already done in `zspec` folder), but we want to do that after everything is migrated, so we can be sure we're not introducing some untested edge case (coverage check is never perfect)._

Koenkk:
Thanks!

Issue #1622, opened 2026-01-23
[ZCL] Move custom clusters to ZHC


Nerivec:
Tracking the list of remaining custom clusters to move to ZHC, if anyone wants to tackle some!
We'll mark priority ones (ID conflict) as we encounter them.

TODO:
- [ ] custom attributes and commands within standard clusters (too many to list here...)
- [ ] manuSpecificTuya
- [x] manuSpecificTuya2
- [x] manuSpecificTuya3
- [x] manuSpecificSinope
- [x] manuSpecificLumi
- [x] manuSpecificSmartThingsArrivalSensor
- [x] manuSpecificSiglisZigfred
- [x] owonClearMetering
- [x] zosungIRTransmit
- [x] zosungIRControl
- [x] manuSpecificProfalux1
- [x] manuSpecificPhilips*
  - priority, conflict with yet-to-be-moved `manuSpecificAssaDoorLock`
- [x] manuSpecificPhilips2
- [x] manuSpecificSamsungAccelerometer
  - priority, several manufacturer codes used for same cluster, must declare per-code for proper "default" behavior
- [x] manuSpecificClusterAduroSmart
- [x] manuSpecificOsram
- [x] manuSpecificLegrandDevices
- [x] manuSpecificLegrandDevices2
- [x] manuSpecificLegrandDevices3
- [x] wiserDeviceInfo
- [x] manuSpecificCentraliteHumidity
- [x] tradfriButton
- [x] schneiderSpecificPilotMode
- [x] elkoOccupancySettingClusterServer
- [x] elkoSwitchConfigurationClusterServer*
  - priority, conflict with yet-to-be-moved `manuSpecificSchneiderLightSwitchConfiguration`
- [x] manuSpecificSchneiderLightSwitchConfiguration*
  - priority, conflict with yet-to-be-moved `elkoSwitchConfigurationClusterServer`
- [x] manuSpecificSchneiderFanSwitchConfiguration
- [x] sprutVoc
  - priority, not using the proper "custom" range
- [x] sprutNoise
  - priority, not using the proper "custom" range
- [x] sprutIrBlaster
  - priority, not using the proper "custom" range
- [x] manuSpecificAssaDoorLock*
  - priority, conflict with yet-to-be-moved `manuSpecificPhilips` (and apparently some other yet-to-be-supported devices)
- [x] manuSpecificDoorman
- [x] manuSpecificAmazonWWAH (kept in ZH due to "global" usage)

[...]

PR #1714, opened 2026-03-22
fix: Move "Tradfri" commands used in ZHC to ZHC / ikea.ts

svhelge:


svhelge:
@Nerivec, test fails after removal of tradfri commands. Please advice.

andrei-lazarov:
> commands not used in ZHC

What do you mean? They are definitely used in ZHC: 
`ikeaArrowClick()` and `ikeaBilresaDouble()` functions, for the STYRBAR and BILRESA remotes.

The tradfri commands effectively cycle scenes on IKEA bulbs (I documented them in [STYRBAR](https://www.zigbee2mqtt.io/devices/E2001_E2002_E2313.html#scenes))

So note it's also useful to send them with Z2M: use them in automations, or at least test if your bulbs support them (to ensure compatibility with the remotes).

svhelge:
> > commands not used in ZHC
> 
> What do you mean? They are definitely used in ZHC: `ikeaArrowClick()` and `ikeaBilresaDouble()` functions, for the STYRBAR and BILRESA remotes.
> 
> The tradfri commands effectively cycle scenes on IKEA bulbs (I documented them in [STYRBAR](https://www.zigbee2mqtt.io/devices/E2001_E2002_E2313.html#scenes))
> 
> So note it's also useful to send them with Z2M: use them in automations, or at least test if your bulbs support them (to ensure compatibility with the remotes).

You're absolutely right.  The commands can not be removed, I believe the conclusion is then to move them to lib/ikea.ts in ZHC and assign the cluster to the devices with extend. 

Thiis PR was initiated by  #1622.

Nerivec:
> move them to lib/ikea.ts in ZHC and assign the cluster to the devices with extend

Indeed. Just like a custom cluster, except this is extending a standard cluster, so has to have the same cluster name & ID. There are many instances of this in ZHC already if examples needed.

svhelge:
@Nerivec , what about the test that failed? Replace `tradfriArrowSingle` with a different command?

Nerivec:
Looks like the two failures in `zcl.test.ts` are specific to tradfri, so can remove the two tests.
The one in `controller.test.ts` should be replaced indeed. From what I can tell, it is using a raw payload that would never be received (command not response), so, can refactor heavily:
https://github.com/svhelge/zigbee-herdsman/blob/c3c2c1cb811e6dbf9729696eed27a2c37fa71ea0/test/controller.test.ts#L2630-L2631
Replace these two lines with a simpler call to `Zcl.Frame.create(...)` that is easier to manipulate for expected result.
Something like this:
```ts
const frame = Zcl.Frame.create(Zcl.FrameType.SPECIFIC, Zcl.Direction.SERVER_TO_CLIENT, false, undefined, 29, "addRsp", "genScenes", {status: 0, groupId: 10, sceneId: 2}, {});
```
And change `expected` to match.

svhelge:
@Nerivec, Tests updated, but ran into coverage issue. 
```
ERROR: Coverage for lines (99.99%) does not meet global threshold (100%)
ERROR: Coverage for statements (99.99%) does not meet global threshold (100%)
ERROR: Coverage for branches (99.98%) does not meet global threshold (100%)
```

Nerivec:
Looks like that awkward controller test was hiding a coverage issue.

andrei-lazarov:
This will also remove the option to execute this command, correct?
Do I need to send a raw payload instead?

image

Nerivec:
It moves all customs to devices that actually support them (how it should be). It will show in dev console if the device supports it. This avoids nasty conflict issues with current implementation.

andrei-lazarov:
But how do you know if the bulb supports receiving it? 
Currently we only know that the remotes send it.

In that case we must do `addIkeaGenScenesCluster()` for all IKEA devices, not just the remotes?

Nerivec:
That's done during new device support implementation, same as custom clusters & co, requires manual testing from the implementer.

andrei-lazarov:
So the answer is yes, we should add it to all IKEA lights in zhc. 
Since basically all the lights are compatible with the remotes.

Nerivec:
I don't know these devices, but this only needs adding to a device converter _if the device actually implements and uses that attribute/cmd (talks about it with Z2M)_.

andrei-lazarov:
Ok, I'll do a part 2 for zhc after this is merged, to implement these commands and expose them to the users

Nerivec:
Unless I'm mistaken, from what I see in the ZHC PR, it should already be covered?
This stuff is sent _from remotes_ but doesn't have any impact on Z2M side since the behavior is device/group-handled. So this is mostly just to quiet the "not supported" you'd otherwise get in Z2M when the coordinator does group passthrough. That's why it actually _isn't used_ by Z2M, per svhelge's original assessment.

Though I'll let Koenkk confirm, he knows converters better. :grin:

andrei-lazarov:
Command sent from remote
- Used by ZHC. It receives _TradfriArrowSingle_ and translates it to `arrow_left_click` as an action
- **Covered in PR**

Command sent from Z2M (emulate the remote)
- Not used by ZHC
- Currently only possible to send from dev console (my screenshot above) -> I asked if this PR  would remove this option
- **I will add the possibility to send these commands to IKEA lights from their Exposes tab,** to prevent losing this feature. Something like this: ikea_scene_next, ikea_scene_prev, ikea_scene_cycle, ikea_scene_cycle_inverted, ikea_scene_cycle_stop

Nerivec:
I'm not sure that is desirable. My guess is, there are likely plenty of reasons this could always-fail for some devices (remote required, no longer in group, never had support...), hence not something to expose to users unless a device specifically, and always will work with it (also is the matter of it being group not device-specific). It's not intended to work that way, so, have to be sure, else we'll just get a flood of "doesn't work" issues, because, well, it doesn't. :sweat_smile: 

Note: we can add back "by ID" in the frontend's dev console if necessary so it can be used to send any attr/cmd/cluster by way of ID for lower level access (without having to use `raw` action).

andrei-lazarov:
It is pretty robust.
Whenever the bulb receives the command, it adds itself to the required group and creates the scenes if there aren't any.
I can do `if (fw_version > x)` to make sure it's only shown when supported. And I'm around to answer the issues 🙂

Koenkk:
LGTM, thanks!

GSzabados:
> Command sent from remote
> 
> * Used by ZHC. It receives _TradfriArrowSingle_ and translates it to `arrow_left_click` as an action
> * **Covered in PR**

Please excuse me for resurrecting this PR conversation, but those devices do far more than just _Click_. They should provide _Hold_ and _Release_ as well. And as far as I can tell, from trying to use any of my IKEA Tradfri E1524/E1810 devices, the  arrow_left_hold, arrow_left_release, arrow_right_hold, arrow_right_release actions are currently completely lost. I could see them covered in the removed code, but I cannot see any in the new code.

PR #1727, opened 2026-03-29
Some Touchlink changes

andrei-lazarov:
We wrote in the [Touchlink doc](https://www.zigbee2mqtt.io/guide/usage/touchlink.html) that ember does not receive scanning responses, _but other operations are not affected_.

However, they are affected.. because **identify and reset include scanning the channel and waiting for a response**. So they don't work either.

If I remove the scan, identify and reset are sent, but my devices show no effect.
**I made the operation continue, even when there's no scan response**. Are there drawbacks to this?

After that, ember still gave an error. It was trying to restore the channel while it was still scanning.
**I made it wait for the transmission to complete before restoring,** as the comments suggested 🙂 

**I also set identify time to 5 seconds.**

Now identify and reset work as expected in my testing, with Hue and Tuya bulbs.
What do you think @Nerivec @Koenkk?

Exceptions:
TRADFRI bulbs do factory reset and leave their network, but they don't enter pairing mode...
My KAJPLATS bulb ignores identify time (always does 15s) and completely ignores the reset..

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -2165,7 +2165,9 @@ export class EmberAdapter extends Adapter {
                 throw new Error(`~x~> [ZCL TOUCHLINK to=${ieeeAddress}] Failed to send with status=${SLStatus[status]}.`);
             }
 
-            // NOTE: can use ezspRawTransmitCompleteHandler if needed here
+            await new Promise<void>((resolve) => {
+                this.ezsp.once("rawTransmitComplete", resolve);
+            });
         });
     }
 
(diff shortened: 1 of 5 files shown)


Nerivec:
About the changes on the touchlink specs, would have to re-check the spec, I'm not that familiar with it.
I wouldn't trust tests on neither Hue nor Tuya though, Hue because they heavily customized Touchlink (not sure what's affected in the end), and Tuya, well....

About the changes in ember driver, I'm not sure what kind of timing is involved, but as-is, it runs the risk of race condition, which could end up in a forever hang. Should at least use `startWaitingForEvent` (instead of promise+once) from the waitress so we time out eventually.
Also, I have not kept track of changes for InterPAN (been a lot since 7.4), I'm not sure what would trigger, so, not sure that code would work for all versions.

I'd much prefer finding the bug/lack in the firmware rather than piling on workarounds in Z2M though, but this requires digging (I think it's around the filtering, and being in NCP mode), although it's not very high priority, doesn't seem to be used much, never got any feedback on it except for that original thread from way back... Have to prioritize time :sweat_smile: I'll try to send a message to Silabs, see if we get somewhere.

PR #1729, opened 2026-03-30
feat: emberAdapter: add SEND_MULTICASTS_TO_SLEEPY_ADDRESS stack config.

adriweb:
Note: several commercial Zigbee stacks/coordinators use this option.

P.S. This happens to be a "solution" to https://github.com/koenkk/zigbee2mqtt/issues/30247.

Tested successfully on a SONOFF Dongle-M coordinator.

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -172,6 +172,8 @@ type StackConfig = {
     TRANSIENT_KEY_TIMEOUT_S: number;
     /**@see Ezsp.ezspSetRadioIeee802154CcaMode */
     CCA_MODE?: keyof typeof IEEE802154CcaMode;
+    /** (Default: undefined) @see EzspConfigId.SEND_MULTICASTS_TO_SLEEPY_ADDRESS */
+    SEND_MULTICASTS_TO_SLEEPY_ADDRESS?: boolean;
 };
 
 /**
@@ -194,6 +196,7 @@ export const DEFAULT_STACK_CONFIG: Readonly<StackConfig> = {
     END_DEVICE_POLL_TIMEOUT: 8, // zigpc: 8
     TRANSIENT_KEY_TIMEOUT_S: 300, // zigpc: 65535
     CCA_MODE: undefined, // not set by default
+    SEND_MULTICASTS_TO_SLEEPY_ADDRESS: undefined, // not set by default
 };
 /** Default behavior is to disable app key requests */
 const ALLOW_APP_KEY_REQUESTS = false;
@@ -355,6 +358,11 @@ export class EmberAdapter extends Adapter {
             if (config.CCA_MODE && IEEE802154CcaMode[config.CCA_MODE] === undefined) {
                 config.CCA_MODE = undefined;
                 logger.error("[STACK CONFIG] Invalid CCA_MODE, ignoring.", NS);
+            }
+
+            if (config.SEND_MULTICASTS_TO_SLEEPY_ADDRESS != null && typeof config.SEND_MULTICASTS_TO_SLEEPY_ADDRESS !== "boolean") {
+                config.SEND_MULTICASTS_TO_SLEEPY_ADDRESS = undefined;
+                logger.error("[STACK CONFIG] Invalid SEND_MULTICASTS_TO_SLEEPY_ADDRESS, ignoring.", NS);
             }
 
             logger.info(`Using stack config ${JSON.stringify(config)}.`, NS);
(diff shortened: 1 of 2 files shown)


Nerivec:
Per Silabs source:
> This can be changed, but doing so is not ZigBee Pro Compliant and is possibly NOT interoperable.

Also, this is hardcoded in a lot of converters (since this is spec compliance), even with this, the error in the linked issue would still be thrown.
So, while this bypass would be effective at the stack level (to unknown effect), Z2M would still reject. Would need a much larger refactor, which at first glance due to the unknown effects, is probably not a good idea (e.g. could end up with plenty of routers just dropping these messages...). Probably best kept as a "done in custom firmware build" for tinkerers.

adriweb:
Yep, I noted the non-compliance thing in the commit I have locally for the zigbee2mqtt doc if this ends up being merged.

Anyway, I mean, I've been testing it with zigbee2mqtt with several brands of routers and end devices, so far so good.
And for instance both Somfy and Screen Innovations are doing this so it must not be so broken 😅

It seemed fine to me as a custom optional user option only toggleable in a json config file. Do you know of specific issues that this would trigger on the z2m side so I could look into it?

Nerivec:
Every converter in ZHC with a hardcoded group reject would not work, even with this, since it rejects long before it reaches the adapter. The whole codebase is treating this as impossible, so, bits that won't work are going to be spread all over.

It has the potential to break entire networks (takes just one router discarding these messages to make a mess), so, very tricky in my opinion. Can't rely on what a brand is doing to judge if it's safe, way too many are doing custom stuff, as long as it works with _their hub_...

Anyhow, without compatibility overhaul on Z2M side, it wouldn't work. We can't even be sure if it is properly wired "everywhere" in the Silabs stack, since that's a custom that would always be disabled in internal testing. :hot_face:

adriweb:
Oh, well, yes it would definitely be much more manageable if this was done some other more standard-friendly way, I'm just not aware of anything else now.
Even without that, it could also be maybe some kind of group-scoped property so users could adjust it accordingly when then _know_ the devices in the group are actually listening often-enough / aren't really _sleepy_ devices like sensors (I've seen some poll control stuff that are very weird, out there anyway... probably not passing any kind of certification lol).

In fact, I wonder if this stack option is toggleable on the fly (i.e. what if we try to enable it right before sending out a command to such a special flagged group, then disable it back afterwards), I haven't checked...

Note for context: several people (me included) tried to group window-covering devices (some of them battery-powered, so those are definitely not routers), and we see that the battery ones don't get group commands (or multicast in general, as expected) but they are still kind of always listening anyway (very frequent poll) since they are controllable individually just fine. With this option turned on, it does work fine for group too. Anyway, I'm sure you understood where this came from.

Anyway, is there any hope of having this merged if we insist that "this is an option that might at best not work, at worst break stuff, and it shouldn't be used unless you know what you're doing and assume all responsibility yourself for whatever happens" or something like that? Or should the PR just be closed for now (until a zigbee2mqtt overhaul to better handle the consequences of that - in which case, we should probably know a bit more about what that involves)?

adriweb:
Any word from a(nother?) maintainer, or should we just close it now since it's going to expire anyway?
Thanks!

andrei-lazarov:
I understand it's not safe for main. In this case I'd like to see it added as an extension (if possible)
https://github.com/Koenkk/zigbee2mqtt-user-extensions/

Nerivec:
As mentioned before, it's not just about it being not safe. It's that it won't work in a lot of cases, even with this. This is only stack-level (the extent of support being unknown, since untested/not recommended), but Z2M is also respecting the spec (save for device workarounds), and plenty of places will reject even without going to the adapter. It would only work for the subset that isn't actually properly validating (which hopefully, they aren't too many... despite this).

adriweb:
Yep, this doesn't change anything for normal commands sent by any frontend or normal Z2m high-level MQTT commands. It only makes it possible for [raw `{base}/bridge/request/action` mqtt topic messages](https://www.zigbee2mqtt.io/guide/usage/mqtt_topics_and_messages.html#action-raw) to work if some client sends broadcast commands with something like this:

```js
{
    action: 'raw',
    params: {
        network_address: broadcastAddress, // 0xFFFF for instance for all devices, even if sleepy
        dst_endpoint: 0xFF, // broadcast endpoint
        src_endpoint: 1,
        cluster_key: rawCommand.cluster_key,
        disable_response: true,
        zcl: {
            frame_type: 1,
            direction: 0,
            disable_default_response: true,
            command_key: rawCommand.command_key,
            payload: rawCommand.payload,
        }
    }
}
```
You'd have to construct your `rawCommand` as expected there.

( that's what I added in my app anyway, use it similarly at your own risks :) )

PR #1743, opened 2026-04-11
fix: Move Tuya specific attributes in lightingColorCtrl cluster from ZH to ZHC

svhelge:
Added "custom cluster" for Tuya command test in controller.test.ts.
Ref: https://github.com/Koenkk/zigbee-herdsman-converters/pull/11916


Koenkk:
Thanks!

PR #1751, opened 2026-04-26
feat: metrics instrumentation via typed EventEmitter

tomwilkie:
## Summary

- Replaces the injectable `Metrics` interface / `setMetrics` / `noopMetrics` pattern with a typed `EventEmitter<MetricsEventMap>` singleton exported as `metrics`
- Instruments send methods (`sendZclFrameToEndpoint`, `sendZdo`, `sendZclFrameToGroup`, `sendZclFrameToAll`) directly in the controller via a shared `instrumentSend` helper, with timing and success/failure status
- Adds metrics emission for adapter retries, ZDO responses, ZCL payload receives, and request queue length/duration
- Enqueue timestamp (`enqueuedAt`) stored directly on `Request` alongside `expires`, replacing a parallel `WeakMap` on `RequestQueue`
- Consumers subscribe via `metrics.on("adapterSendZclUnicast", (data) => ...)` etc. — metrics are always emitted unconditionally

## Test plan

- [ ] `npm test` — all 1744 tests pass
- [ ] Subscribe to `metrics.on(...)` events and verify they fire on adapter sends

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -6,6 +6,7 @@ import equals from "fast-deep-equal/es6";
 import type {Backup} from "../../../models";
 import {BackupUtils, Queue, wait} from "../../../utils";
 import {logger} from "../../../utils/logger";
+import {MetricType, metrics} from "../../../utils/metrics";
 import * as ZSpec from "../../../zspec";
 import type {Eui64, ExtendedPanId, NodeId, PanId} from "../../../zspec/tstypes";
 import * as Zcl from "../../../zspec/zcl";
@@ -1750,21 +1751,7 @@ export class EmberAdapter extends Adapter {
     //---- ZDO
 
     // queued, non-InterPAN
-    public async sendZdo(
-        ieeeAddress: string,
-        networkAddress: number,
-        clusterId: Zdo.ClusterId,
-        payload: Buffer,
-        disableResponse: true,
-    ): Promise<void>;
-    public async sendZdo<K extends keyof ZdoTypes.RequestToResponseMap>(
-        ieeeAddress: string,
-        networkAddress: number,
-        clusterId: K,
-        payload: Buffer,
-        disableResponse: false,
-    ): Promise<ZdoTypes.RequestToResponseMap[K]>;
-    public async sendZdo<K extends keyof ZdoTypes.RequestToResponseMap>(
+    protected async sendZdoImpl<K extends keyof ZdoTypes.RequestToResponseMap>(
         ieeeAddress: string,
         networkAddress: number,
         clusterId: K,
@@ -1930,7 +1917,7 @@ export class EmberAdapter extends Adapter {
     //---- ZCL
 
     // queued, non-InterPAN
-    public async sendZclFrameToEndpoint(
+    protected async sendZclFrameToEndpointImpl(
         ieeeAddr: string,
         networkAddress: number,
         endpoint: number,
@@ -2020,6 +2007,13 @@ export class EmberAdapter extends Adapter {
                 }
 
                 if (status === SLStatus.ZIGBEE_MAX_MESSAGE_LIMIT_REACHED || status === SLStatus.BUSY) {
+                    /* v8 ignore next 6 */
+                    metrics.emit("metric", {
+                        type: MetricType.AdapterRetry,
+                        adapterType: "ember",
+                        ieeeAddr,
+                        reason: SLStatus[status] ?? String(status),
+                    });
                     await wait(QUEUE_BUSY_DEFER_MSEC);
                 } else if (status === SLStatus.NETWORK_DOWN) {
                     await wait(QUEUE_NETWORK_DOWN_DEFER_MSEC);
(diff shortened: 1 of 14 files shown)


Nerivec:
This appears very targeted for Prometheus use (it is also not a no-op).
If we are to implement metrics in ZH, I think it should be something generic that any dependency-user (be it Z2M, iobroker, etc.) can use with the usual ZH events API (e.g. Z2M health system can make use of it, same as the rest).

We should also be very careful about possible performance degradation (be it small but widespread, or very large). Since this is "extra", it should have a very minimal impact overall.

tomwilkie:
Hi @nerivec! Thank you for your feedback - I mostly just opened this PR to see if you were open to this and get feedback on the direction to go.

> This appears very targeted for Prometheus use 

I tried to make it as non-prometheus specific as I could - e.g. not encoding counter, gauge or histogram specifics, but instead specific events to measure.  I'll admit I'm only familiar with Prometheus and OpenTelemetry, but if you have any other suggestions please let me know.

> it is also not a no-op

I tried to make the calling sites as simple as possible (e.g. not having each one check if the metrics received is null but instead of having a no-op receiver) - but I could change that?  WDYT?

I could also change the instrumentation of the adapter so that it only gets added if there is a metrics callback?

> If we are to implement metrics in ZH, I think it should be something generic that any dependency-user (be it Z2M, iobroker, etc.) can use with the usual ZH events API (e.g. Z2M health system can make use of it, same as the rest).

100%.  By "the usual ZH events API" do you mean e.g. making the Metrics interface follow the same pattern as ControllerEventMap and emit?

src/adapter/metricsAdapter.ts
Nerivec, line 1: Should remove this. Calling metrics on-site (controller/device/endpoint levels) would provide better details anyway.

src/controller/controller.ts
```diff
@@ -35,6 +37,7 @@ interface Options {
     databaseBackupPath: string;
     backupPath: string;
     adapter: AdapterTsType.AdapterOptions;
+    enableMetrics?: boolean;
```
Nerivec: Should remove this as well.
_Note: was incorrect in any case, since would only be for adapter wrapping._

src/utils/metrics.ts
```diff
@@ -0,0 +1,70 @@
...
+    adapterSendZclGroup: [data: AdapterSendZclGroupPayload];
+    adapterSendZclBroadcast: [data: AdapterSendZclBroadcastPayload];
+    adapterRetry: [data: AdapterRetryPayload];
+    adapterReceiveZclPayload: [data: AdapterReceiveZclPayloadPayload];
+    adapterReceiveZdoResponse: [data: AdapterReceiveZdoResponsePayload];
+    requestQueueLength: [data: RequestQueueLengthPayload];
+    requestQueueDuration: [data: RequestQueueDurationPayload];
+}
+
+export const metrics = new EventEmitter<MetricsEventMap>();
```
Nerivec, line 76: Use of this in e.g. Z2M would require to `import {metrics} from "zigbee-herdsman";`, i.e. entirely separate from Controller (main entrypoint). I think that's fine for this purpose though.
Note: need to be sure we can easily add a mock in Z2M tests (similar to Controller), no matter the design we end up with.

src/utils/metrics.ts
```diff
@@ -0,0 +1,70 @@
...
+}
+
+export interface RequestQueueDurationPayload {
+    ieeeAddr: string;
+    endpointId: number;
+    outcome: "sent" | "expired";
+    durationSeconds: number;
+}
+
+export interface MetricsEventMap {
```
Nerivec, line 72: I'm not sure about this one.
Is it better to have one event, with a discriminator `type` in the payload?
Current would require a lot of `.on("...", ...)`, not sure that's ideal for metrics (same "end logic").

src/adapter/ember/adapter/emberAdapter.ts
Nerivec, line 1: Should also identify available adapter's protocol-specific metrics.
I'm not sure if any other than `ember` currently have any, but currently we have a [3rd party tool](https://nerivec.github.io/z2m-ember-helper/) to parse this from logs. Would be much better integrated with metrics directly.

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -2007,8 +2008,11 @@ export class EmberAdapter extends Adapter {
                 }
 
                 if (status === SLStatus.ZIGBEE_MAX_MESSAGE_LIMIT_REACHED || status === SLStatus.BUSY) {
+                    /* v8 ignore next */
+                    metrics.emit("adapterRetry", {adapterType: "ember", ieeeAddr, reason: SLStatus[status] ?? String(status)});
                     await wait(QUEUE_BUSY_DEFER_MSEC);
                 } else if (status === SLStatus.NETWORK_DOWN) {
+                    metrics.emit("adapterRetry", {adapterType: "ember", ieeeAddr, reason: "NETWORK_DOWN"});
```
Nerivec: These are not "retry".

tomwilkie:
Thanks for the review feedback! I'll work through it and integrate the changes into https://github.com/Koenkk/zigbee2mqtt/pull/31645.  I've got a build of the HA add on for testing in my local setup too, and need to build some dashboards etc..  so probably a few more days before I've got it working end to end.

src/utils/metrics.ts
tomwilkie: I've refactored into a single event now, with enum sum-type.

tomwilkie:
I've got this all working end to end now, used in a PR in z2m (https://github.com/Koenkk/zigbee2mqtt/pull/31645), built a add on container, deployed at home and built a nice set of dashboards in Grafana: 

image

I think it could be ready for proper review? I've done a manual review myself.  The thing to flag is none of the send metrics were working - intercepting send calls at the controller didn't work as the seems to be 20+ sites that call the adapater itself.  I've tried to make the most minimal change, but putting the instrumentation in the base call, but would be keen on you thoughts.  Thanks!

PR #1752, opened 2026-04-27
fix: moveToLevel min and max

andrei-lazarov:
- closes https://github.com/Koenkk/zigbee-herdsman-converters/pull/12017
- related https://github.com/Koenkk/zigbee-herdsman-converters/pull/11244

@Nerivec I think `min: 0` is also needed


Nerivec:
`min: 0` can be removed, it has no effect.

Nerivec:
`min: 1, ` in the comment can be left though, that one is correct.

Koenkk:
Thanks!

PR #1756, opened 2026-05-14
fix: ignore default response during OTA blocks sending

Nerivec:
Should cover the undesired stalling behavior observed in a few OTA logs (mostly Hue?) like in https://github.com/Koenkk/zigbee2mqtt/issues/32008
Devices are expected to continue requesting blocks as appropriate after sending a non-zero default response.

cc: @andrei-lazarov


Koenkk:
thanks!

PR #1761, opened 2026-05-17
feat: add support for Zigbee Direct cluster

### LINKED ISSUE #1760: Support for Zigbee Direct Configuration cluster (`0x003d`)
Just installed a Ubisys S1-R and noticed that the Bluetooth interface for [Zigbee Direct](https://github.com/user-attachments/files/24265978/20-27688-041-zigbee_direct_spec_v1.1.pdf) is still active after commissioning. 🤨
Turns out, there's a cluster for that!

The **Zigbee Direct Configuration Cluster** `0x003d`, in the case of Ubisys on endpoint `232`, can be used to configure that mode.

IMHO, we should at least implement the attributes and commands used by Ubisys so Bluetooth can be disabled.
Reducing attack surface as well as 2.4 GHz airtime is always a good thing.

cc. @Nerivec

Nerivec:
Fixes https://github.com/Koenkk/zigbee-herdsman/issues/1760

### DIFF
--- a/src/zspec/zcl/definition/cluster.ts
+++ b/src/zspec/zcl/definition/cluster.ts
@@ -2964,6 +2964,43 @@ export const Clusters: Readonly<Record<ClusterName, Cluster>> = {
         commands: {},
         commandsResponse: {},
     },
+    zigbeeDirectConfig: {
+        // NOTE: only accepts unicast
+        name: "zigbeeDirectConfig",
+        ID: 0x003d,
+        attributes: {
+            /** 0: disabled, 1: enabled, other: reserved */
+            interfaceState: {name: "interfaceState", ID: 0x0000, type: DataType.BITMAP8, max: 0xff, required: true},
+            anonymousJoinTimeout: {name: "anonymousJoinTimeout", ID: 0x0001, type: DataType.UINT24, max: 0x100000, required: true},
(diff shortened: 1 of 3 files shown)


src/zspec/zcl/definition/cluster.ts
```diff
@@ -2964,6 +2964,43 @@ export const Clusters: Readonly<Record<ClusterName, Cluster>> = {
         commands: {},
         commandsResponse: {},
     },
+    zigbeeDirectConfig: {
+        // NOTE: only accepts unicast
+        name: "zigbeeDirectConfig",
```
burmistrzak: We're not always using _the exact cluster name_ from spec, right?

I'd say we go with either the full name `*Configuration` or standardize a short form like `*Config` or `*Cfg` across ZH.

burmistrzak (commented):
Ayyy, you took my PR! 🫶

src/zspec/zcl/definition/cluster.ts
Nerivec: Fixed, spec name is always better indeed.
Lots of old names in spec, but renaming clusters is a pain, breaks too many things down the line.

burmistrzak (approved):
LGTM

PR #1762, opened 2026-05-18
feat: Enable APS layer encryption for Zigbee Direct cluster operations

burmistrzak:
As discussed in https://github.com/Koenkk/zigbee-herdsman/issues/1760, we need to enable APS layer encryption to read/write the **Zigbee Direct Configuration** cluster.

This can be done by checking the destination cluster ID and setting the respective APS frame options.

- [x] Proof-of-Concept (`ember`)
- [x] Adapter: `ember`
- [x] Tests

### DIFF
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -1146,11 +1146,11 @@ export class ZStackAdapter extends Adapter {
         const transactionID = this.nextTransactionID();
         const response = this.znp.waitFor(Type.AREQ, Subsystem.AF, "dataConfirm", undefined, transactionID, undefined, timeout);
 
-        let frameOptions = 0;
+        let options = 0;
 
         // Zigbee Direct cluster, enable APS layer encryption
         if (clusterID === Zcl.Clusters.zigbeeDirectConfiguration.ID) {
-            frameOptions |= Constants.AF.options.EN_SECURITY;
+            options |= Constants.AF.options.EN_SECURITY;
         }
 
         await this.znp.request(
@@ -1162,7 +1162,7 @@ export class ZStackAdapter extends Adapter {
                 srcendpoint: sourceEndpoint,
                 clusterid: clusterID,
                 transid: transactionID,
-                options: frameOptions,
+                options,
                 radius: radius,
                 len: data.length,
                 data: data,
--- a/src/adapter/z-stack/adapter/zStackAdapter.ts
+++ b/src/adapter/z-stack/adapter/zStackAdapter.ts
@@ -1146,6 +1146,13 @@ export class ZStackAdapter extends Adapter {
         const transactionID = this.nextTransactionID();
         const response = this.znp.waitFor(Type.AREQ, Subsystem.AF, "dataConfirm", undefined, transactionID, undefined, timeout);
 
+        let frameOptions = 0;
+
+        // Zigbee Direct cluster, enable APS layer encryption
+        if (clusterID === Zcl.Clusters.zigbeeDirectConfiguration.ID) {
+            frameOptions |= Constants.AF.options.EN_SECURITY;
+        }
+
         await this.znp.request(
             Subsystem.AF,
             "dataRequest",
@@ -1155,7 +1162,7 @@ export class ZStackAdapter extends Adapter {
                 srcendpoint: sourceEndpoint,
                 clusterid: clusterID,
                 transid: transactionID,
-                options: 0,
+                options: frameOptions,
                 radius: radius,
                 len: data.length,
                 data: data,
--- a/src/adapter/zboss/driver.ts
+++ b/src/adapter/zboss/driver.ts
@@ -8,6 +8,7 @@ import {Waitress} from "../../utils";
 import {AsyncMutex} from "../../utils/async-mutex";
 import {logger} from "../../utils/logger";
 import type * as ZSpec from "../../zspec";
+import * as Zcl from "../../zspec/zcl";
 import * as Zdo from "../../zspec/zdo";
 import type {TsType} from "..";
 import {ZDO_REQ_CLUSTER_ID_TO_ZBOSS_COMMAND_ID} from "./commands";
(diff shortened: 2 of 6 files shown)


burmistrzak:
@Nerivec How should we go about this? Especially writing a somewhat useful test...

I'm also only familiar with `ember` and `z-stack` adapters. 😬

Nerivec:
I'm not sure this stuff is available in all adapters. Determining that would be the first order of business I guess. Need to dig into the types for all adapters, see if there are APS options somewhere, and if so, if one is about encryption/security :sweat_smile:

burmistrzak:
> I'm not sure this stuff is available in all adapters.

True, Silabs is quite ahead.

> Determining that would be the first order of business I guess. Need to dig into the types for all adapters, see if there are APS options somewhere, and if so, if one is about encryption/security 😅

I looked through the `z-stack` code and found https://github.com/Koenkk/zigbee-herdsman/blob/a62a4d5d7f039c3b4de4bdaf2130a8d0bfa09a01/src/adapter/z-stack/adapter/zStackAdapter.ts#L1149-L1164

According to TI docs, security _can_ be enabled:

> Transmit options bit mask according to the following defines
from AF.h: bit 1: sets ‘Wildcard Profile ID’; bit 4: turns on/off
‘APS ACK’; bit 5 sets ‘discover route’; bit 6 sets ‘APS
security’; bit 7 sets ‘skip routing’.

Question is, why is this attribute hardcoded to `0`?

**Edit:** Looks like @Koenkk had a similar question back then 😄
https://github.com/Koenkk/zigbee-herdsman/blob/a62a4d5d7f039c3b4de4bdaf2130a8d0bfa09a01/src/adapter/z-stack/adapter/zStackAdapter.ts#L1206

**Edit 2:** Bitmask values defined here: https://github.com/Koenkk/zigbee-herdsman/blob/a62a4d5d7f039c3b4de4bdaf2130a8d0bfa09a01/src/adapter/z-stack/constants/af.ts#L16-L23

These transmit options are all disabled and also a bit different from what's available on `ember`.
So setting `options: Constants.AF.options.EN_SECURITY` _should_ do the trick.

Nerivec:
https://github.com/Koenkk/zigbee-herdsman/blob/fb2b2fe411f7c4c5038244971ba3927d73d9cd85/src/adapter/deconz/adapter/deconzAdapter.ts#L379

https://github.com/dresden-elektronik/deconz-lib/blob/3b058797089bddeaac020c038a5cea6fc9862701/deconz/aps.h#L92-L98

cc: @manup

---

https://github.com/Koenkk/zigbee-herdsman/blob/fb2b2fe411f7c4c5038244971ba3927d73d9cd85/src/adapter/zboss/driver.ts#L320

Based on NCP serial protocol v1.5
```
TX options bitmap
0x01 = Security enabled transmission
0x02 = Obsolete
0x04 = Acknowledged transmission
0x08 = Fragmentation permitted
0x10 = Include extended nonce in APS security frame.
0x20 = Force mesh route discovery for this request.
0x40 = Send route record for this request.
```

cc: @kirovilya 

---

As long as it's always scoped to the ZD cluster, we should be fine merging this after all these + zstack are handled. Since ZD cluster support is new anyway, and the potential breakage surface is minimal.

Nerivec:
Need to implement for all four before ready. We have the data, so, should mostly be replicating.

burmistrzak:
> Need to implement for all four before ready. We have the data, so, should mostly be replicating.

Do you want me to implement them as well?
Z-Stack _should_ be fine, but I have zero experience with the other two. 😬

Nerivec:
Should just be adding the same logic as ember in the two places I linked above (and where you linked for zstack), with the values described (can just hardcode the number for now, if you don't find the appropriate const/enum).
ZD is unicast only, so, that should be all.
zstack will need a test for coverage, the others don't have any.

burmistrzak:
@Nerivec Would you mind taking a look? 🤞

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -1146,6 +1146,13 @@ export class ZStackAdapter extends Adapter {
         const transactionID = this.nextTransactionID();
         const response = this.znp.waitFor(Type.AREQ, Subsystem.AF, "dataConfirm", undefined, transactionID, undefined, timeout);
 
+        let frameOptions = 0;
```
Nerivec:
```suggestion
        let options = 0;
```

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -1146,6 +1146,13 @@ export class ZStackAdapter extends Adapter {
...
+
+        // Zigbee Direct cluster, enable APS layer encryption
+        if (clusterID === Zcl.Clusters.zigbeeDirectConfiguration.ID) {
+            frameOptions |= Constants.AF.options.EN_SECURITY;
```
Nerivec:
```suggestion
            options |= Constants.AF.options.EN_SECURITY;
```

src/adapter/z-stack/adapter/zStackAdapter.ts
```diff
@@ -1155,7 +1162,7 @@ export class ZStackAdapter extends Adapter {
                 srcendpoint: sourceEndpoint,
                 clusterid: clusterID,
                 transid: transactionID,
-                options: 0,
+                options: frameOptions,
```
Nerivec:
```suggestion
                options,
```

src/adapter/zboss/driver.ts
```diff
@@ -317,7 +324,7 @@ export class ZBOSSDriver extends EventEmitter {
             srcEndpoint: srcEp,
             radius: 3,
             dstAddrMode: 3, // ADDRESS MODE ieee
-            txOptions: 2, // ROUTE DISCOVERY
+            txOptions: txOptions,
```
Nerivec:
```suggestion
            txOptions,
```

Nerivec (approved):
Looks good.
I say we merge after next release.

burmistrzak:
@Nerivec Thx! But what about ZoH? 👀

Nerivec:
It's not exposed in ZoH for now. Would need to refactor that side. Haven't had time to finish the big Z4 refactor though, need that one first. Very edge-case feature, it can wait a bit. We'll probably do the support for this on ZoH side, since this is spec.

burmistrzak:
> Looks good. I say we merge after next release.

Good thing that I've already disabled ZD with my locally patched ZH. 😜

Koenkk:
Thanks!

PR #1770, opened 2026-06-02
fix: possible OTA mismatch due to no tsn involved

Nerivec:
Might be the solve for the weird behavior in https://github.com/Koenkk/zigbee2mqtt/issues/32168
Due to having no TSN matching in OTA, the `defaultRsp` could possibly end up matching against the wrong waiter.

We don't have access to the fully decoded frame at that point, but I think the simple byte index retrieval is enough for this scenario (`defaultRsp` format is simple and static). I don't think it's worth the much bigger refactor that would otherwise be needed.

I don't have any Hue device with the triggering behavior, so, can't confirm it _actually_ is the fix for above issue.
@dwieland @SakuraiSatoru if you are able to confirm, that would be great.


Nerivec:
Closing, not the right approach, cmdId in defaultRsp doesn't follow the expected waiter pattern (looks like copilot identified this as well :sweat_smile:)...

PR #1771, opened 2026-06-02
fix: defaultRsp matching for OTA

Nerivec:
Supersedes https://github.com/Koenkk/zigbee-herdsman/pull/1770
Attempts to solve https://github.com/Koenkk/zigbee2mqtt/issues/32168

Due to having no TSN matching in OTA, the `defaultRsp` could possibly end up matching against the wrong waiter.
OTA also uses an off-pattern matching (send cmd rsp, wait on next req from device) to current waiter implementation.
We force the use of a separate arg in that context, to prevent issues.

@Koenkk it's a bit "involved" for such a simple use-case... but only other approach I can think of would be to completely disable `defaultRsp` matching for OTA cluster or when no TSN present (would effectively result in same bypass as https://github.com/Koenkk/zigbee-herdsman/pull/1756, just may be a bit "rougher")
```ts
(matcher.transactionSequenceNumber !== undefined && header.frameControl.frameType === Zcl.FrameType.GLOBAL && header.commandIdentifier === 0x0b))
```
Thoughts?

### DIFF
--- a/src/adapter/adapter.ts
+++ b/src/adapter/adapter.ts
@@ -18,6 +18,7 @@ export interface ClusterWaitressMatcher {
     clusterId: number;
     endpoint?: number;
     commandId?: number;
+    defaultRspCommandId?: number;
     transactionSequenceNumber?: number;
 }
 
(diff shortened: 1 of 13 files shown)


Koenkk:
LGTM, thanks!

PR #1774, opened 2026-06-05
fix: ember: skip backup devices without a link key in FORM_BACKUP

evertide:
device.linkKey is optional in the open coordinator backup format. The FORM_BACKUP path dereferenced it with a non-null assertion, crashing with "TypeError: Cannot read properties of undefined (reading 'key')" when the coordinator_backup.json (e.g. produced by zigpy/bellows during a cross-adapter migration) contained a device without a link key. Filter those devices out before building the APS link-key list.

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -957,16 +957,22 @@ export class EmberAdapter extends Adapter {
                 logger.info("[INIT TC] Forming from backup.", NS);
                 // `backup` valid in this `action` path (not detected by TS)
                 /* v8 ignore start */
+                const keyList: LinkKeyBackupData[] = [];
+
                 // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                const keyList: LinkKeyBackupData[] = backup!.devices.map((device) => ({
-                    deviceEui64: ZSpec.Utils.eui64BEBufferToHex(device.ieeeAddress),
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    key: {contents: device.linkKey!.key},
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    outgoingFrameCounter: device.linkKey!.txCounter,
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    incomingFrameCounter: device.linkKey!.rxCounter,
-                }));
+                for (const device of backup!.devices) {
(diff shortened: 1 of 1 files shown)


src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -958,15 +958,18 @@ export class EmberAdapter extends Adapter {
...
-                const keyList: LinkKeyBackupData[] = backup!.devices.map((device) => ({
-                    deviceEui64: ZSpec.Utils.eui64BEBufferToHex(device.ieeeAddress),
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    key: {contents: device.linkKey!.key},
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    outgoingFrameCounter: device.linkKey!.txCounter,
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    incomingFrameCounter: device.linkKey!.rxCounter,
-                }));
+                const keyList: LinkKeyBackupData[] = backup!.devices
```
Nerivec: Should use a single for loop instead of chained calls to have proper typing and remove all the assertion suppressions.

evertide: Thanks! Switched to a single for loop — proper narrowing, removed the non-null assertions and suppressions.

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -957,16 +957,22 @@ export class EmberAdapter extends Adapter {
...
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    key: {contents: device.linkKey!.key},
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    outgoingFrameCounter: device.linkKey!.txCounter,
-                    // biome-ignore lint/style/noNonNullAssertion: ignored using `--suppress`
-                    incomingFrameCounter: device.linkKey!.rxCounter,
-                }));
+                for (const device of backup!.devices) {
+                    // link_key is optional in the open coordinator backup format, skip devices without one
+                    if (device.linkKey === undefined) {
```
Nerivec: Let's go with this instead, since this is coming from JSON, just in case...
```suggestion
                    if (device.linkKey == null) {
```

evertide: Good point — applied, thanks.

Nerivec:
Thanks!

PR #1776, opened 2026-06-06
fix: zoh: only close port (if needed) on init port failure

Nerivec:
Fixes https://github.com/Nerivec/zigbee-on-host/discussions/124

CC: @chris-1243 @bitcoin1192


chris-1243:
Just tried the latest-dev and `zoh.save` is not partially erased anymore

PR #1778, opened 2026-06-06
feat: Support for Frient install codes

burmistrzak:
Devices from Frient also use pipe-separated install codes, but they add a leading and trailing pipe to the mix... 😒
This PR updates the existing regex to also accept this otherwise identical type of install code.


burmistrzak:
An alternative solution could be to first strip off the leading and trailing pipe `|` and then pass it along to an unchanged regex... 🤔

Nerivec:
> An alternative solution could be to first strip off the leading and trailing pipe

The regex basically already does this with the capture groups. I think it's fine like this.

Koenkk:
Thanks!

PR #1792, opened 2026-06-21
Support for Control4 keypads, tested with 6 button C4-KP6-Z

ArcadeMachinist:
[c4-kp6-z.js](https://github.com/user-attachments/files/29182549/c4-kp6-z.js)
Attempt #2 to add my Control4 keypad thing to upstream.
Requires Silabs dongle with EZSP firmware.
I'm unaware if it can be made to work with any other dongles, the basic requirement is to be able to use non-compliant ZCL frames to/from custom endpoint.

Yes, buttons and LEDs work.

PoC: https://youtu.be/eDS3Qqf_STI

Let me know if any improvements required.

Converter: attached. 
[c4-kp6-z.js](https://github.com/user-attachments/files/29182554/c4-kp6-z.js)

### DIFF
--- a/src/adapter/ezsp/adapter/ezspAdapter.ts
+++ b/src/adapter/ezsp/adapter/ezspAdapter.ts
@@ -55,7 +55,9 @@ export class EZSPAdapter extends Adapter {
             frame.apsFrame.profileId === ZSpec.HA_PROFILE_ID ||
             frame.apsFrame.profileId === 0xffff ||
             // Shelly custom Clusters require a special profile ID
-            frame.apsFrame.profileId === ZSpec.CUSTOM_SHELLY_PROFILE_ID
+            frame.apsFrame.profileId === ZSpec.CUSTOM_SHELLY_PROFILE_ID ||
+            // Control4 devices use a custom profile ID
+            frame.apsFrame.profileId === ZSpec.CONTROL4_PROFILE_ID
         ) {
             const payload: ZclPayload = {
                 clusterID: frame.apsFrame.clusterId,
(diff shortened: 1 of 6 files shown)


Nerivec:
This is far too much customs for ZH. Lots of this should be in ZHC.

ArcadeMachinist:
> This is far too much customs for ZH. Lots of this should be in ZHC.

I do not think there is way to do ZCL handling from an external converter.
At least I do not see how.
The whole ordeal with C4 is that they decided to use propietary ZCL frames that are not actual ZCL.
APS frames are not ZCL, but ASCII with Control4 custom protocol.

If there is anyway in ZH to parse and construct APS from an external converter - I would be happy to move that.

Nerivec:
That would seem to be beyond the scope of Z2M entirely. If the manuf doesn't want support in the Open Source community, it's their choice I guess. Piling on every custom protocol into Z2M is a non-starter.

Koenkk:
This requires indeed to much an invasive changes in the core (zigbee-herdsman), the [quirks lookup](https://github.com/Koenkk/zigbee-herdsman/pull/1792/changes#diff-6d00af5fc389c40b8c4dfc390d80e3eca3946d331dff84afdc1632cb5568bca2R901) changes would be fine by me but the others are not.

ArcadeMachinist:
Ok, I'll try to keep my fork updated for the C4 stuff.
I just like the design and LED feedback in these particular keypads.

If in the future there would be a way to extend profile handling or even to construct APS on supported sticks from the external converter - that would be cool, but until, I'm ok upkeeping the C4 fork.

I do not plan to add any more weird devices, but all C4-KPx-Z series keypads should work, any number of buttons.

bharat:
@ArcadeMachinist following up on this thread, because your closing wish (a way to extend profile handling, and to construct APS frames from a converter) is the path we are now trying to walk upstream in small pieces. Your converter and this PR did the protocol groundwork that everything below builds on.

Koenkk/zigbee-herdsman#1837 takes just the profile-acceptance slice, following the Shelly precedent and covering the ember adapter. Next is an interview-quirk PR, per Koenkk's note above that those changes would be fine. If both land, we plan to propose a generic raw-APS send so converter-side C4 support needs no fork at all.

Since your fork has been carrying an equivalent profile patch, a look at #1837 from you would be valuable, and if it lands it should let your C4-KP6-Z fleet run closer to stock too. Review comments welcome in either direction.

PR #1793, opened 2026-06-22
fix: possible race condition with OTA abort

Nerivec:
Prevent race between running abort sequence, and throwing from abort, which possibly resulted in OTA being aborted without ZH actually sending the abort response.
CC: @andrei-lazarov 

Also fix vitest package version mismatch.


andrei-lazarov:
Thanks for the quick fix! <3
I tested a few times, doesn't seem to happen anymore

Koenkk:
Thanks!

PR #1797, opened 2026-06-25
Support for Sonoff Dongle-PP10 auto-discovery.

CubeZ2mDeveloper:
Support for automatic discovery of SONOFF Dongle-PP10 in Z2M. After setup, I tested it in Z2M v2.12.0, and it was successfully detected. The dongle is currently not on the market, with official sales scheduled for late July.


Koenkk:
thanks!

PR #1799, opened 2026-06-27
fix(zstack): increase BDB commissioning timeout from 60s to 100s

### LINKED ISSUE #1767: zStack adapter: BDB commissioning timeout is hardcoded at 60s, too short for some coordinators after NVRAM erase
## Summary

The zStack adapter's BDB commissioning timeout is hardcoded at 60 seconds (`manager.ts` line 623). After an NVRAM erase on a CC2652-based coordinator (SLZB-06), commissioning consistently takes ~76 seconds, causing Z2M to fail with "network commissioning timed out" on every startup attempt. There is no configuration option to override this timeout.

## Environment

- **Coordinator:** SLZB-06 (CC2652P, network-connected via TCP, PoE)
- **Firmware tested:** zStack `20250321`, `20260310`, `20260311` — all exhibit the same behavior
- **Z2M:** Latest via HAOS add-on
- **Host:** Home Assistant Yellow running HAOS 17.3, HA Core 2026.5.4

## Steps to Reproduce

1. Perform an NVRAM erase on a CC2652-based coordinator (e.g. via ZHA's "create network" flow, or any method that clears NVRAM)
2. Configure Z2M to use the coordinator with zStack adapter
3. Start Z2M

**Expected:** Z2M starts successfully after commissioning completes.

**Actual:** Z2M fails with "network commissioning timed out" because commissioning takes ~76 seconds but the timeout is 60 seconds.

## Analysis

The timeout is at [`src/adapter/z-stack/adapter/manager.ts` line 623](https://github.com/Koenkk/zigbee-herdsman/blob/master/src/adapter/z-stack/adapter/manager.ts#L623) — a hardcoded 60,000ms wait for `WaitressTimeoutError`.

The coordinator radio is fully functional. Using ZHA to send the same NVRAM erase + commission sequence to the same hardware succeeds — it just takes longer than 60s. The commissioning was observed completing at ~76 seconds consistently across multiple firmware versions.

Once the timeout fires, Z2M exits. On next startup it attempts commissioning again from scratch, hitting the same timeout — creating an unrecoverable loop.

## Workaround

No practical workaround exists within Z2M configuration:
- The timeout is not exposed as a config option
- Container filesystem patches don't persist on HAOS
- Removing the antenna to speed up commissioning (fewer neighbours to discover) did not reduce the time below 60s

The only workaround was migrating to a different adapter chipset (EmberZNet/EFR32), which shouldn't be necessary given the radio hardware works fine.

## Suggestion

Either:
1. Make the BDB commissioning timeout configurable (e.g. `advanced.commissioning_timeout` in Z2M config, passed through to herdsman)
2. Increase the default to something more generous (120s would cover observed cases with margin)
3. Both — increase the default and make it configurable

---

— on behalf of @Flameeyes

bobtheskull-flameeyes:
As requested in #1767 — changes the BDB commissioning timeout from 60s to 100s.

The 60-second timeout is too short for some CC2652-based coordinators (observed on SLZB-06) after an NVRAM erase, where commissioning consistently takes ~76s across multiple firmware versions. This causes an unrecoverable startup loop.

100s matches the value suggested in the issue discussion and provides margin for slower hardware while remaining reasonable for a one-time commissioning operation.

### Change

 line 372:  →  (the  timeout in ).

The separate 60s timeout in  (line 287) is left unchanged — that path is for regular coordinator startup, not initial commissioning, and has not been reported as problematic.

Fixes #1767

### DIFF
--- a/src/adapter/z-stack/adapter/manager.ts
+++ b/src/adapter/z-stack/adapter/manager.ts
@@ -370,7 +370,7 @@ export class ZnpAdapterManager {
             await this.znp.request(Subsystem.APP_CNF, "bdbSetChannel", {isPrimary: 0x0, channel: 0x0});
 
             /* perform bdb commissioning */
-            const started = this.znp.waitFor(UnpiConstants.Type.AREQ, Subsystem.ZDO, "stateChangeInd", undefined, undefined, 9, 60000);
+            const started = this.znp.waitFor(UnpiConstants.Type.AREQ, Subsystem.ZDO, "stateChangeInd", undefined, undefined, 9, 100000);
             await this.znp.request(Subsystem.APP_CNF, "bdbStartCommissioning", {mode: 0x04});
             try {
                 await started.start().promise;


Koenkk:
Thanks!

PR #1804, opened 2026-06-30
fix: greenpower attr parsing

### LINKED ISSUE #1803: BUG - buffaloZcl readGpdFrame parser
https://github.com/Koenkk/zigbee-herdsman/blob/88f1de08dd2091a09693a6e7d30ab4dfdc9dbc67/src/zspec/zcl/buffaloZcl.ts#L620

Hey,

Just noticed that there is potentially a bug here. The 'recordsLength' is the total in bytes and not the amount of records there are. 

Image

"total size in octets" (octets = bytes)

I just verified through Wireshark. 

Image

Nerivec:
Fixes https://github.com/Koenkk/zigbee-herdsman/issues/1803

CC: @joutei


Koenkk:
thanks!

PR #1816, opened 2026-07-21
fix: optimize attribute reads during interview logic

Nerivec:
Related to https://github.com/Koenkk/zigbee2mqtt/issues/32583

Change attr reads from "last valid wins" to "first valid wins", since per spec, basic attrs we read during interview are expected to match across clusters (device-wide, not endpoint-specific). And for a device that doesn't respect that, neither first nor last can be relied upon to be "the valid one" in any case.
Should remove a few read calls for multi-endpoint devices.

Also fix a "valid" detection problem (i.e. `0` can be valid).


Koenkk:
Thanks!

PR #1818, opened 2026-07-21
Clear #genBasic field when removing device from a database

Xsarius:
When device is removed cache for given IEEE is persisted. This can lead to eg. malformed fields saved.

image

The issue is that even after deleting the device and re-joining network, cached values are still used over valid/latest/actual values. Manual re-interview helps, but this can be fixed with a simple clearing out all fields when deleting a device. Then it works just fine:

image

Resolves https://github.com/Koenkk/zigbee-herdsman/pull/1818

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -1253,8 +1253,10 @@ export class Device extends Entity<ControllerEventMap> {
         Device.devices.delete(this.ieeeAddr);
 
         // Clear all data in case device joins again
-        // Green power devices are never interviewed, keep existing interview state.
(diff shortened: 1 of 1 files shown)


Nerivec:
The screenshots show a before/after firmware change on a custom device; not a corruption scenario. This requires a manual re-interview, which clears the cache (or a remove/restart/re-pair).

Clearing the cache more aggressively in normal operation has far-reaching consequences re devices that mishandle joins and other quirks (too many to even evaluate the potential for breaks from a change like this...).
As-is, it also introduces uneven cache clearing (clears only _some_ data).
It also penalizes normal operations by forcing to retrieve all these attributes again; that is unnecessary for all non-custom cases.

PR #1819, opened 2026-07-22
feat: allow to remove device from cache on request

Nerivec:
Completely remove the device from lookup instead of keeping a "deleted" reference.
Once the caller (e.g. Z2M) is done processing, ref is dropped, and the device will be GCed; next join will be as if fresh out of the box.

Replaces https://github.com/Koenkk/zigbee-herdsman/pull/1818


Koenkk:
Thanks!

PR #1821, opened 2026-07-23
fix(deconz): bind onPortClose so TCP socket close does not crash

### LINKED ISSUE #1820: deconz: unbound onPortClose crashes on TCP socket close (this.emitStateEvent is not a function)
## What happened

Running the deconz adapter over a TCP serial bridge (`serial.port: tcp://...`), any
close of the TCP connection crashes the whole process with an uncaught TypeError
instead of emitting `DriverEvent.Disconnected`:

```
TypeError: this.emitStateEvent is not a function
    at Socket.onPortClose (/app/node_modules/.pnpm/zigbee-herdsman@10.6.1/node_modules/zigbee-herdsman/src/adapter/deconz/driver/driver.ts:828:14)
    at Object.onceWrapper (node:events:631:26)
    at Socket.emit (node:events:509:28)
    at TCP.<anonymous> (node:net:351:12)
```

Note the receiver in the first frame: `Socket.onPortClose` — the handler runs with
`this` bound to the `net.Socket`, not the driver.

## Root cause

In `src/adapter/deconz/driver/driver.ts` (master @ 43a9e5ae5b), the serial path
registers the close handler bound:

https://github.com/Koenkk/zigbee-herdsman/blob/43a9e5ae5b/src/adapter/deconz/driver/driver.ts#L859

```ts
this.serialPort.on("close", this.onPortClose.bind(this));
```

but the TCP path registers it unbound:

https://github.com/Koenkk/zigbee-herdsman/blob/43a9e5ae5b/src/adapter/deconz/driver/driver.ts#L924

```ts
this.socketPort!.once("close", this.onPortClose);
```

so when the socket closes, `onPortClose` executes with `this` = the Socket and
`this.emitStateEvent(DriverEvent.Disconnected)` (L828) throws. The intended
`Disconnected` event never fires, the exception is uncaught, and the process
exits — Zigbee2MQTT's own watchdog/reconnect logic never gets a chance to run.

## How to reproduce

1. Expose a ConBee II via any serial-to-TCP bridge (ser2net, socat, etc.).
2. Zigbee2MQTT config: `serial.port: tcp://<host>:<port>`, `serial.adapter: deconz`.
3. Start Zigbee2MQTT (works fine), then restart the TCP bridge (or drop the
   connection any other way).
4. Zigbee2MQTT crashes with the TypeError above instead of reconnecting.

Reproduced 2/2 times; the serial (`/dev/tty*`) path handles the same event
correctly, as its handler is bound.

## Expected behavior

Socket close should emit `DriverEvent.Disconnected` (same as the serial path),
letting the adapter/application layer handle reconnection gracefully.

## Suggested fix

```diff
-            this.socketPort!.once("close", this.onPortClose);
+            this.socketPort!.once("close", this.onPortClose.bind(this));
```

(or make `onPortClose` an arrow function property, consistent with `onParsed`).

## Environment

- Zigbee2MQTT 2.12.1 (docker, `koenkk/zigbee2mqtt:latest`)
- zigbee-herdsman 10.6.1
- Adapter: ConBee II (deconz), via TCP serial bridge on macOS host
- Node: container default

trailro:
Fixes #1820

## Problem

The deconz driver registers `onPortClose` **bound** on the serial path but **unbound** on the TCP path:

```ts
this.serialPort.on("close", this.onPortClose.bind(this));   // serial: bound
this.socketPort!.once("close", this.onPortClose);            // TCP: unbound
```

So on the TCP path the handler runs with `this` bound to the `net.Socket`, and `this.emitStateEvent(DriverEvent.Disconnected)` throws:

```
TypeError: this.emitStateEvent is not a function
    at Socket.onPortClose (.../deconz/driver/driver.ts:828:14)
    at Object.onceWrapper (node:events:631:26)
    at Socket.emit (node:events:509:28)
    at TCP.<anonymous> (node:net:351:12)
```

The exception is uncaught and kills the process, so `DriverEvent.Disconnected` is never emitted and the reconnect path never runs. Any socket close (bridge restart, network blip) takes Zigbee2MQTT down instead of reconnecting.

## Fix

Bind `onPortClose` in the constructor, right next to the existing `onParsed` bind, so every registration site gets the bound handler and the serial site no longer needs its own `.bind(this)`. This follows the pattern already used in this file and prevents the same mistake at any future registration site.

## Verification

- `pnpm run build` — passes
- `pnpm run check` — 251 files, 0 errors
- `pnpm test` — 1755 passed, 1 skipped
- **Functional test against the actual failure**, ConBee II over a TCP serial bridge (`serial.port: tcp://...`, Zigbee2MQTT 2.12.1 / zigbee-herdsman 10.6.1):

  *Before* — closing the socket crashed the process with the TypeError above (reproduced 2/2).

  *After* — the driver emits `Disconnected` and reconnects on its own; the process stays alive and Zigbee2MQTT recovers without a restart:

  ```
  Failed to send ZCL request (2) Error: Cleanup in state: WaitToReconnect
  Failed to send ZCL request (3) Error: Cleanup in state: Connecting
  ```

### DIFF
--- a/src/adapter/deconz/driver/driver.ts
+++ b/src/adapter/deconz/driver/driver.ts
@@ -188,6 +188,7 @@ class Driver extends events.EventEmitter {
         }, 100);
 
         this.onParsed = this.onParsed.bind(this);
+        this.onPortClose = this.onPortClose.bind(this);
         this.frameParserEvent.on("deviceStateUpdated", (data: number) => {
             this.checkDeviceStatus(data);
         });
@@ -856,7 +857,7 @@ class Driver extends events.EventEmitter {
                 this.writer.pipe(this.serialPort);
                 this.serialPort.pipe(this.parser);
                 this.parser.on("parsed", this.onParsed);
-                this.serialPort.on("close", this.onPortClose.bind(this));
+                this.serialPort.on("close", this.onPortClose);
                 this.serialPort.on("error", this.onPortError.bind(this));
             }
 


src/adapter/deconz/driver/driver.ts
```diff
@@ -188,6 +188,7 @@ class Driver extends events.EventEmitter {
         }, 100);
 
         this.onParsed = this.onParsed.bind(this);
+        this.onPortClose = this.onPortClose.bind(this);
```
Nerivec, line 191: Should remove these and bind on-site; less error-prone (easy to spot if an event is registered without `bind` when it's on same line).

PR #1823, opened 2026-07-28
fix(ember): answer unsolicited OTA queryNextImageRequest with NO_IMAGE_AVAILABLE

bueste:
### What
Adds a dedicated `genOta` case in `Device.onZclData()` that answers an unsolicited `queryNextImageRequest` with `NO_IMAGE_AVAILABLE`, guarded by a new `_pendingOtaCheck` flag.

### Why
When a device sends a `queryNextImageRequest` on its own initiative (i.e. no adapter-driven OTA check via `#notifyOta` is currently in progress), the request was previously left unanswered:

- `command.response` is defined for this command, so the generic default-response path (`sendDefaultResponse`) is skipped.
- There was no dedicated case for `genOta` in the `isSpecific` switch to send the actual `queryNextImageResponse`.

Devices that never receive a reply keep retrying on their own short interval indefinitely. Observed in practice as a query roughly every 3 seconds, indefinitely, from a mains-powered router (SONOFF MINI-ZBRBS) that was never manually queried for an update.

### How
`_pendingOtaCheck` is set to `true` for the duration of `#notifyOta` (the adapter-initiated check-for-update flow) and reset in a `finally` block. The new `genOta` case only answers unsolicited requests when this flag is `false`, so it never intercepts or interferes with a real, adapter-initiated OTA check that is already waiting for the same command via `#waitForOtaCommand`.

### Tested
- `tsc --noEmit` and `biome check` pass against the full project.
- Applied locally against a live network (ioBroker.zigbee) and confirmed the previously constant ~3s polling from the affected device stops after the first `NO_IMAGE_AVAILABLE` reply.
- Have not verified this doesn't affect adapters other than EmberZNet, since that's the only stack I could test against.


Nerivec:
This behavior is left up to the higher layer depending on how the application deals with OTA. Doing this in ZH would break that possibility.
Example in Z2M:
https://github.com/Koenkk/zigbee2mqtt/blob/6835cbf5a73116aa4bfa097f2ad0dfb972018997/lib/extension/otaUpdate.ts#L181-L188

PR #1825, opened 2026-07-28
feat: handle `special` in ZCL spec

### LINKED ISSUE #32336: Unable to set previous in Color temp startup

Nerivec:
Supersedes https://github.com/Koenkk/zigbee-herdsman/pull/1791
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/31452
Fixes https://github.com/Koenkk/zigbee2mqtt/issues/32336

Implements the first part of https://github.com/Koenkk/zigbee-herdsman/pull/1791#issuecomment-4762953354
Changed the type to be more perf-friendly on this side (wasn't used before, so technically not a breaking change); when used for display, can afford more expensive ops for value/label access.
Optimized call locations for proper behavior, but "cheapest checks first".

Will fix the main issue, which are a few attrs in barrier ctrl, lighting color ctrl and illuminance measurement that require this to actually access these values (outside allowed ranges).

Probably some are missing (ZAP is limited in clusters it defines - which is where these special values were added from), we'll have to add these as they come (or go through whole spec and identify special values :sweat_smile:).

---

The ZHC ties (see above comment), well, that's a much bigger job :grin: 
Would be good to have full exposes<>attributes auto-matching, but likely can't without breaking changes... and lots of time!


Koenkk:
Thanks!

PR #1830, opened 2026-07-30
fix: add support for some recent `EzspValueId`

Nerivec:
Not marking this as `feat` since it's just a couple unused-by-zh extras.


Koenkk:
thanks!

PR #1612, opened 2026-01-08
feat!: OTA refactor

Nerivec:
## Design changes

- https://github.com/Koenkk/zigbee-herdsman/issues/1608
- Use separate util class for OTA session
  - Use `AsyncGenerator` to handle block/page/end requests
  - Rework progress updates: initial will send estimate with chunks stats, and after 30sec, will start using calculation from current offset
- Use `check` function inside `update` for consistency (no duplicate logic)
- `check` and `update` return types have changed to adapt to new features
- If `check` is called without `queryNextImageRequest` payload (undefined), automatically notify the device to trigger it then reply `NO_IMAGE_AVAILABLE` (fallback to prevent device from requesting again)
- If `update` is called without `queryNextImageRequest` payload (undefined), automatically notify the device to trigger it, then reply as appropriate based on image availability (which will make the device start OTA update or not).
- Prevent running `update` while one is running
- Data settings handled per update trigger instead of globally
- Remove public `Endpoint.waitForCommand` API that was solely intended for OTA use, now handled internally
- Logging changed quite a bit to match rest of ZH logging / new features

Logic changes & co can be derived from list of below tests.

## CI Tests

- Checks
  - returns no image when index is empty
  - performs notify flow when current payload is missing
  - matches OTA images using extra metadata overrides
  - matches OTA images using manufacturer name and hardware ranges
  - matches OTA images using extra manufacturer metadata when device manufacturer name mismatches
  - handles failures when waiting for OTA notify response
  - fails when remote index fetching fails
  - merges override index (prio) with URL index
  - auto-adds meta for override index local firmware when unspecified
  - allows use of remote override index even if URL index fetching fails
  - allows use of local override index even if default URL index fetching fails
  - allows use of local override index even if custom URL index fetching fails
  - overrides lumi file version with meta and reports upgrade availability
  - writes scenes group before checking OTA for PP-WHT-US
- Updates
  - applies an upgrade image end-to-end (using several sample images)
  - applies a downgrade image end-to-end
  - considers an upgrade successful even if no device announce
  - handles receiving non-value for data size
  - handles out-of-order block offsets
  - handles negative misaligned block offset - resends portion already sent
  - handles positive misaligned block offset - skips portion
  - handles page requests
  - applies manufacturer-specific timeout
  - applies manufacturer-specific and imageType-specific timeout
  - applies manufacturer-specific data size
  - handles URL as source URL
  - handles filesystem as source URL
  - applies fallback timeouts when manufacturer is unknown
[...]

PR #1835, opened 2026-08-02
fix(ota): send relative upgrade time in upgradeEndResponse

midlan:
## The problem

Since the OTA refactor (#1612), `upgradeEndResponse` is sent with `currentTime = <real ZCL-UTC now>` and `upgradeTime = currentTime + 1`. The line even carries a TODO suspecting trouble:

> `upgradeTime: currentTime + 1, // TODO: could this tiny offset be a problem for some stacks?`

It is. Per ZCL spec the client should wait `UpgradeTime - CurrentTime`, but some stacks instead compare `UpgradeTime` against **their own UTC clock**. The NXP JN51xx OTA client (JN-SW-4170 SDK, used by many DIY firmwares such as hellozigbee) only schedules relatively when `currentTime == 0`; otherwise it arms an absolute compare. A device without a Time cluster boots its ZCL UTC clock at 0, so the upgrade gets scheduled **~26 years into the future**:

- the transfer and device-side CRC succeed,
- Z2M logs "Finished update" and waits for a device announce that never comes,
- the device keeps running the old image and (on JN51xx) sits in `COUNT_DOWN` state ignoring all further OTA commands.

Reproduced and verified on an Aqara QBKG11LM running hellozigbee: with current master the device never applies (traced in the JN-SW-4170 client sources, `OTA_ClientUpgradeManager.c`, `E_CLD_OTA_COMMAND_UPGRADE_END_RESPONSE` handling — the `currentTime != 0` branch stores `upgradeTime` as an absolute local-clock deadline); with this patch it applies ~1 s after the grant.

## The fix

Send the spec's relative form: `currentTime: 0` ("server does not support UTC time") with `upgradeTime: 1`. Spec-compliant clients compute the identical 1-second delta, and clock-comparing stacks apply immediately. This also matches the pre-#1612 behaviour, where these devices updated fine.

Added a payload assertion to the end-to-end success test so the contract is locked in; all 91 tests in `test/device-ota.test.ts` pass.

https://claude.ai/code/session_01TxTanjAztZW5G7fkHFaCjZ

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -1754,13 +1754,7 @@ export class Device extends Entity<ControllerEventMap> {
                         manufacturerCode: image.header.manufacturerCode,
                         imageType: image.header.imageType,
                         fileVersion: image.header.fileVersion,
-                        // currentTime=0 signals "no UTC clock, UpgradeTime is a relative delay" per ZCL spec 11.13.9.4.
-                        // Sending the real UTC time here breaks stacks that compare UpgradeTime against their own
-                        // (unsynced) clock instead of computing the delta: e.g. the NXP JN51xx OTA client only
-                        // schedules relatively when currentTime is 0, otherwise it arms an absolute compare against
-                        // a clock that starts at 0 on boot - the upgrade lands ~26 years out and the device never
-                        // reboots even though the transfer and CRC succeeded. Spec-compliant clients compute
-                        // UpgradeTime - CurrentTime, which is identical either way.
+                        // using 0 tells the device to use `upgradeTime` as offset (11.13.8.2.8), preventing issues with UTC Time support
                         currentTime: 0,
                         upgradeTime: 1,
                     },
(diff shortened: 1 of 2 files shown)


src/controller/model/device.ts
```diff
@@ -1747,17 +1747,22 @@ export class Device extends Entity<ControllerEventMap> {
...
                         fileVersion: image.header.fileVersion,
-                        currentTime,
-                        upgradeTime: currentTime + 1, // TODO: could this tiny offset be a problem for some stacks?
+                        // currentTime=0 signals "no UTC clock, UpgradeTime is a relative delay" per ZCL spec 11.13.9.4.
+                        // Sending the real UTC time here breaks stacks that compare UpgradeTime against their own
+                        // (unsynced) clock instead of computing the delta: e.g. the NXP JN51xx OTA client only
+                        // schedules relatively when currentTime is 0, otherwise it arms an absolute compare against
+                        // a clock that starts at 0 on boot - the upgrade lands ~26 years out and the device never
+                        // reboots even though the transfer and CRC succeeded. Spec-compliant clients compute
+                        // UpgradeTime - CurrentTime, which is identical either way.
```
Nerivec:
```suggestion
                        // using 0 tells the device to use `upgradeTime` as offset (11.13.8.2.8), preventing issues with UTC Time support
```

Koenkk:
Thanks!

PR #1838, opened 2026-08-09
fix: prevent crash on backup failure

Nerivec:
_Can't believe we haven't "caught" that one before_ :rofl: 

Technically won't make much difference, except for a cleaner stop, since if this happens, it's usually because the adapter crashed (so Z2M still ends up restarting).

Sidenote: something's off with vitest, even without the catch it's not detecting the uncaught promise? :confused: Probably time to upgrade everything to latest v4!


Koenkk:
Thanks!

PR #1840, opened 2026-08-10
feat: Interview Control4 devices via a manufacturerCode-keyed quirk

bharat:
Completes the interview for Control4 in-wall dimmers and keypads by keying an interview quirk on the node descriptor's manufacturerCode (`0xabcd`) and stubbing `manufacturerName` / `modelId` / `powerSource`. These devices never answer genBasic reads, so without this the interview always fails before a modelID is known and the existing modelID-keyed quirks can never match them. The stubbed modelId is what converters fingerprint the family on.

Includes a controller test backed by a mock device that reproduces the real interview failure shape: node descriptor succeeds, genBasic reads go unanswered, and the simple descriptor request for the proprietary endpoint times out.

In production this quirk has paired 31 devices across the C4-APD120 / C4-KD120 / C4-KC120277 families since early July, on an ember coordinator, with zero interview failures.

Builds on #1837 and takes the approach suggested in the #1792 review, where the quirks-lookup change was called out as acceptable.

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -886,6 +886,19 @@ export class Device extends Entity<ControllerEventMap> {
             return true;
         }
 
+        // Control4 in-wall dimmers/keypads do not answer genBasic reads at all, so the interview always
+        // fails before a modelID is known and the modelID-keyed quirks below can never match them. The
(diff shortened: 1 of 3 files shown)


Koenkk:
Thanks!

PR #1842, opened 2026-08-12
fix: add fingerprint for different zbt-2 batch

Nerivec:
Will fix mismatches when baudrate or rtscts was not explicitly specified (assuming user is using "common" firmware builds).


Koenkk:
Thanks!

PR #1843, opened 2026-08-12
feat: Add Endpoint.sendRaw for non-ZCL APS payloads

bharat:
Adds `Endpoint.sendRaw(clusterId, data, options)`, which sends a bare APS payload to the endpoint exactly as provided, with no ZCL framing. Some vendors carry proprietary protocols in APS payloads that are not ZCL frames (Control4 in-wall keypads carry ASCII text on custom profile `0xC25C`), and every existing send path serializes a ZCL header, so such payloads currently cannot be produced at all. The `profileId` option and the adapter plumbing for it already exist; the mandatory ZCL framing is the only gap this closes.

The payload rides the normal request pipeline (request queue, sendPolicy, recovery) via a minimal frame-shaped carrier whose `toBuffer()` returns the data verbatim. No response correlation is performed and `disableResponse` defaults to true accordingly: replies, if any, arrive through the normal incoming message path, which #1837 already admits for whitelisted custom profiles. If you would prefer a first-class raw-frame type or an adapter-level API over the carrier object, I am happy to rework it to whatever shape you want; this is the minimal version.

Tests assert the payload reaches the adapter byte-for-byte with the requested profileId and the disableResponse default, plus the error path.

In production this has driven 31 Control4 devices on an ember coordinator since early July: every LED command, config write, and sequence-matched configuration read goes through this path.

This completes the sequence started in #1837 and #1840, and is the capability ArcadeMachinist asked about when closing #1792 (constructing APS frames from a converter): with it, Control4 support can live entirely in converters with no forked core.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -452,6 +452,55 @@ export class Endpoint extends ZigbeeEntity {
     /*
      * Zigbee functions
      */
+
+    /**
+     * Send a raw, non-ZCL APS payload to this endpoint.
+     *
+     * Some vendors (e.g. Control4) carry proprietary protocols in bare APS payloads that are
+     * not ZCL frames. The regular command/write paths always serialize a ZCL header, so they
+     * cannot produce such payloads. This sends `data` verbatim on `clusterId`, typically
+     * combined with `options.profileId` to address a custom profile.
+     *
+     * No response correlation is performed: replies (if any) are emitted through the normal
+     * incoming message path for the converter to match. `options.disableResponse` defaults
+     * to true accordingly.
+     */
+    public async sendRaw(clusterId: number, data: Buffer, options?: Options): Promise<void> {
+        const optionsWithDefaults = this.getOptionsWithDefaults({disableResponse: true, ...options}, true, Zcl.Direction.CLIENT_TO_SERVER, undefined);
+        const transactionSequenceNumber = optionsWithDefaults.transactionSequenceNumber ?? zclTransactionSequenceNumber.next();
+        // A minimal frame-shaped carrier. The request pipeline and the adapters only read
+        // cluster.ID, command, header.frameControl and toBuffer(), and serialize the message
+        // with toBuffer(), so the payload goes out exactly as provided with no ZCL framing.
+        const frame = {
+            cluster: {ID: clusterId, name: `raw_0x${clusterId.toString(16)}`},
+            command: {ID: 0, name: "raw", parameters: []},
+            header: {
+                transactionSequenceNumber,
+                frameControl: {
+                    frameType: Zcl.FrameType.SPECIFIC,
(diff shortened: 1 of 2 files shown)


Nerivec:
That kind of custom protocol addition was denied introduction before. They just bring too much trouble from maintenance to proper support. That "rougher shape" might work for these devices, but what if next custom protocol needs a slightly different logic... another custom function..?
ZH should remain spec-following, with the exception of a few quirks.

bharat:
I read through the concerns in #1792, where vendor protocol logic ended up inside ZH, and am trying to thread the needle here: keep ZH clean while adding a small generic escape hatch.

The intent here is to NOT put vendor protocol logic inside ZH - quite the opposite. Sorry if I didn't frame it well. `sendRaw` contains no protocol knowledge at all: it sends the given bytes on the given cluster and profile, and that is the whole function.

If a future custom protocol needs different logic, that logic lives in its converter and doesn't require changes or special knowledge in ZH. In that sense, this is an attempt to support the devices behind #1792 and similar efforts in a ZH-friendly, spec-friendly way.

It is also the TX half of something ZH already does on RX: since #1418 and #1837 the adapters deliberately accept non-ZCL frames on whitelisted custom profiles and deliver them to consumers. Today those consumers can listen but cannot speak without forking core.

Happy to rework this in any way you'd like if you can suggest a better approach - including constraining it (for example refusing the standard HA/ZGP profile ids, or documenting it as an expert API with no support guarantees) if the worry is the support burden. I'm just hoping to make this accessible to folks on these non-standard platforms without them having to carry a fork, or putting more burden on the ZH team.

Nerivec:
> no protocol knowledge at all

That behavior is definitely specific to this device. That's the whole problem of custom protocols, each will have their own (even if it's just one bit that needs flipping). Not to mention, potential internal updates to custom protocols would require reworking every aspect to support that as well...

Then we'd also end up with issues in specific adapters, because one drops these frames, another cannot process them and crashes, etc..
And probably the whole waiter aspect fails as well...

A custom profile that follows the protocol is _very different_ from an entirely custom protocol.

bharat:
Thanks for taking the time to explain. I don't want to make your life more difficult! If I have to maintain a fork to keep this working, so be it.

That said... is it possible that there's a creative solution? I notice that the APS data service (APSDE-DATA.request) is itself part of the Zigbee spec, with ZCL as one payload format above it, and the adapter plumbing for profileId is already generic. Would an explicit, capability-gated APSDE-DATA.request primitive be acceptable?

Concretely that would be an adapter capability flag, implemented for ember first, with an unsupported error everywhere else so no adapter can crash on a frame it cannot handle. Response waiters would be documented as out of scope, and we'd refuse the standard HA and ZGP profile ids so it can't be used to bypass normal ZCL paths.

That would be exposing a spec layer boundary rather than accommodating any particular vendor... totally get it if you don't like that solution either. I'm just spitballing over here 😄

PR #1849, opened 2026-08-15
fix: Prevent duplicate configured reportings

pedrolamas:
Found while investigating Koenkk/zigbee2mqtt#32857, where a plug silently stopped reporting and the Reporting tab showed six entries for three attributes.

### The problem

`configureReporting` and `saveClusterAttributeReportConfig` used different rules to decide whether a stored configured reporting matched the one being written, so each could leave a second entry behind for the same attribute:

- `configureReporting` only replaced an existing entry when the stored `manufacturerCode` key was absent or held the exact same value, so an entry stored with another value was left in place.
- `saveClusterAttributeReportConfig` treated an undefined incoming manufacturer code as matching anything, but never matched a legacy entry stored without one when reading with a manufacturer code, and pushed a duplicate.

Neither handled the other's case, which is the mechanism that produces the duplicates.

### The change

Both paths now share `isSameConfiguredReporting`, where an undefined manufacturer code on either side matches any. `fromDatabaseRecord` additionally drops duplicates already stored by older versions, keeping the last stored entry, and normalizes `manufacturerCode: null` to `undefined` — older versions stored `null` rather than omitting the key, which otherwise prevents a stored entry from ever matching one written with an actual manufacturer code. Normalizing also keeps `null` from reaching `Zcl.Utils.getCluster`.

### Trade-off worth reviewing

The symmetric wildcard means a standard and a manufacturer-specific entry that share an attribute ID in the same cluster will collapse to one entry. Each of the two previous predicates already had that wildcard in one direction, so this is not a regression, but it is a deliberate choice and worth a second opinion.

"Keeping the last stored entry" is array position, which reflects creation order. An entry updated in place by `saveClusterAttributeReportConfig` keeps its original position, so in the rare case where an earlier-positioned duplicate holds fresher values the dedupe keeps the stale one.

### Testing

Three tests added, all of which fail against the current code. Verified against a live 34 device network, where 7 devices — smart plugs from two vendors — had duplicate entries cleaned up on startup, every one of them stored with `"manufacturerCode": null`.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -266,10 +266,12 @@ export class Endpoint extends ZigbeeEntity {
         }
         /* v8 ignore stop */
 
-        // Drop duplicates left by older versions, keeping the last stored entry @deprecated 3.0
+        // Older versions stored `null` manufacturer codes and duplicate entries, keep the last of each @deprecated 3.0
         const configuredReportings: ConfiguredReportingInternal[] = [];
 
-        for (const entry of (record.configuredReportings || []) as ConfiguredReportingInternal[]) {
+        for (const stored of record.configuredReportings || []) {
+            const entry: ConfiguredReportingInternal = {...stored, manufacturerCode: stored.manufacturerCode ?? undefined};
+
             const duplicateIdx = configuredReportings.findIndex((c) =>
                 Endpoint.isSameConfiguredReporting(c, entry.cluster, entry.attrId, entry.manufacturerCode),
             );
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -266,6 +266,24 @@ export class Endpoint extends ZigbeeEntity {
         }
         /* v8 ignore stop */
 
+        // Drop duplicates left by older versions, keeping the last stored entry @deprecated 3.0
+        const configuredReportings: ConfiguredReportingInternal[] = [];
+
+        for (const entry of (record.configuredReportings || []) as ConfiguredReportingInternal[]) {
+            const duplicateIdx = configuredReportings.findIndex((c) =>
+                Endpoint.isSameConfiguredReporting(c, entry.cluster, entry.attrId, entry.manufacturerCode),
+            );
+
+            if (duplicateIdx === -1) {
+                configuredReportings.push(entry);
+            } else {
+                const dropped = configuredReportings[duplicateIdx];
+
+                logger.debug(() => `Dropping duplicate configured reporting for ${deviceIeeeAddress}/${record.epId}: ${JSON.stringify(dropped)}`, NS);
+                configuredReportings[duplicateIdx] = entry;
+            }
+        }
+
         return new Endpoint(
             record.epId,
             record.profId,
@@ -276,7 +294,7 @@ export class Endpoint extends ZigbeeEntity {
             deviceIeeeAddress,
             record.clusters,
             record.binds || [],
-            record.configuredReportings || [],
+            configuredReportings,
             record.meta || {},
         );
     }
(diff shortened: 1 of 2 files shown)


Nerivec:
Haven't had time to look into that bug, but if memory serves, there's a simpler, cleaner fix, which is to properly record the manuf in the entries (and probably tweak a couple predicates to match).
And I also think this needs proper propagation in Z2M codebase.

pedrolamas:
Thanks you @Nerivec, I think you're right and I'd like to check one detail before rewriting it!

I'm reading "properly record the manuf in the entries" as: store the code derived from the *attribute definition* rather than the call-wide request code, then compare strictly instead of wildcarding on `undefined`. Concretely:

1. Resolve per entry using the logic that already exists in `ensureManufacturerCodeIsUniqueAndGet` (`attribute.manufacturerCode ?? options.manufacturerCode ?? cluster.manufacturerCode`), and write it from both `configureReporting` and `saveClusterAttributeReportConfig`. That also fixes something I noticed on the way: the in-place update branch of `saveClusterAttributeReportConfig` never refreshes `manufacturerCode`, so an entry stored without one stays that way forever.
2. Make `isSameConfiguredReporting` compare strictly, resolving a stored `undefined` through the same attribute lookup at match time. Both call sites already have the cluster and `customClusters` in hand.
3. Leave `fromDatabaseRecord` doing only `null → undefined` plus an exact dedupe; codes get backfilled on first write.

The detail I want to confirm is step 3, because I don't think I can backfill at load. `fromDatabaseRecord` gets only the raw record, `Device.fromDatabaseEntry` builds endpoints before `new Device(...)`, and more decisively `customClusters` are never persisted — they're registered by zigbee-herdsman-converters after load. So anything needing definitions has to stay lazy, for the same reason `configuredReportings` resolves clusters in the getter. Is lazy migration what you had in mind, or were you picturing something that keeps it eager?

One consequence worth naming: strict matching makes the test I added at `controller.test.ts:4823` obsolete — it seeds an entry stored under a manufacturer code for an attribute that has none, which can't arise once the code is derived rather than taken from the request. The pre-existing legacy test at `:4799` still passes, since `0xfc57` resolves to the Amazon code on both sides.

On propagation into Z2M: `ConfiguredReporting` and `Bind` aren't exported today — Z2M re-declares `Bind` by hand and reaches `ConfiguredReporting` structurally through the getter's return type. Happy to export them properly as part of this.

Data point for whichever way we go: on a 34-device network, 7 devices had duplicates, every one stored as `"manufacturerCode": null` against a sibling with the key absent. Both normalise to `undefined`, so an exact dedupe still catches that case.

exSnake:
Another real-world data point, in case it helps settle the lazy-vs-eager question.

My install: Z2M 2.10.1 (herdsman 10.0.7), device TS011F_plug_1 (`_TZ3000_w0qqde0g`), endpoint 1. The Reporting tab shows eight rows, four duplicated pairs across `haElectricalMeasurement` (rmsVoltage, rmsCurrent, activePower) and `seMetering` (currentSummDelivered). In database.db every stale entry carries `"manufacturerCode": null` (the device was paired before #1146 changed the default to `undefined`); every sibling written later by the frontend omits the key entirely. Standard clusters only, no manufacturer-specific attributes anywhere.

The delete-removes-the-wrong-row symptom reproduces exactly as described: terminating the `null` row removes the sibling instead (the filter in `configureReporting` matches the key-absent entry through `!("manufacturerCode" in c)` but `null === undefined` never matches), and editing it just resurrects the pair.

Relevant to the open question: for legacy data like mine, `null → undefined` normalization plus an exact dedupe at load cleans everything without needing attribute definitions, so a lazy backfill of resolved codes would lose nothing here.

PR #1850, opened 2026-08-16
feat: Add request to verify device bindings

pedrolamas:
Follows up on your review of Koenkk/zigbee2mqtt#32858, where you noted that the binding verification logic belongs here rather than in Zigbee2MQTT. This is that logic.

Context is Koenkk/zigbee2mqtt#32857: a plug silently stopped reporting while Zigbee2MQTT showed it as healthy throughout — availability `online`, `last_seen` current, LQI normal. Commands, reads and availability pings all kept working, because none of them depend on the binding table. Reading the device's own binding table answered it directly: `bindingTableEntries: 0`, with all three coordinator bindings gone on a device that was still fully reachable.

### The problem

`Device.bindingTable()` already reconciles the cached bindings with the device's own table, but it calls `saveBindings()` per endpoint and returns only the raw ZDO table. The previous state is discarded, so a device that dropped its bindings is indistinguishable from a correctly bound one. Any caller wanting to know has to snapshot `endpoint.binds` before and after and diff it by hand — which is exactly what Zigbee2MQTT was doing.

### The change

`Device.verifyBindings()` performs the same read and reconciliation, returning per endpoint:

- `bindings` — what the device holds now
- `added` — bindings the device holds that were not cached
- `removed` — cached bindings the device does not hold
- `unreportableClusters` — clusters with a configured reporting but no binding to the coordinator, so the device cannot report them

`unreportableClusters` is derived from current state rather than a diff, which matters: reading the binding table reconciles the cache, so `removed` only reports a fault the first time it is seen. `unreportableClusters` keeps reporting it until the device is reconfigured. Both are useful — `removed` is the only signal for bindings to a target other than the coordinator, such as a switch bound to a group.

`bindingTable()` keeps its signature and behaviour; both now share a private paginated read. `saveBindings()` returns its own `{added, removed}` rather than `void`, so the diff is computed where the replacement happens.

`Bind` and `ConfiguredReporting` are now exported. Both were already de-facto public API — Zigbee2MQTT re-declares `Bind` by hand and reaches `ConfiguredReporting` structurally through the `configuredReportings` getter's return type.

### Note on the coordinator

`unreportableClusters` treats "bound to the coordinator" as the target device being `Device.byType("Coordinator")[0]`, rather than a specific coordinator endpoint. Reports only reach the controller through a binding to it, so that is the relevant question, but say the word if you would rather it were endpoint-specific or configurable.

### Testing

Five tests added covering unchanged bindings, bindings the device holds that were not cached, bindings the device no longer holds, group bindings not counting as coordinator bindings, clusters left unreportable, and a device that does not support the request. Verified against a live 34 device network.

### DIFF
--- a/src/controller/model/device.ts
+++ b/src/controller/model/device.ts
@@ -26,10 +26,21 @@ import type {
     OtaUpdateAvailableResult,
     ZigbeeOtaImageMeta,
 } from "../tstype";
-import Endpoint, {type BindInternal} from "./endpoint";
+import Endpoint, {type Bind, type BindInternal} from "./endpoint";
 import Entity from "./entity";
 
 const NS = "zh:controller:device";
+
+export interface EndpointBindingDiff {
+    endpoint: Endpoint;
+    bindings: Bind[];
+    /** Bindings the device holds that were not cached */
+    added: Bind[];
+    /** Cached bindings the device does not hold */
+    removed: Bind[];
+    /** Clusters with a configured reporting but no binding to the coordinator, the device cannot report those */
+    unreportableClusters: Cluster[];
+}
 
 const INTERVIEW_GENBASIC_ATTRIBUTES = [
     "modelId",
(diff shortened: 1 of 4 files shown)


Nerivec:
We should just be exposing the bind API which already does the sync, i.e. add a Z2M API endpoint to allow calling the existing ZH API.
I don't see why the ZH API surface has to change this much. Not sure ZH even needs a single change (API is already there, it's just not used by Z2M).

pedrolamas:
That is fair, `bindingTable()` already reconciles, and `added/removed` don't justify an API change.

There is however one piece I'd still argue for, though it needs no ZH change either: the cross-reference of `configuredReportings` against bindings to the coordinator.

Once the read has reconciled the cache, nothing distinguishes a healthy device from a broken one on a second call - `bindings: []` only means something if you also know reporting was configured for those clusters. That check is the only signal that survives repeated calls, which is what turned the original fault from a one-shot observation into something reproducible.

It's `endpoint.binds` against `endpoint.configuredReportings`, both already public, so I'll put it in Z2M unless you'd rather have it here? `bind.ts` already hand-rolls the same composition in `getSetupReportingEndpoints`.

PR #1851, opened 2026-08-20
fix: Improve checkWaitForDataRequestTimeouts()

bueeler:
**ISSUE**
Zigbee2MQTT stops detecting devices as offline, sometimes for hours, sometimes forever, until restarted.
Every outgoing request waits in openRequestsQueue for its confirmation.

A one second timer checks this queue and rejects requests that took too long. This is what marks a device offline when it doesn't respond.
But the timer only checks the first entry in the queue (openRequestsQueue[0]). It never checks entry 1, 2, 3, etc.
If that first entry never gets confirmed, the timer does nothing on every tick forever.
This blocks the timeout check for every entry behind it too. Not just for one device. For all devices sharing the queue.

A confirmation is matched to a queued entry by clusterId and transactionSequenceNumber only. The destination address is not checked.
Two different devices can use the same cluster and sequence number at the same time. A confirmation meant for one request can get matched to the other.
If that happens to the first entry in the queue, it never gets marked confirmed. It sits there forever and blocks everything behind it.

**FIX**
Loop through the whole queue instead of only checking index 0.
Reject any entry whose timeout has passed, regardless of position or confirmed state.
Iterate in reverse so splice() during the loop doesn't skip entries.
Prevent unhandled promise rejection when no listener is attached


Koenkk:
thanks!

PR #1854, opened 2026-08-21
fix: remove every member when deleting a group

### LINKED ISSUE #1853: Group removal skips every other member and leaves persistent ghost memberships
## Summary

`Group.removeFromNetwork()` iterates the live `this._members` array while every successful `endpoint.removeFromGroup(this)` removes that endpoint from the same array. Mutating an array underneath its iterator makes the loop skip the element that shifts into the removed element's index.

On current `master` (`b3e89bd0`), ordinary non-force deletion therefore behaves deterministically:

- members `[A, B]`: only A receives `genGroups.remove`; B is never contacted;
- members `[A, B, C, D]`: only A and C receive it; B and D are never contacted.

The group is then removed from the database. The operation resolves successfully because the skipped endpoints were never sent a command and therefore cannot return an error.

Affected code:
https://github.com/Koenkk/zigbee-herdsman/blob/b3e89bd002f6395aa258a7a0350d41753484ed29/src/controller/model/group.ts#L165-L171

## Minimal reproduction

1. Start a controller and create a group.
2. Join four mock devices and add endpoint 1 from each to the group.
3. Clear `sendZclFrameToEndpoint` calls.
4. Run `await group.removeFromNetwork()`.
5. Inspect the destination IEEE addresses.

Expected: four `genGroups.remove` unicasts, one per endpoint.

Actual: two unicasts, to members 1 and 3; members 2 and 4 are silently skipped. The group is absent from the database afterward.

A regression test against the unmodified code fails with:

```
expected 4 calls
received 2 calls
```

## Consequences

This creates a persistent physical/logical split-brain during the normal Zigbee2MQTT GUI deletion path:

- skipped devices retain the group ID in their nonvolatile Groups table;
- zigbee-herdsman/Zigbee2MQTT delete the only record that those memberships exist;
- the inconsistency survives process restarts and normally device power cycles;
- when the freed numeric group ID is reused, the hidden members execute unrelated future groupcasts;
- lamps or Groups-capable switched loads can turn ON/OFF unexpectedly while Zigbee2MQTT and MQTT continue to report their logical state;
- no diagnostic is emitted for skipped endpoints.

For lighting this produces apparently spontaneous physical state changes. For plugs or other actuators supporting the Groups cluster, the same defect can have safety, security, or property consequences.

## History

The live-array loop was introduced by:
https://github.com/Koenkk/zigbee-herdsman/commit/df201422759d23ecc78d867e0b30b73458900680

That commit dates from 2020-07-07 and first appears in `v0.12.100`; the defect remains present in `10.9.0` and current `master`.

Ironically, the method was added to address stale physical group membership:
https://github.com/Koenkk/zigbee2mqtt/issues/3861

## Proposed fix

Iterate over a snapshot:

```ts
for (const endpoint of [...this._members]) {
    await endpoint.removeFromGroup(this);
}
```

This preserves the existing sequential command/error behavior while ensuring that every original member is contacted.

I have a two-file patch with the regression test. With the fix, the targeted test, Biome, TypeScript, the full test suite (1767 passed, 1 skipped), and 100% coverage all pass. A PR will follow and reference this issue.

gbarkan:
## Summary

- iterate over a snapshot of `Group._members` while removing a group from the network;
- add a four-member regression test proving that every original member receives `genGroups.remove`;
- preserve the existing sequential command and error semantics.

Fixes #1853.

## Root cause

`endpoint.removeFromGroup(this)` removes the endpoint from the same `_members` array that `Group.removeFromNetwork()` is iterating. Iterating the live array skips every element that shifts into the just-removed index: a two-member group contacts one member, and a four-member group contacts members 1 and 3 only.

Afterward the group database record is deleted, leaving the skipped devices with hidden physical memberships. Reuse of that numeric group ID can make those devices execute unrelated future groupcasts.

## Validation

Before the fix, the new regression test fails with 2 calls instead of 4. With this patch:

- targeted regression test: pass;
- Biome: pass;
- TypeScript `tsc --noEmit`: pass;
- full test suite: 1767 passed, 1 skipped;
- coverage: 100% statements, branches, functions, and lines.

### DIFF
--- a/src/controller/model/group.ts
+++ b/src/controller/model/group.ts
@@ -163,7 +163,7 @@ export class Group extends ZigbeeEntity {
     }
 
     public async removeFromNetwork(): Promise<void> {
-        for (const endpoint of this._members) {
+        for (const endpoint of [...this._members]) {
             await endpoint.removeFromGroup(this);
         }
 
--- a/test/controller.test.ts
+++ b/test/controller.test.ts
@@ -4088,6 +4088,30 @@ describe("Controller", () => {
         );
     });
 
+    it("Removes every member when removing a group from the network", async () => {
+        await controller.start();
+        const group = controller.createGroup(4);
+
+        const devices = [
+            {networkAddress: 129, ieeeAddr: "0x129"},
+            {networkAddress: 170, ieeeAddr: "0x170"},
+            {networkAddress: 171, ieeeAddr: "0x171"},
+            {networkAddress: 175, ieeeAddr: "0x175"},
+        ];
+
+        for (const device of devices) {
+            await mockAdapterEvents.deviceJoined(device);
+            group.addMember(controller.getDeviceByIeeeAddr(device.ieeeAddr)!.getEndpoint(1)!);
+        }
+
+        mocksendZclFrameToEndpoint.mockClear();
+        await group.removeFromNetwork();
+
+        expect(mocksendZclFrameToEndpoint).toHaveBeenCalledTimes(4);
+        expect(mocksendZclFrameToEndpoint.mock.calls.map((call) => call[0])).toStrictEqual(devices.map((device) => device.ieeeAddr));
+        expect(controller.getGroupByID(4)).toBeUndefined();
+    });
+
     it("Removes group from database", async () => {
         await controller.start();
 


Koenkk:
thanks!

Issue #1796, opened 2026-06-24
Group ID issues


andrei-lazarov:
Hi. I found multiple issues regarding Group IDs.
Can you help with some? cc @Nerivec 

<img width="80%" src="https://github.com/user-attachments/assets/52816009-e10f-481e-9bd0-18af1fb763b6" />
<p/>

Mainly the range seems it should be `0x0001` - `0xfff7`

- [x] 1. WindFront has wrong min and max Group ID
  https://github.com/Nerivec/zigbee2mqtt-windfront/pull/592
- [x] 2. WindFront allows creating groups past min and max limits. The arrows work in the interval, but you can type any number
- [x] 3. Z2M allows creating groups past `0xfff7`
  https://github.com/Koenkk/zigbee-herdsman/pull/1795
- [x] 4. Z2M allows adding devices to groups `0xfff8` - `0xffff`
- [x] 5. Z2M thinks the device was added successfully to `0xfff8` - `0xffff`, when they actually fail with Invalid Value `0x87`
  https://github.com/Koenkk/zigbee-herdsman/pull/1855
  <img width="60%" src="https://github.com/user-attachments/assets/4ab0bfae-13ff-4001-a974-069862fe5588" />
<p/>

- [ ] 6. Can't add devices to `default_bind_group`.
  Z2M creates `default_bind_group` in `database.db` and binds some devices to it during config (e.g. [IKEA E1524/E1810](https://www.zigbee2mqtt.io/devices/E1524_E1810.html#ikea-e1524-e1810)), but it doesn't add it to `configuration.yaml` - so it says it doesn't exist when trying to use it.
- [ ] 7. Some Aqara devices are in "group 0" by default. Must be removed during reconfiguration to avoid random toggles. (Like in [ZHA](https://github.com/zigpy/zha-device-handlers/blob/6a3822c1348f9e40ad243eb881084a1c075da277/zhaquirks/xiaomi/aqara/plug_eu.py#L45))
- [x] 8. Check fractional groups?
  https://github.com/Koenkk/zigbee-herdsman/pull/1863
  https://github.com/Nerivec/zigbee2mqtt-windfront/pull/677

Nerivec:
- 2 - should be solved with PR for 1. (see comment)
- 4 - should be solved with PR for 3 (if group can't exist, can't add to it).
- 5 - status check is enabled for `addToGroup`'s `zclCommand`, so, it should fail
[...]

PR #1855, opened 2026-08-21
fix: honor group command response status

gbarkan:
## Summary

- keep the generic/Foundation `Endpoint.checkStatus()` behavior unchanged;
- preserve Default Response `statusCode` validation and add Groups-specific `status` checks inside `addToGroup()` and `removeFromGroup()`;
- reject Add Group and Remove Group failures before mutating zigbee-herdsman's membership database;
- treat Remove Group `NOT_FOUND` as idempotent success, because the endpoint is already absent from the physical group;
- add regression tests for Add Group `INSUFFICIENT_SPACE`, Remove Group `FAILURE`, and Remove Group `NOT_FOUND`.

Addresses #1796, specifically item 5.

## Root cause

Groups cluster-specific responses are objects shaped as `{status, groupid}`. The generic status checker is intentionally limited to Foundation response shapes: array records with `status`, and Default Response objects with `statusCode`. Consequently, it does not reject a failed Groups `addRsp` or `removeRsp`.

This lets logical and physical membership diverge in both directions:

- an endpoint can return `INSUFFICIENT_SPACE` to Add Group, while herdsman records a membership the endpoint rejected;
- an endpoint can return `FAILURE` to Remove Group, while herdsman erases a membership that may still exist physically.

The latter becomes a hidden live membership when the group ID is later reused. The former creates a phantom logical membership and missed groupcasts while optimistic state may still claim success.

The fix deliberately validates `payload.status` only in the two Groups operations where that field is defined as a ZCL Groups status. Existing Foundation/Default Response checking remains in place, so a Groups request that receives a failed `defaultRsp` is still rejected.

## Specification

Zigbee Cluster Library R8 defines the unicast Add Group Response and Remove Group Response payloads as `Status` plus `Group ID`; Add Group may return `INSUFFICIENT_SPACE`, and Remove Group returns `NOT_FOUND` when the endpoint does not hold that group.

https://csa-iot.org/wp-content/uploads/2022/01/07-5123-08-Zigbee-Cluster-Library-1.pdf

Relevant sections: 3.6.2.3.2.2 / 3.6.2.4.1 and 3.6.2.3.5.2 / 3.6.2.4.4.

## Validation

Against the unmodified code, the new Add Group and Remove Group failure tests fail because both promises resolve. With this patch:

- all three targeted regressions pass;
- the existing failed-Default-Response Groups regression also passes;
- Biome: pass;
- TypeScript `tsc --noEmit`: pass;
- full test suite: 1770 passed, 1 skipped;
- coverage: 100% statements, branches, functions, and lines.

### DIFF
--- a/src/controller/model/endpoint.ts
+++ b/src/controller/model/endpoint.ts
@@ -452,10 +452,10 @@ export class Endpoint extends ZigbeeEntity {
     /*
      * Zigbee functions
      */
-    private checkStatus(payload: [{status: Zcl.Status}] | {status?: Zcl.Status; statusCode?: Zcl.Status}): void {
-        const codes = Array.isArray(payload) ? payload.map((i) => i.status) : [payload.status ?? payload.statusCode];
-        const invalid = codes.find((c) => c !== undefined && c !== Zcl.Status.SUCCESS);
-        if (invalid !== undefined) throw new Zcl.StatusError(invalid);
+    private checkStatus(payload: [{status: Zcl.Status}] | {cmdId: number; statusCode: number}): void {
+        const codes = Array.isArray(payload) ? payload.map((i) => i.status) : [payload.statusCode];
+        const invalid = codes.find((c) => c !== Zcl.Status.SUCCESS);
+        if (invalid) throw new Zcl.StatusError(invalid);
     }
 
     public async report<Cl extends number | string, Custom extends TCustomCluster | undefined = undefined>(
@@ -1082,7 +1082,20 @@ export class Endpoint extends ZigbeeEntity {
     }
 
     public async addToGroup(group: Group): Promise<void> {
-        await this.zclCommand("genGroups", "add", {groupid: group.groupID, groupname: ""}, undefined, undefined, true, Zcl.FrameType.SPECIFIC);
+        const response = await this.zclCommand(
+            "genGroups",
+            "add",
+            {groupid: group.groupID, groupname: ""},
+            undefined,
+            undefined,
+            true,
+            Zcl.FrameType.SPECIFIC,
+        );
+
+        if (response?.payload.status !== undefined && response.payload.status !== Zcl.Status.SUCCESS) {
+            throw new Zcl.StatusError(response.payload.status);
+        }
+
         group.addMember(this);
     }
 
@@ -1105,7 +1118,7 @@ export class Endpoint extends ZigbeeEntity {
      */
     public async removeFromGroup(group: Group | number): Promise<void> {
         try {
-            await this.zclCommand(
+            const response = await this.zclCommand(
                 "genGroups",
                 "remove",
                 {groupid: group instanceof Group ? group.groupID : group},
@@ -1114,6 +1127,14 @@ export class Endpoint extends ZigbeeEntity {
                 true,
                 Zcl.FrameType.SPECIFIC,
             );
+
+            if (
+                response?.payload.status !== undefined &&
+                response.payload.status !== Zcl.Status.SUCCESS &&
+                response.payload.status !== Zcl.Status.NOT_FOUND
+            ) {
+                throw new Zcl.StatusError(response.payload.status);
+            }
         } catch (error) {
             if (!(error instanceof Zcl.StatusError) || error.code !== Zcl.Status.NOT_FOUND) {
                 throw error;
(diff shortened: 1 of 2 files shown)


gbarkan:
Suspect both tests are flaky in CI:

Error: Error: Test timed out in 5000ms.
 ❯ test/controller.test.ts:757:5

Error: AssertionError: expected 174.964919 to be greater than or equal to 175
 ❯ test/device-ota.test.ts:2143:33

test/controller.test.ts
gbarkan, line 1: Suspect a flaky controller start timeout

Nerivec:
I'm not sure we can do that, I have a suspicion this is going to create false-positives (any payload using `status` would be checked - not all could be for ZCL status). The Foundation array-payload and `statusCode` were specific enough to prevent conflict - and were designed that way.

I think checkStatus should be left alone (too wide an impact for targeted checks like this), and an extra check should be added specifically in removeFromGroup & co.

src/controller/model/endpoint.ts
```diff
@@ -1104,15 +1117,29 @@ export class Endpoint extends ZigbeeEntity {
...
+                undefined,
+                undefined,
+                true,
+                Zcl.FrameType.SPECIFIC,
+            );
+
+            if (
+                response?.payload.status !== undefined &&
+                response.payload.status !== Zcl.Status.SUCCESS &&
+                response.payload.status !== Zcl.Status.NOT_FOUND
```
Nerivec: Should not throw for this one,and remove the whole try/catch.

gbarkan:
Agreed. I restored `checkStatus()` unchanged and moved the Groups-specific checks into `addToGroup()` and `removeFromGroup()`.

Both methods retain `checkStatus=true` for Foundation Default Responses, then validate the Groups-specific `payload.status` locally. Add accepts only `SUCCESS`; Remove accepts `SUCCESS` and idempotent `NOT_FOUND`.

The full suite passes (1770 passed, 1 skipped), with 100% statement, branch, function, and line coverage.

src/controller/model/endpoint.ts
Nerivec: Also might want to `info` log if NOT_FOUND; it's useful information.
```
if !SUCCESS => throw
else if NOT_FOUND => info log
```

gbarkan:
Thanks — both suggestions are now addressed.

- Removed the whole `try/catch`. A Groups-specific `removeRsp.status === NOT_FOUND` is treated as idempotent success; all other non-`SUCCESS` Groups statuses and Foundation/default-response failures propagate normally.
- Added an `info` log for `NOT_FOUND`, including the group ID and endpoint, plus a regression assertion.

Implemented in `42bd346` and `fe8b9d4`. The full local suite passes (1770 passed, 1 skipped) with 100% coverage, and GitHub CI is green.

gbarkan:
Re-styled to look nice and symmetrical

gbarkan:
Hello, @Nerivec! Can you please check the latest committed state? I took all your comments into account.

src/controller/model/endpoint.ts
```diff
@@ -1082,7 +1082,13 @@ export class Endpoint extends ZigbeeEntity {
     }
 
     public async addToGroup(group: Group): Promise<void> {
-        await this.zclCommand("genGroups", "add", {groupid: group.groupID, groupname: ""}, undefined, undefined, true, Zcl.FrameType.SPECIFIC);
+        const payload = {groupid: group.groupID, groupname: ""};
```
Nerivec: Put this directly in the zclCommand call like it was before, otherwise we don't get the nice typing/automatic link in code editors.
Same for remove.

gbarkan: Sorry, took long, got distracted at work. Now fixed! I also guess I switch from inline style to multiline correctly as it becomes out of 150 chars limit.

gbarkan:
Seems everything is in order now

Nerivec:
@andrei-lazarov any chance you can give this a spin in real world scenario?

andrei-lazarov:
Yes, it looks good!

Adding fails with the correct error:
(patched to allow invalid groups)

image
image

Removing clears the database and info-logs the error:

image

andrei-lazarov (approved):

Nerivec (approved):

Koenkk:
Thanks!

PR #1856, opened 2026-08-28
fix: patch serialport bindings-cpp for Node v26.3+

Nerivec:
Pending fixes in upstream, https://github.com/serialport/bindings-cpp/pull/239 https://github.com/serialport/bindings-cpp/pull/240

Prevents issues with latest Node 26 versions - *Windows side untested*.
Also removes prebuilds to force build during install.


Koenkk:
Thanks!

PR #1858, opened 2026-09-02
fix: better handle discovery errors

Nerivec:
Should also watch out for next version of bonjour-service (once released) that will fix a deep, currently-uncatchable, throw.
That and this, will take care of https://github.com/Koenkk/zigbee2mqtt/issues/32920


Nerivec:
We're going to have to fix the two nasty buggers in tests, race condition problems, they seem to show up quite a lot recently:
https://github.com/Koenkk/zigbee-herdsman/actions/runs/33644423715/attempts/1

Koenkk:
> We're going to have to fix the two nasty buggers in tests, race condition problems, they seem to show up quite a lot recently: https://github.com/Koenkk/zigbee-herdsman/actions/runs/33644423715/attempts/1

Yeah I also noticed, seems to be some kind of rounding error.

PR #1859, opened 2026-09-03
fix(ember): bounded transient-send retry for ZDO and ZCL broadcast; make permitJoin transactional

analienx:
## Problem

On EmberZNet NCPs (observed on a Sonoff Dongle-M, EmberZNet 9.1.1 / EZSP 19, under real production send pressure), `EmberAdapter.sendZdo()` — including the `permit_join` ZDO broadcast — and `EmberAdapter.sendZclFrameToAll()` fail the whole request on transient NCP send statuses (`SLStatus.BUSY`, `SLStatus.ZIGBEE_MAX_MESSAGE_LIMIT_REACHED`). These statuses mean the NCP never accepted the frame, so the request is lost even though a deferred re-send is safe. The ZCL **unicast** path in the same adapter already implements exactly this recovery; the ZDO and ZCL broadcast paths did not.

Additionally, `Controller.permitJoin()` opens the adapter permit (and the Green Power stage) sequentially without cleanup: if a later stage throws, the earlier stages stay open — the network remains joinable while the caller receives an error (a half-open state we reproduced and measured).

## Changes

1. **`sendZdo()`** — bounded transient retry within the same logical request: same `messageTag`/payload (the NCP never accepted the frame), max `QUEUE_MAX_SEND_ATTEMPTS` (3) attempts, `QUEUE_BUSY_DEFER_MSEC` (500 ms) defer, only for `BUSY` / `ZIGBEE_MAX_MESSAGE_LIMIT_REACHED`; other non-OK statuses fail immediately with unchanged error strings. Mirrors the existing ZCL unicast recovery parameters.
2. **`sendZclFrameToAll()`** — same bounded transient retry for the ZCL broadcast path (the Green Power stage of `Controller.permitJoin()`).
3. **`Controller.permitJoin()`** — adapter + Green Power stages now form one transaction; on partial failure both stages are rolled back (closed) with identical target semantics, cleanup errors are logged and never replace the original error, and the original error is rethrown.

No behavior changes for non-transient failures; no unrelated group/multicast methods touched.

## Tests

- `emberAdapter.test.ts`: ZDO retry matrix (OK; BUSY→OK; BUSY→BUSY→OK; BUSY×3 → terminal failure; non-transient → immediate failure) and the same matrix for ZCL broadcast (fake timers, deferred-send assertions).
- `controller.test.ts`: permitJoin rollback of both stages on partial failure, original error preserved.
- Full suite: 34 files / 1774 passed, 1 skipped. `biome check` clean on all touched files.

## Real-world evidence

Transaction-tagged A/B on a production EmberZNet 9.1.1 network before/after this patch (10 s windows, MQTT API path, no pairing):

| Scenario | Before | After |
|---|---|---|
| Coordinator-only permit join | 3/3 OK | 3/3 OK |
| Permit Join All (broadcast) | 1/5 OK (3× `ZDO PERMIT_JOINING_REQUEST BROADCAST to=65532 status=BUSY`, 1× `ZCL BROADCAST destination=65533 status=BUSY`) | **10/10 OK**, zero terminal BUSY |

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -1789,56 +1789,71 @@ export class EmberAdapter extends Adapter {
             let status: SLStatus | undefined;
             let apsSequence: number | undefined;
 
-            if (ZSpec.Utils.isBroadcastAddress(networkAddress)) {
-                logger.debug(
-                    () => `~~~> [ZDO ${clusterName} BROADCAST to=${networkAddress} messageTag=${messageTag} payload=${payload.toString("hex")}]`,
-                    NS,
-                );
+            const isBroadcast = ZSpec.Utils.isBroadcastAddress(networkAddress);
 
-                [status, apsSequence] = await this.ezsp.ezspSendBroadcast(
-                    ZSpec.NULL_NODE_ID, // alias
-                    networkAddress,
-                    0, // nwkSequence
-                    apsFrame,
-                    ZDO_REQUEST_RADIUS,
-                    messageTag,
-                    payload,
-                );
-
-                apsFrame.sequence = apsSequence;
-
-                logger.debug(`~~~> [SENT ZDO BROADCAST messageTag=${messageTag} apsSequence=${apsSequence} status=${SLStatus[status]}]`, NS);
-
-                if (status !== SLStatus.OK) {
-                    throw new Error(
-                        `~x~> [ZDO ${clusterName} BROADCAST to=${networkAddress} messageTag=${messageTag}] Failed to send request with status=${SLStatus[status]}.`,
+            // Transient send-pressure statuses are retried within the same logical request
+            // (same messageTag/payload: the NCP never accepted the frame), mirroring the
+            // recovery the ZCL unicast path already has.
+            for (let attempt = 1; attempt <= QUEUE_MAX_SEND_ATTEMPTS; attempt++) {
+                if (isBroadcast) {
+                    logger.debug(
+                        () => `~~~> [ZDO ${clusterName} BROADCAST to=${networkAddress} messageTag=${messageTag} payload=${payload.toString("hex")}]`,
+                        NS,
                     );
+
+                    [status, apsSequence] = await this.ezsp.ezspSendBroadcast(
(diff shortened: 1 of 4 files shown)


src/controller/controller.ts
```diff
@@ -390,8 +390,29 @@ export class Controller extends events.EventEmitter<ControllerEventMap> {
             // never permit more than uint8, and never permit 255 that is often equal to "forever"
             assert(time <= 254, "Cannot permit join for more than 254 seconds.");
 
-            await this.adapter.permitJoin(time, device?.networkAddress);
-            await this.#greenPower.permitJoin(time, device?.networkAddress);
+            try {
```
Nerivec, line 393: This should be two try/catch around each call instead of nested try/catch.

Also need to reduce the amount of comments - a lot...

src/adapter/ember/adapter/emberAdapter.ts
```diff
@@ -2112,17 +2127,34 @@ export class EmberAdapter extends Adapter {
...
             logger.debug(() => `~~~> [ZCL BROADCAST apsFrame=${JSON.stringify(apsFrame)} header=${JSON.stringify(zclFrame.header)}]`, NS);
-            const [status] = await this.ezsp.send(
-                EmberOutgoingMessageType.BROADCAST,
-                destination,
-                apsFrame,
-                data,
-                0, // alias
-                0, // alias seq
-            );
+            let status: SLStatus = SLStatus.FAIL;
```
Nerivec, line 2130: Don't think we want this on broadcast. Chances are, that would put the NCP further into stress. Same for the broadast path of sendZdo.

Nerivec:
I'm not sure what's causing this on your end but considering the percentage of failures reported, seems you have a problem other than the software.
I recently re-paired a whole network on MG24 9.1.1, did dozens of permit join (bcast) at various stages of busyness, never had a single failure. We've also stress-tested the ember driver with over 300 devices (on MG24 coordinator), and this did not surface.
Should also mention that we had retries on all queued stuff in initial implementation of ember driver, which we removed in favor of unicast-only, because overall it caused more problems than it solved.

On top of the other posted comments, the ember changes would also change the overall logic on the base API compared to other adapters.

PR #1863, opened 2026-09-04
fix: prevent creating groups with fractional IDs

andrei-lazarov:
image

image

### DIFF
--- a/src/controller/model/group.ts
+++ b/src/controller/model/group.ts
@@ -144,7 +144,7 @@ export class Group extends ZigbeeEntity {
     }
 
     public static create(groupID: number): Group {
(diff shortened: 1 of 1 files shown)


src/controller/model/group.ts
```diff
@@ -144,7 +144,7 @@ export class Group extends ZigbeeEntity {
     }
 
     public static create(groupID: number): Group {
-        assert(typeof groupID === "number", "GroupID must be a number");
+        assert(Number.isInteger(groupID), "GroupID must be a natural number");
```
Koenkk:
```suggestion
        assert(Number.isInteger(groupID), "GroupID must be a whole number");
```

Koenkk:
Thanks!

PR #1864, opened 2026-09-04
feat(ember): expose source route table size in stack config

mod242:
Certain legacy devices can become unreachable after the Ember NCP learns and stores a multi-hop source route. In the reproduced case described in Koenkk/zigbee2mqtt#32910, configuring the source route table size to zero kept direct
unicasts stable.

This change exposes SOURCE_ROUTE_TABLE_SIZE through the existing Ember stack_config.json expert configuration.
- Accepts integer values from 0 through 255
- Preserves zero as a valid value
- Leaves the existing NCP configuration unchanged when omitted
- Does not change concentrator mode, MTORR behavior, APS options, or source route discovery mode
- Does not claim to disable source routing network-wide

This is intended as an expert compatibility workaround and diagnostic option for affected networks.

### DIFF
--- a/src/adapter/ember/adapter/emberAdapter.ts
+++ b/src/adapter/ember/adapter/emberAdapter.ts
@@ -162,6 +162,8 @@ type StackConfig = {
      * A value of 0 will be converted to the EMBER_MAX_HOPS value set by the stack.
      */
     CONCENTRATOR_MAX_HOPS: number;
+    /** Optional Ember NCP source route table size override <0-255>. @see EzspConfigId.SOURCE_ROUTE_TABLE_SIZE */
+    SOURCE_ROUTE_TABLE_SIZE?: number;
     /** <6-64> (Default: 6) @see EzspConfigId.MAX_END_DEVICE_CHILDREN */
     MAX_END_DEVICE_CHILDREN: number;
     /** <-> (Default: 10000) @see EzspValueId.TRANSIENT_DEVICE_TIMEOUT */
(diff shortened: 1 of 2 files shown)


Nerivec:
There appears to be a weird side-effect from doing this rather than disabling concentrator mode (proper behavior). We cannot rely on this being stable long term, not even short term really (could change with any new EmberZNet release).

As mentioned in that thread, other devices on the network may also do source routing, which would render this setting useless (i.e. may work for some users, may not for others, depending on other devices on the network...).

It also would not fix the problem with other drivers (since caused by the devices in question, rather than the coordinator - and confirmed as such in that thread).

Realistically though, I don't think we'll ever be able to fix all the problems associated with these old _bad_ chips. And I'd rather not potentially introduce new ones from a config with an unknown side-effect.

I can create an EmberZNet firmware branch for these, if users want to test further. It provides the same as this PR would, without affecting Z2M at all.

@andrei-lazarov I think you've tracked down several affected models. If not already done, we should add big caution messages in the docs "DO NOT BUY" :sweat_smile:

burmistrzak:
I'd agree with @Nerivec here.
Best we can do is being spec-compliant and report problematic devices to the manufacturer. If they're unwilling or unable to fix their products, flag them in docs as **DO NOT BUY**.

This is of course assuming we have implemented everything correctly and ZHA/zigpy is affected as well. 🤔

mod242:
I totally understand your position, that's why I declared this as an expert compatibility workaround and diagnostic option for affected networks. But I understand that it might still not be desired to have such an option at all. I can go ahead and close this PR from my side. I can live with my workaround, patching the node-modules directly, since it works for me in my specific envrionment and others who are affected can do the same, I guess.
I really do understand that you don't want to bloat the configuration options just to support edge-cases where the manufactor did something stupid.

Nerivec:
Closing, see [reason](https://github.com/Koenkk/zigbee2mqtt/issues/32910#issuecomment-5545581705).

andrei-lazarov:
> This is of course assuming we have implemented everything correctly and ZHA/zigpy is affected as well. 🤔

If I understand correctly:

ZHA offers the option
```yaml
zha:
  zigpy_config:
    source_routing: true
```

Which is (still) disabled by default
https://github.com/search?q=repo%3Azigpy%2Fzigpy+source_routing&type=code

For compatibility
> You might see some non-responsive devices after activating source routing.
https://www.reddit.com/r/homeassistant/comments/1iq9dav/comment/mczu8l0/

I'm not sure if other adapters are affected. I only saw ember reports. From same comment:
> If you're using a TI coordinator and ZHA, you should already be safe to toggle it on, since ZHA/zigpy-znp will fall back to using route discovery if an initial request with forced relays didn't go through (source routing always done by ZHA here).

Anyway thanks for looking into it! I can document the crappy devices and maybe add a reference in troubleshooting

burmistrzak:
@andrei-lazarov AFAICT, seems like TI is affect as well. See https://github.com/Koenkk/zigbee2mqtt/issues/32910#issuecomment-5416885478

With ZHA having a similar workaround, it's reasonable to assume that lightbulb firmware/hardware is to blame.

PR #1869, opened 2026-09-10
chore: update to pnpm v12

Nerivec:
Looks like patching is broken, won't work when deletes are present :cold_sweat: 
@Koenkk I'm thinking it might be easier to use a git source with a fork of bindings-cpp (can specify a tarball URL from GH release of fork - no npm pub needed)?


Koenkk:
The git source sounds easier indeed

DragonWork:
@Koenkk Yeah, just like I suggested two weeks ago. https://github.com/serialport/bindings-cpp/pull/239#issuecomment-5430491151

@Nerivec, for the bindings part of this, I’ve now published `@dragonwork/serialport-bindings-cpp@13.1.0` **until the upstream fixes land**.

It includes the Unix and Windows callback fixes and rebuilt native binaries. I updated some dependencies because GitHub Actions would fail otherwise. It’s built and published directly through GitHub Actions/OIDC with **verified provenance**. **GitHub release immutability is enabled** too, and the commit and tag are SSH-signed. (You can verify the provenance at https://www.npmjs.com/package/@dragonwork/serialport-bindings-cpp at the end of the page. I find this important because it ships prebuilt binaries for all supported platforms.)

Alongside the pnpm v12 update, you could replace its existing dependency with:

```json
"@serialport/bindings-cpp": "npm:@dragonwork/serialport-bindings-cpp@13.1.0"
```

Then remove the old bindings patch and refresh the lockfile. The dependency alias carries through to Zigbee2MQTT without relying on downstream patch application then.

Nerivec:
@DragonWork this is not the road we want to take; this would split the development, we only want to patch until upstream is fixed. Also Koenkk will own the repository for this.

DragonWork:
> @DragonWork this is not the road we want to take. Also Koenkk will own the repository for this.

Alright. Just wanted to help. It was thought as a temporary fix anyways (you wouldn’t depend on that package but add an npm alias until upstream is fixed). It stays published from my side and with a fully transparent and verifiable build pipeline. You decide. 🫶🏻

PR #1872, opened 2026-09-12
feat!: unified transport

Nerivec:
Long time coming...
Since we're probably looking at a serialport overhaul in the near future, I figured it was worth taking the time to check that one off the todo list. Should make migration much easier if we move to napi-rs based, as currently being investigated.
It's a rather aggressive update that concentrates and optimizes the port-related stack as much as possible.
_Tests were rather nasty to update..._ :hot_face: 
Spectacular cleanup though :notes: 

Also included an alternative workaround for the node.js > v26.3 serialport problem that does not require any kind of patching in dependencies... probably best course of action for now... hopefully fixed soon in upstream lib.

@Koenkk will need an in-depth review of that one, make sure I didn't miss anything.
Also, if you see anything that can be further improved (particularly with the new transport API)... :grimacing:

Tested with:
- ember (MG24)
- zstack (CC2674P10)
- zoh

Supersedes & closes https://github.com/Koenkk/zigbee-herdsman/pull/1821
Supersedes & closes https://github.com/Koenkk/zigbee-herdsman/pull/1828

TODO:
- [x] test coverage for the new transport class
- [ ] future PR: revamp flow control option to `"none" | "software" | "hardware"` instead of awkwardly toggling port options with boolean checks (will require a migration in Z2M to prevent breaking change).

CC: @kirovilya @manup @chris-1243 @asgothian @canardos @DragonWork @andrei-lazarov


Koenkk:
Looks really good, nice cleanup, I'm wondering if we should wait for the 1 Aug release?

DragonWork:
~62 callbacks/sec (via these no-op timers) seems acceptable as a temporary workaround. The native fix resumed the Promise in about 0.4 ms in the local check; here it can wait for the next 16 ms tick, or longer under load. That trade-off seems fine for now, though it adds idle work even without serial traffic.

What worries me is the timer lifecycle. I tested with a mocked port: three open/close cycles left three intervals running, and a failed open added a fourth.

So this isn’t capped at ~62 callbacks/sec. Every retry or reopen adds another timer, and they all keep firing for the rest of the daemon’s lifetime. It grows linearly with open attempts.

Could we keep one timer per transport and clear it on failure, close and disconnect? Otherwise it leaks.

Edit: [Here’s the link to the affected code.](https://github.com/Koenkk/zigbee-herdsman/blob/79d22aeabe4d688315fff66ed56ddefe38662f9e/src/adapter/transport.ts#L187-L196)

DragonWork:
Now I’m happy 🙃

Nerivec:
@Koenkk yes should wait after Oct release, hopefully we get feedback from multiple sources with multiple adapters.

@DragonWork nice catch, forgot to clear it :+1: 

Sidenote: I still think we should keep the engine limit in Z2M for now. Less chance of troubles.

PR #1874, opened 2026-09-13
fix: Preserve 0xFFFF manufacturerCode in genOta requests

syenchuk:
## Issue

Some devices without a CSA-assigned manufacturer code can use `0xFFFF` which is one of the reserved values. Manufacturer code `0xFFFF` can also be used as a wildcard in various scenarios according to CSA, for instance:

> **11.13.9.6.3 Manufacturer Code**
> Manufacturer code MAY be sent using wildcard value of 0xfffff in order to apply the command to all devices disregard of their manufacturers. See Chapter 2 for detailed description.

The issue is that `0xFFFF` is also a ZCL non-value for UINT16, so `processParameterRead` replaces `0xFFFF` by `Number.NaN` unless `0xFFFF` is used as a `min` or `max` parameter in the attribute specification:

```typescript
    // if type does not have an `invalid` (undefined) it won't match since value is checked above
    if (value === ZCL_TYPE_INVALID_BY_TYPE[parameter.type]) {
        // if value is same as max or min, ignore invalid sentinel
        if (isMinOrMax(parameter, value)) {
            return value;
        }

        // return NaN for both number & bigint to keep logic consistent
        return Number.NaN as T;
    }
```

However, current definitions for `manufacturerCode` don't specify any min / max bounds:

```ts
{name: "manufacturerCode", type: DataType.UINT16},
```

And therefore the index lookup in `Device.findMatchingOtaImage` fails:

```ts
i.manufacturerCode === current.manufacturerCode // 65535 === NaN
```

The consequence is that since #1503, OTA update detection never works for these devices: no image is ever matched and `update_available` is always `false`.

The fix is to simply define `0xFFFF` as the max value so that `0xFFFF` is treated as is and never converted to `NaN`.


Nerivec (approved):

Koenkk:
Thanks!

PR #1876, opened 2026-09-14
fix: Cancel interrupted adapter commands during recovery and shutdown

### LINKED ISSUE #1875: EZSP recovery can strand device commands when AsyncMutex.clear() drops waiting promises
In zigbee-herdsman 10.9.2, `AsyncMutex.clear()` discards queued resolvers without settling their promises. EZSP recovery calls it through `Driver.reset()` → `stop(false)` → `Ezsp.close()`.

An adapter operation waiting on that mutex can remain pending indefinitely, retaining its network-address slot in `Queue`. Subsequent commands to that device cannot dispatch; other devices can continue working. The ZCL response timeout starts after dispatch, so it cannot release this slot.

### Isolated reproduction

Confirmed using the installed `AsyncMutex` and `Queue` classes:

1. Hold an `AsyncMutex.run()` call with a deferred promise.
2. Through `Queue(4).execute(..., deviceA)`, submit another call to that mutex and let it enqueue.
3. Call `mutex.clear()`, then resolve the first operation.
4. Submit another adapter job for `deviceA` and one for `deviceB`.

Observed: the interrupted operation stays pending, the subsequent `deviceA` job never starts, and `deviceB` completes.

### Proposed transport-level simulation

Keep the real EZSP driver and queues; mock the serial transport. Suppress an outgoing frame's ACK and its retry ACK while device commands are queued. Advance fake timers through both ACK timeouts, then supply successful reset/startup responses. Submit fresh commands for the affected devices and an unaffected device.

Expected: interrupted commands reject, their adapter slots release, and fresh commands complete after recovery.

joejulian:
`AsyncMutex.clear()` rejects queued acquisitions with `MutexCancelledError` while preserving mutex ownership. Cancellation propagates through device commands and response-waiter cleanup, bypasses ZiGate command retries and Z-Stack node-descriptor route discovery, and removes cancelled requests from endpoint deferred queues.

EZSP marks shutdown before clearing its mutex. Cancelled or closing watchdog requests do not count as heartbeat failures or trigger resets, and active UART commands stop retrying after transport close.

Regression tests cover immediate and delayed shutdown with a queued watchdog, shutdown during the UART retry delay, cancellation handling in Z-Stack/ZiGate/ZBOSS and endpoint queues, and real EZSP recovery/startup with mocked serial transport followed by successful device commands.

Fixes #1875.


Nerivec:
Rejecting like this can easily causing problems (race conditions) with shutdown behavior (rejections happening too late, causing undesirable errors).
I just had to remove a scenario like this in another project a few days ago because of that.

joejulian:
`Ezsp.close()` marks shutdown before cancelling waiters. Watchdog cancellation does not count as a heartbeat failure or emit reset, and UART retries abort during close. Typed cancellation bypasses ZiGate retries, Z-Stack node-descriptor route discovery, and endpoint deferred retention; ZBOSS propagates it with response-waiter cleanup.

Regression tests cover immediate/delayed close with a queued watchdog, shutdown during the retry delay, and those cancellation paths. The mocked-serial recovery/startup test and subsequent device commands pass.

joejulian:
@Nerivec did I address your concerns sufficiently?

PR #1878, opened 2026-09-15
feat(group): accept explicit manufacturer-specific cluster definitions

SisyphusMD:
Hue effects fail when a Zigbee group contains both Hue bulbs and an Inovelli switch, because the Philips custom cluster is not shared by every member. Original report: https://github.com/Koenkk/zigbee-herdsman-converters/pull/12008#issuecomment-4365112667

Allow `Group.command` to accept an explicit manufacturer-specific cluster definition. At least one endpoint must advertise the cluster in the requested direction and have a matching registered name, ID and manufacturer code. A conflicting manufacturer override is rejected. The command is sent once to the original group.

This deliberately adds an opt-in exception to the all-member requirement discussed in #1318. Named/numeric lookup and group reads/writes retain their existing behavior.

Companion converters PR: https://github.com/Koenkk/zigbee-herdsman-converters/pull/13197 (draft, awaiting this API and a release containing it).

Validation: build and lint pass; 1,777 tests pass with 100% coverage. The new regression checks exact group frame bytes, both member orders and directions, and rejection of unsupported definitions. It fails before the API change.


src/controller/model/group.ts
```diff
@@ -367,14 +367,41 @@ export class Group extends ZigbeeEntity {
...
         payload: ClusterOrRawPayload<Cl, Co, Custom>,
         options?: Options,
     ): Promise<undefined> {
         const customClusters = this.#customClusters[options?.direction === Zcl.Direction.SERVER_TO_CLIENT ? 1 : 0 /* default to CLIENT_TO_SERVER */];
-        const cluster = Zcl.Utils.getCluster(clusterKey, options?.manufacturerCode, customClusters);
+        const explicit = clusterKey !== null && typeof clusterKey === "object";
+        const cluster = explicit ? clusterKey : Zcl.Utils.getCluster(clusterKey, options?.manufacturerCode, customClusters);
+        if (explicit) {
+            if (
+                cluster.ID < 0xfc00 ||
```
Koenkk, line 388: Is this number not arbitrary?

src/controller/model/group.ts
```diff
@@ -367,14 +367,41 @@ export class Group extends ZigbeeEntity {
...
+        clusterKey: Cl | Cluster,
         commandKey: Co,
         payload: ClusterOrRawPayload<Cl, Co, Custom>,
         options?: Options,
     ): Promise<undefined> {
         const customClusters = this.#customClusters[options?.direction === Zcl.Direction.SERVER_TO_CLIENT ? 1 : 0 /* default to CLIENT_TO_SERVER */];
-        const cluster = Zcl.Utils.getCluster(clusterKey, options?.manufacturerCode, customClusters);
+        const explicit = clusterKey !== null && typeof clusterKey === "object";
+        const cluster = explicit ? clusterKey : Zcl.Utils.getCluster(clusterKey, options?.manufacturerCode, customClusters);
+        if (explicit) {
```
Koenkk, line 386: I'm wondering whether it's worth the complexity to do all this checking here, from the z2m perspective the tzConverter only gets added when there is a group member with this, so the check will always be true (?)

PR #1882, opened 2026-09-19
fix: ember: prevent possible crash on bad ZDO frame

Nerivec:
Re https://github.com/Koenkk/zigbee2mqtt/issues/32567
More of a defensive measure than a fix really.
Seems the root cause is corruption that makes these frames pass as ZDO when they just aren't at all.
Since these bad device frames are somehow able to escape early validation, they are still liable to cause all kinds of problems...

Note: root cause may have been fixed in latest firmware version (report was on very old firmware).


Koenkk:
Thanks!

PR #1884, opened 2026-09-19
Add opt-in deCONZ coordinator service endpoints

strobejb:
## Summary

Adds opt-in support for named coordinator service endpoints, with an initial deCONZ implementation.

This PR introduces a small shared registry for coordinator-side service endpoints. The first registered endpoint is `mieleGateway`, which describes a local coordinator endpoint for Miele's private profile `0xC51E`.

By default nothing changes. No extra service endpoints are enabled unless an adapter is explicitly given a service endpoint name.

## Why

Some devices expect the coordinator itself to expose an application/service endpoint. A device converter can decode messages after they arrive, but it cannot make the coordinator advertise an additional endpoint during ZDO discovery.

The Miele KM6839 induction hob is one example. After joining, it sends a ZDO Match Descriptor request for profile `0xC51E`. On deCONZ/ConBee II, the hob only started sending live private-profile telemetry after the coordinator advertised a matching local endpoint.

## New API

This PR adds a small coordinator endpoint registry in `src/adapter/coordinatorEndpoints.ts`.

It provides:

- named endpoint descriptors, currently `mieleGateway`
- adapter-specific endpoint IDs where required
- validation for duplicate endpoint IDs and invalid cluster/profile data
- helpers for adapters to resolve only the endpoint names they have been asked to enable

The important part is that endpoint definitions are shared, but adapter behavior is still explicit. An adapter only advertises an endpoint if it is passed that endpoint name.

This keeps the API generic without changing default coordinator behavior.

## Adapter behavior

Most coordinator implementations can use the new registry in the normal way: resolve the requested service endpoint descriptors, then append/register them alongside their existing coordinator endpoints during startup.

For example, an adapter with flexible endpoint registration can keep its existing Home Automation and Green Power endpoints, then add `mieleGateway` as another local endpoint.

deCONZ is different because ConBee II firmware appears to expose only two configurable `STK_ENDPOINT` slots. Because of that, this PR implements the deCONZ behavior conservatively:

Default layout remains unchanged:

- slot 0: Home Automation endpoint
- slot 1: Green Power endpoint

When `mieleGateway` is explicitly enabled:

- slot 0: Home Automation endpoint
- slot 1: Miele gateway endpoint, profile `0xC51E`

So Green Power is only replaced on deCONZ when this endpoint is explicitly requested.

## Scope

This PR only wires the registry into deCONZ, because that is the adapter/firmware combination tested with the KM6839.

Other adapters keep their existing behavior. They can opt into the same registry in future PRs by resolving the requested endpoint names and registering those descriptors through their normal local endpoint registration path.

## Follow-up

A Zigbee2MQTT PR is needed to expose a user-facing configuration option and pass the selected service endpoint names into zigbee-herdsman.

A zigbee-herdsman-converters PR adds the KM6839 device converter that decodes the private `0xFD02` telemetry once the coordinator endpoint is available.

## Testing

Tested locally with:

- Zigbee2MQTT 2.12.0
- zigbee-herdsman 10.4.0 plus this patch
- deCONZ adapter
- ConBee II / RaspBee II firmware `0x26780700`
- Miele KM6839 induction hob

Observed behavior:

- default startup keeps the existing Green Power endpoint
- enabling `mieleGateway` replaces the Green Power slot with profile `0xC51E`
- the KM6839 sends live private-profile telemetry after the coordinator advertises the matching endpoint
- existing joined devices continued to operate during testing


Koenkk:
This change is too specific for a too small use-case, what is not working without this change?

strobejb:
Yes, OK, agreed that `mieleGateway` is too device-specific for zigbee-herdsman. 

The actual missing capability is more generic: some devices expect the coordinator itself to advertise a local application endpoint with a specific profile/clusters. A device converter can decode messages after they arrive, but it cannot make the coordinator participate in ZDO Match Descriptor discovery for an additional profile.

Most devices we see support the standard 0x0104 endpoint, so normally this does not matter. The Miele 'Con@ctivity' hobs are maybe just the first case I have hit where a different profile is needed. The hob requests profile `0xC51E`;  on my deCONZ/ConBee II setup, without a matching coordinator endpoint, the live `0xFD02` telemetry does not start. (I believe this affects all coordinators though)

I believe the underlying issue is not deCONZ-specific: any adapter that does not advertise the requested coordinator-side profile would have the same discovery problem. deCONZ/ConBee II is just the setup where I have tested and confirmed it.

I'll rework this PR so zigbee-herdsman only exposes a generic additional coordinator endpoint API, with no Miele-specific preset/name. Instead, Zigbee2MQTT would own the `mieleGateway` preset/config and pass a plain endpoint descriptor into herdsman.

Something like:

```ts
additionalCoordinatorEndpoints?: CoordinatorEndpoint[]

strobejb:
This PR is trying to solve the coordinator-side part of device support for devices that expect the coordinator itself to advertise an application endpoint for a non-standard/private profile.

In this case, the Miele KM6839 hob performs Match Descriptor discovery for the private Miele profile `0xC51E`. The already-merged zigbee-herdsman-converters support can identify and decode the hob once messages arrive, but it cannot make the coordinator advertise a matching local endpoint. Without that coordinator endpoint, the Miele `0xFD02` live telemetry does not start.

## Generic API (current API)

I’ve reworked this so zigbee-herdsman does not contain a Miele-specific endpoint or preset. There is no hardcoded Miele endpoint ID in Zigbee2MQTT either. Existing adapter drivers already define/register coordinator endpoints differently, and some have adapter-specific constraints, so this PR adds a generic opt-in API instead of baking Miele directly into each driver.

The new flow is:

- Zigbee2MQTT passes a descriptor for the required coordinator service endpoint: profile ID, device ID/version, and input/output clusters.
- zigbee-herdsman validates that descriptor.
- Each adapter that supports this feature assigns/registers the actual local endpoint ID using its existing endpoint registration mechanism.
- If additional coordinator endpoints are configured for an adapter that does not support them yet, startup fails with a clear error instead of silently ignoring the setting.

This PR currently wires the mechanism for deCONZ and Ember:

- deCONZ maps the additional endpoint into the available ConBee endpoint slot. On ConBee II this necessarily replaces the Green Power slot when enabled.
- Ember appends the additional endpoint to its existing fixed endpoint registration list, so Green Power remains available.

## Alternative

I considered the smaller alternative of hardcoding the Miele endpoint directly into the affected adapter drivers. That would avoid adding a generic API, but it would also put device-specific knowledge into zigbee-herdsman and would need to be repeated or adapted per adapter.

This PR takes the other approach: zigbee-herdsman provides only the generic mechanism for registering additional coordinator service endpoints, while Zigbee2MQTT owns the device-specific preset/configuration. That keeps Miele-specific details out of herdsman, and gives future devices that need coordinator-side Match Descriptor support the same path.

If the preferred direction is to avoid adding this API for now, I can rework this into adapter-local Miele support instead.

PR #1885, opened 2026-09-21
fix: Ember: stop re-registering a group once the multicast table is full

jonwilliams84:
## What

Stop re-attempting a multicast group registration that the NCP has permanently refused.

## Why

The NCP's multicast table is a fixed size compiled into its firmware — 8 entries by default across the Silabs SDK — and cannot be resized by the host. `emberAdapter.ts` already notes this:

> `WARNING: From here on EZSP commands that affect memory allocation on the NCP should no longer be called (like resizing tables)`

So when a group does not fit, `ezspSetMulticastTableEntry` answers `INVALID_STATE`, and retrying cannot succeed.

The dynamic registration path in `onMessageSent` treated every failure as transient:

```ts
} catch (error) {
    // remove to allow retry on next occurrence
    this.multicastTable.splice(tableIdx, 1);
    logger.error((error as Error).message, NS);
}
```

Dropping the local entry means the next message to that group registers again — and so does the one after that, for the life of the process.

## Evidence

From a live network with nine groups against an eight-entry table (EmberZNet 8.0.2 [GA] build 397, EZSP v14, z2m 2.14.1, 120 devices):

```
[2026-09-21 12:12:36] error: 	zh:ember: Failed to register group '1' in multicast table with status=INVALID_STATE.
[2026-09-21 12:12:59] error: 	zh:ember: Failed to register group '1' in multicast table with status=INVALID_STATE.
[2026-09-21 12:13:36] error: 	zh:ember: Failed to register group '1' in multicast table with status=INVALID_STATE.
[2026-09-21 12:13:59] error: 	zh:ember: Failed to register group '1' in multicast table with status=INVALID_STATE.
```

Roughly twice a minute, unbroken across the whole retained log:

```
log window : 2026-09-12 22:57:24  ->  2026-09-21 12:14:37
total lines: 8489
this error : 7747   (91%)
```

The practical cost is not the retries themselves but what they bury. In that same log were 259 failed TRV writes and ~35 failed availability pings per day, neither of which anyone had noticed.

## What changes

- Statuses meaning "no room" (`INVALID_STATE`, `FULL`, `ALLOCATION_FAILED`) are distinguished from transient failures.
- Such a group is recorded in `multicastTableFull` and not retried.
- One warning is logged instead, naming the group and what can be done:

  > Group '1' does not fit in the NCP's multicast table (8 entries in use, status=INVALID_STATE). The table size is fixed in the NCP firmware and cannot be changed by the host. This group will not receive group-addressed traffic; reduce the number of groups, or flash firmware built with a larger multicast table.

- Any other status keeps the existing retry-and-log-error behaviour, so this is not a blanket "swallow multicast errors".

No behavioural change to which groups work: the multicast table governs *receiving* group-addressed traffic, so an unregistered group could not hear a device bound directly to it before this change either.

## Tests

Two added:

- `Gives up registering a group once the multicast table is full` — five sends produce one registration attempt and one warning. Verified to fail without the source change (`expected "spy" to be called 1 times, but got 5 times`).
- `Still retries registering a group after a transient failure` — pins that `FAIL` still retries and still logs an error.

```
pnpm run check   ✔ 251 files, no errors or warnings
pnpm run build   ✔
pnpm run test    ✔ 34 files, 1779 passed
coverage         ✔ 100% stmts/branch/funcs/lines, emberAdapter.ts included
```

## Related

Reports of the same limit: Koenkk/zigbee2mqtt#32521, #29315, #28297.

---


Nerivec:
Statuses used are too wide on this particular case (it doesn't signals "table full" in a unique fashion), that's why a "stop" wasn't implemented; the cost is "only" 1 failed request on each try.
With this PR, you could be permanently refusing a registration that could otherwise work again on next try.

A better alternative would be to retrieve the table size on startup, and cache it as the limit, but it's not fully reliable as it would require maintaining perfect sync between the host and the coordinator (coordinator could be doing maintenance on its side which could break the sync with the host).

PS: I recommend using a firmware build with higher limit, these are 26 (which should cover 95% of cases): https://github.com/Nerivec/silabs-firmware-builder/releases

jonwilliams84:
Fair — the statuses are too broad, and a permanent refusal is worse than one failed request per retry. Closing.

Went the firmware route instead. `smlight_slzb06m_zigbee_ncp_8.2.2.0_115200_sw_flow.gbl` (v2025.6.2-update1) onto an SLZB-06M:

- EmberZNet `8.0.2 build 397` → `8.2.2.0 build 436`
- `CONFIG_MULTICAST_TABLE_SIZE` 8 → 26
- Network intact, no re-pairing, errors gone

Stuck to stable 8.x rather than 9.1.1 to avoid the stack major. Thanks for the pointer.

